PR #83, opened 2012-02-28
Add git documentation for integrator

sbrunner:


vlasvlasvlas:
Would be great to have a video-tutorial as integrator guide, its really hard to follow.

All the attempts i made for installing c2cgeportal i always got an error and left the installation.

I love this project but its hard / impossible to get an OK installation..

videotutorial as integrator guide may help a lot. thanks!

PR #1167, opened 2015-01-06
Test theme webservice

### LINKED ISSUE #1166: Unique id for layers in the theme service
Each node in the themes tree should have its own unique id.

sbrunner:
And fix it, also fix #1166 , Geoportail-Luxembourg/geoportailv3#129


asaunier:
No objection!

PR #1168, opened 2015-01-07
Add missing WMS layers in the theme webservice

sbrunner:


c2cgeoportal/views/entry.py
```diff
@@ -244,10 +244,12 @@ def _layer(self, layer, wms, wms_layers, time):
                 self._fill_wmts(l, layer, wms, wms_layers, errors)
         elif isinstance(layer, LayerInternalWMS):
             l["type"] = "internal WMS"
+            l["layers"] = layer.layer
```
asaunier: Shouldn't it be a list/array? Something like
```
l["layers"] = [layer.layer]
```

sbrunner: No, in fact `layer.layer` should be named as `layer.layers` and can be a coma separated list of layers...

asaunier: OK, that makes sense.

PR #1177, opened 2015-01-15
Fix master documentation build

sbrunner:


asaunier:
Looks good to me.
But what will Travis say?! :P

PR #1189, opened 2015-01-22
Fix some Make issues mainly dependencies issues

sbrunner:


sbrunner:
@elemoine can you have a look ?

PR #1200, opened 2015-01-29
Validate geometries on update/create

tsauerwein:
Validate the geometries on update and create using the PostGIS functions `ST_IsSimple` and `ST_IsValid`.

### DIFF
--- a/c2cgeoportal/views/layers.py
+++ b/c2cgeoportal/views/layers.py
@@ -43,6 +43,7 @@ from geojson.feature import FeatureCollection, Feature
 
 from shapely.geometry import asShape
 from shapely.ops import cascaded_union
+from shapely.geos import TopologicalError
 
 from papyrus.protocol import Protocol, create_filter
 
@@ -225,7 +226,7 @@ class Layers(object):
 
         layer = self._get_layer_for_request()
 
-        def security_cb(r, feature, o):
+        def check_geometry(r, feature, o):
             geom = feature.geometry
             if geom and not isinstance(geom, geojson.geometry.Default):
                 shape = asShape(geom)
@@ -244,8 +245,16 @@ class Layers(object):
                 if allowed.scalar() == 0:
                     raise HTTPForbidden()
 
-        protocol = self._get_protocol_for_layer(layer, before_create=security_cb)
-        return protocol.create(self.request)
+                # check if geometry is valid
+                self._validate_geometry(spatial_elt)
+
+        protocol = self._get_protocol_for_layer(layer, before_create=check_geometry)
+        try:
+            features = protocol.create(self.request)
+            return features
+        except TopologicalError, e:
+            self.request.response.status_int = 400
+            return {'validation_error': str(e)}
 
     @view_config(route_name='layers_update', renderer='geojson')
     def update(self):
@@ -257,7 +266,7 @@ class Layers(object):
         feature_id = self.request.matchdict.get('feature_id', None)
         layer = self._get_layer_for_request()
 
-        def security_cb(r, feature, o):
+        def check_geometry(r, feature, o):
             # we need both the "original" and "new" geometry to be
             # within the restriction area
             geom_attr, srid = self._get_geom_col_info(layer)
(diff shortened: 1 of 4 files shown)


c2cgeoportal/views/layers.py
```diff
@@ -244,8 +245,16 @@ def security_cb(r, feature, o):
...
 
-        protocol = self._get_protocol_for_layer(layer, before_create=security_cb)
-        return protocol.create(self.request)
+                # check if geometry is valid
+                self.validate_geometry_(spatial_elt)
+
+        protocol = self._get_protocol_for_layer(layer, before_create=check_geometry)
+        try:
+            features = protocol.create(self.request)
+            return features
```
asaunier: Is this intermediary variable "features" really necessary?

tsauerwein: It's not necessary, but I wanted to make it clear that `protocol.create()` returns a list of features.

c2cgeoportal/views/layers.py
```diff
@@ -283,8 +293,27 @@ def security_cb(r, feature, o):
...
 
-        protocol = self._get_protocol_for_layer(layer, before_update=security_cb)
-        return protocol.update(self.request, feature_id)
+            # check is geometry is valid
+            self.validate_geometry_(spatial_elt)
+
+        protocol = self._get_protocol_for_layer(layer, before_update=check_geometry)
+        try:
+            feature = protocol.update(self.request, feature_id)
+            return feature
```
asaunier: Same question about the intermediary variable.

asaunier: Or maybe for the readability of the code? (showing explicitly what is returned here)

asaunier:
Looks good to me!

c2cgeoportal/views/layers.py
tsauerwein: Same reason. :) I had to open the source code of `papyrus` to know what `protocol.update()` returns. This variable should make it a bit clearer.

asaunier: OK, that makes sense!

c2cgeoportal/views/layers.py
```diff
@@ -283,8 +293,27 @@ def security_cb(r, feature, o):
...
+        except TopologicalError, e:
+            self.request.response.status_int = 400
+            return {'validation_error': str(e)}
+
+    def validate_geometry_(self, geom):
```
sbrunner: Why this name ?
To mark it as private it should be prefixed by an underscore, not postfixed ...
https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

vars.yaml
```diff
@@ -14,6 +14,7 @@ vars:
         "role_id" "^-?[0-9]*$"'
     mapfile_data_subselect: "SELECT ST_Collect(ra.area) FROM main.restrictionarea AS ra, main.role_restrictionarea AS rra, main.layer_restrictionarea AS lra, main.treeitem AS la WHERE rra.role_id = %role_id% AND rra.restrictionarea_id = ra.id AND lra.restrictionarea_id = ra.id AND lra.layer_id = la.id AND la.name ="
     fqdn: facter fqdn
+    geometry_validation: False
```
sbrunner: I think that should be true by default :-)

c2cgeoportal/views/layers.py
tsauerwein: Sorry, I confused this with the ol3 code style.

sbrunner:
Missing `prwgma: nocover` 
https://coveralls.io/files/447823921#L312

tsauerwein:
> Missing prwgma: nocover 

So that the coverage does not decrease? You are cheating! ;)

sbrunner:
Yes I'm cheating :-)

More seriously, for me it's acceptable to don't test all exceptions cases, and is we don't use the `pragma: nocover` It's difficult to see it code that should be tested is relay tested :-)

tsauerwein:
Ok, I see your point.

https://github.com/camptocamp/c2cgeoportal/pull/1203

Issue #1195, opened 2015-01-26
Mobile - Marker when passing coordinates


kalbermattenm:
Is there an "easy" way to implement the display of a marker in the mobile app when passing coordinates & zoom through the URL ?

The idea would be to add a custom variable in the URL and to display the map overlaid by a marker (located on the passed coordinates) if this variable exists, something like:

http://example.com/mobile?map_x=554225&map_y=200210&map_zoom=12&marker=1

(This is fake...)
![example](https://cloud.githubusercontent.com/assets/1681332/5902459/e19265c0-a57a-11e4-8fb3-92b807725d05.png)

kalbermattenm:
@gnerred and @rbovard, would that be something interesting for you ?

gnerred:
Yes it would be interesting

rbovard:
Why not, but maybe also think about adding this to the desktop version.

I like the way how OSM.org included this feature in the permalink:

![image](https://cloud.githubusercontent.com/assets/4549666/5918082/d44d1f72-a624-11e4-9e53-5d488cc5293b.png)

kalbermattenm:
Could you give an example of permalink ? I can't reproduce it (yeah... don't ask...) on openstreetmap.org and I would like to see what is the name of the variable ?

rbovard:
It seems the marker can be not in the middle of the map: http://www.openstreetmap.org/?mlat=46.38231&mlon=6.24008#map=19/46.38231/6.24008

rbovard:
FYI to add the marker when generating the permalink:

![image](https://cloud.githubusercontent.com/assets/4549666/5918647/7af63b28-a62b-11e4-8a57-645b7eae43ef.png)

kalbermattenm:
Thanks I did see that, but didn't get that there wasn't a query variable explicitly defining the marker...

Personally, I would rather have a marker added only when it is explicitly declared as a variable... Something like `...&addMarker`

rbovard:
Yes I think your first proposal is fine, the marker is in the middle of the map.

The reference to OSM was more about how to add a marker when generating a permalink.

sbrunner:
Just an additional info:
With Ext it's not possible to have a paramer like `marker=1` the parameter should be named as `map_marker` if it's the map component who manage the marker.
Otherwise that looks good :-)

kalbermattenm:
I have done something...

The code is here: https://github.com/camptocamp/sitn_c2cgeoportal/pull/852

[...]

PR #1204, opened 2015-02-02
Mobile crosshair

kalbermattenm:
This PR adds the possibility to display a crosshair (`map_crosshair`) when calling the mobile app.

Currently it only implements the `map_crosshair` URL variable for the mobile app.

Please review

### DIFF
--- a/c2cgeoportal/scaffolds/update/+package+/static/mobile/app/plugin/StatefulMap.js
+++ b/c2cgeoportal/scaffolds/update/+package+/static/mobile/app/plugin/StatefulMap.js
@@ -70,7 +70,11 @@ Ext.define('App.plugin.StatefulMap', {
                     }, 1);
                 }
                 if (state.lonlat) {
-                    map.setCenter(state.lonlat, state.zoom);
+                    if (state.crosshair === 1) {
+                        main.setCenterCrosshair(state.lonlat, state.zoom);
+                    } else {
+                        map.setCenter(state.lonlat, state.zoom);
+                    }
                 }
                 map.events.on({
                     moveend: this.update,
@@ -117,6 +121,9 @@ Ext.define('App.plugin.StatefulMap', {
         if (q.map_x && q.map_y) {
             state.lonlat = [ q.map_x, q.map_y ];
             state.zoom = q.map_zoom;
+            if (q.map_crosshair) {
+                state.crosshair = 1;
+            }
         }
         return state;
     },
--- a/c2cgeoportal/scaffolds/create/+package+/static/mobile/config.js_tmpl
+++ b/c2cgeoportal/scaffolds/create/+package+/static/mobile/config.js_tmpl
@@ -8,6 +8,7 @@
  */
 
 OpenLayers.Lang.setCode("${lang}");
+OpenLayers.ImgPath = "${request.static_url('{{package}}:static/lib/cgxp/core/src/theme/img/ol/')}";
 
 var dummy = "<% from json import dumps %>";
 jsonFormat = new OpenLayers.Format.JSON();
--- a/c2cgeoportal/scaffolds/update/+package+/static/mobile/app/controller/Main.js
+++ b/c2cgeoportal/scaffolds/update/+package+/static/mobile/app/controller/Main.js
@@ -345,5 +345,39 @@ Ext.define('App.controller.Main', {
                 return false;
             }
         }, this);
+    },
+
+    setCenterCrosshair: function(lonlat, zoom) {
+
+        var map = this.getMainView().getMap();
+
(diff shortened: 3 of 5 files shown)


kalbermattenm:
See https://github.com/camptocamp/c2cgeoportal/issues/1195 for more info

c2cgeoportal/scaffolds/update/+package+/static/mobile/app/plugin/StatefulMap.js
```diff
@@ -117,6 +123,9 @@ Ext.define('App.plugin.StatefulMap', {
         if (q.map_x && q.map_y) {
             state.lonlat = [ q.map_x, q.map_y ];
             state.zoom = q.map_zoom;
+            if (parseInt(q.map_marker) === 1) {
```
sbrunner: I just see that it also exists in the desktop app with the name 'map_crosshair':
https://github.com/camptocamp/cgxp/blob/master/core/src/script/CGXP/widgets/MapPanel.js#L152
I think that good to use the same name...
and we just do an `if (q.map_crosshair) ...`, than it can be '1', 'true' , ...

kalbermattenm: Thanks for the review, I will rename `map_marker` to `map_crosshair` and try to simplify the code as suggested !

kalbermattenm:
Hmmm... I don't really get why Travis is failing...

sbrunner:
A little misstake in the docs:
https://travis-ci.org/camptocamp/c2cgeoportal/jobs/49157812#L1346
https://github.com/camptocamp/c2cgeoportal/pull/1204/files#diff-e97f2398edb60e8ee99533d92f8588c2R182
:-)

kalbermattenm:
Thanks a lot @sbrunner, I did get that the title underlines had to be of the same length as the title...

I applied your suggestions. Ready to be reviewed again (if Travis is happy...)

sbrunner:
@pgiraud can you review this pull request ?

pgiraud:
From what I understand, when the url is updated the "map_crosshair" query param is added if it was already there.
This means that if someone receives a link to a map by mail or text message. He/she opens it in a mobile browser, the crosshair is shown at the center of the map. If the map is panned and page reloaded, the crosshair will still be displayed but a a new coordinate.
Also, this user may want to send a permalink of the map he/she is currently looking at. The recipient will see a crosshair which may not be what the sender wants.
I'd personally leave `update` and `applyState` unchanged.

kalbermattenm:
Hmmmm... You are right, but actually adding the parameter doesn't place a crosshair automatically. The parameter is added to the URL, but is equal to nothing, thus no crosshair on the map.

But I wouldn't mind to do as you suggest. Do you want me to do so ?

Anyway, thanks a lot for the review @pgiraud

kalbermattenm:
I have applied @pgiraud's review and have tested it on our dev instance.

He was right, it looks nicer, because we don't have that `map_crosshair` variable hanging around the URL...

c2cgeoportal/scaffolds/update/+package+/static/mobile/app/plugin/StatefulMap.js
```diff
@@ -70,7 +70,12 @@ Ext.define('App.plugin.StatefulMap', {
                     }, 1);
                 }
                 if (state.lonlat) {
-                    map.setCenter(state.lonlat, state.zoom);
+                    if (state.crosshair === 1) {
+                        App.crosshair = 1;
```
pgiraud: We probably don't need this anymore.

kalbermattenm:
Alright, thanks @pgiraud, sorry for not have seen that...

I applied your second review (and rebased the whole thing, because I do not think it is worth 6 commits...). I just hope I didn't break anything (notably Travis..)

kalbermattenm:
Uh??? https://travis-ci.org/camptocamp/c2cgeoportal/jobs/50632559#L1106

What does this PR have to do with `apache/frontend.conf.mako` ???

sbrunner:
Sorry it my fault, I break the master branch when I merge the 1.5 onto the master ...
I will repaire this soon :-)

kalbermattenm:
No prob ;-)

I leave to you !!

sbrunner:
Just fixed on master :-)

pgiraud:
The code looks good for me. @sbrunner please merge when Travis is happy.

kalbermattenm:
Cool ! Thanks a lot @sbrunner and @pgiraud !

Do I have to do something for Travis to re-run ?

pgiraud:
I relaunched it.

PR #1208, opened 2015-02-03
Fix spelling

pgiraud:


travis/pip.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Used to return the result code given by pip instance of by grep
+# Used to return the result code given by pip instead of grep
```
asaunier: The original comment is weird but does the suggested change really make sense? :P

sbrunner: I add this comment for a previous version that less comprehensible, I think that this comment can be removed.

pgiraud: OK. Please do so.

PR #1210, opened 2015-02-03
Backport modified CONST from lux

sbrunner:


asaunier:
No objection from my point of view.

PR #1212, opened 2015-02-05
Remove about 6000 log spamming lines

sbrunner:


asaunier:
Less is best :)
If you think those lines are not useful, then OK for me.

sbrunner:
6000 of 8000 lines just for the deploy => :-( If we need them for a special case we can reactivate them :-)

PR #1198, opened 2015-01-28
Fix cache for autenticated parent theme

sbrunner:

PR #1214, opened 2015-02-05
Fix, str and int concatenation

sbrunner:
fix https://github.com/camptocamp/c2cgeoportal/pull/1198#issuecomment-71990517


asaunier:
Looks good to me.

PR #1216, opened 2015-02-06
Simplify check_collector project configuration

sbrunner:


asaunier:
Looks good to me.
I just wonder if it would not be more consistent to rename the "disable" setting (in the yaml) to "disabled"?

PR #1221, opened 2015-02-11
Fix instanceid in project.yaml.in template

arnaud-morvan:


c2cgeoportal/scaffolds/create/project.yaml.in_tmpl
```diff
@@ -1,6 +1,6 @@
 project_folder: {{project}}
 project_package: {{package}}
 host: <host>
-checker_path: /${instance-id}/wsgi/check_collector?
+checker_path: /${instanceid}/wsgi/check_collector?
```
asaunier: Shouldn't it be ``${vars:instanceid}`` as in https://github.com/camptocamp/c2cgeoportal/blob/1.5/c2cgeoportal/scaffolds/create/config.yaml.in_tmpl#L91

arnaud-morvan: Les deux fonctionnent, la section "vars" doit être utilisé par défaut sans doute ?

asaunier: It's better to be consistent and use "vars:instanceid" everywhere.
I see that https://github.com/camptocamp/c2cgeoportal/blob/1.5/c2cgeoportal/scaffolds/create/production.ini.in_tmpl#L13 use "instanceid" but "vars:instanceid" a few lines latter:
https://github.com/camptocamp/c2cgeoportal/blob/1.5/c2cgeoportal/scaffolds/create/production.ini.in_tmpl#L48
Ditto for development.ini.in_tmpl.
It would be great if you could make the changes here as well :)

arnaud-morvan: This seems to be out of my skills and attributions.
No error in master branch, which has been corrected here https://github.com/camptocamp/c2cgeoportal/commit/c5d3673a5351d001f6b2056c9c0c45d4f56d7b97

PR #1224, opened 2015-02-11
Add pdfreport webservice

asaunier:
This webservice allows the users to get a full PDF report concerning a given feature (layername+id), using  MapFish Print 3 and some jasper reports.

WIP, still TODO:
* handling mapserver groups

### DIFF
--- a/c2cgeoportal/views/pdfreport.py
+++ b/c2cgeoportal/views/pdfreport.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2011-2015, Camptocamp SA
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# The views and conclusions contained in the software and documentation are those
+# of the authors and should not be interpreted as representing official policies,
+# either expressed or implied, of the FreeBSD Project.
+
+from pyramid.view import view_config
+from pyramid.response import Response
+from pyramid.httpexceptions import HTTPBadGateway, HTTPForbidden, \
+    HTTPInternalServerError
+
+import httplib2
+from urlparse import urlparse
+from json import dumps, loads
+
+from c2cgeoportal.lib.filter_capabilities import get_protected_layers, \
+    get_private_layers
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class PdfReport:  # pragma: no cover
+
+    def __init__(self, request):
+        self.request = request
+        self.config = self.request.registry.settings.get('pdfreport', {})
+
+    def do_print(self, spec):
+        """ Get created PDF. """
+        url = self.config['print_url'] + "/print/buildreport.pdf"
+        http = httplib2.Http()
+        h = dict(self.request.headers)
+        h["Content-Type"] = "application/json"
+        if urlparse(url).hostname != 'localhost':
+            h.pop('Host')
+        try:
+            resp, content = http.request(
+                url, method='POST', headers=h, body=dumps(spec)
+            )
+        except:
+            return HTTPBadGateway()
+
+        headers = {}
+        if 'content-type' in resp:
+            headers['content-type'] = resp['content-type']
+        if 'content-disposition' in resp:
+            headers['content-disposition'] = resp['content-disposition']
+        return Response(
+            content, status=resp.status, headers=headers
+        )
+
+    @view_config(route_name='pdfreport', renderer="json")
+    def getreport(self):
+        id = int(self.request.matchdict['id'])
+        layername = self.request.matchdict['layername']
+        featureid = layername + "." + str(id)
+
+        # check user credentials
+        role_id = None if self.request.user is None else \
+            self.request.user.role.id
+
+        # FIXME: support of mapserver groups
+        if layername in get_private_layers() and \
+                layername not in get_protected_layers(role_id):
+            raise HTTPForbidden
+
+        srs = self.config.get('srs')
+        if srs is None:
+            raise HTTPInternalServerError(
+                'Missing "srs" in service configuration'
+            )
+        params = {
+            'service': 'WFS',
+            'version': '1.0.0',
+            'request': 'GetFeature',
+            'typeName': layername,
+            'featureid': featureid,
+            'srsName': srs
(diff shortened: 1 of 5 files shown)


c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+            headers['content-disposition'] = resp['content-disposition']
+        return Response(
+            content, status=resp.status, headers=headers
+        )
+
+    @view_config(route_name='pdfreport', renderer="json")
+    def getreport(self):
+        id = int(self.request.matchdict['id'])
+        layername = self.request.matchdict['layername']
+        if id is None or layername is None:
```
sbrunner: Not possible ... :-)

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+                layername not in get_protected_layers(role_id):
+            raise HTTPForbidden
+
+        params = {
+            'service': 'WFS',
+            'version': '1.0.0',
+            'request': 'GetFeature',
+            'typeName': layername,
+            'featureid': featureid,
+            'outputFormat': 'geojson'
```
sbrunner: GeoJson ? 
I think you should add come documentation about this ...

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+        }
+        srs = self.config.get('srs_name')
+        if srs is not None:
+            params['srsName'] = srs
+
+        request_url = self.config['mapserv_url'] + "?" \
+            + "&".join(["%s=%s" % i for i in params.items()])
+
+        baselayer = self.config.get("layers", {}). \
+            get(layername, {}).get('baselayers')
```
sbrunner: layers.baselayers is a new settings ?

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+        if baselayer is None:
+            baselayer = self.config.get('default_baselayers')
```
sbrunner: also ?

c2cgeoportal/views/pdfreport.py
asaunier: OK I'll remove it.

c2cgeoportal/views/pdfreport.py
asaunier: Yes, this requires adding a "geojson" outputformat in the mapfile. Some doc is still needed for this part.

c2cgeoportal/views/pdfreport.py
sbrunner: You use layers namespace but it's normally reserved for layers view!
I think that should be 'pdfreport' or 'pdfreport.layers'

sbrunner: in GeoMapFish 2.0 we prefer the therm 'backgroundlayers' instance of 'baselayers'.

asaunier: Doc needed here as well. Here is an example of config in config.yaml.in:
```
pdfreport:
    mapserv_url: https://${host}/${vars:instanceid}/wsgi/mapserv_proxy
    print_url: http://localhost:8080/print-c2cgeoportal-v3-${vars:instanceid}
    default_baselayers: ['grp_ly_tilegenerierung_landeskarte']
    default_imageformat: 'image/png'
    srs_name: 'EPSG:21781'
    layers:
        ly_a020_belastete_standorte_point:
            baselayers: ['grp_ly_tilegenerierung_landeskarte','ly_a020_belastete_standorte_point']
```

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+        spec = {
+            "layout": layername,
+            "outputFormat": "pdf",
+            "attributes": {
+                "paramID": id,
+                "map": {
+                    "projection": srs,
+                    "dpi": 254,
+                    "rotation": 0,
+                    "bbox": [0, 0, 1000000, 1000000],
```
sbrunner: This looks like a workaround... is it report i mapfish-print ?

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+                    "zoomToFeatures": {
+                        "zoomType": "center",
```
sbrunner: shouldn't be bbox instance of center ?

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,162 @@
...
+                                "fillColor": "red",
+                                "fillOpacity": 0.2,
+                                "symbolizers": [{
+                                    "strokeColor": "red",
+                                    "strokeWidth": 1,
+                                    "type": "point",
+                                    "pointRadius": 10
+                                }]
+                            }
+                        },
```
sbrunner: I think that the style should be in the config

c2cgeoportal/views/pdfreport.py
sbrunner: I think that the MapFish Print support the GML, isn't it the case ?

c2cgeoportal/views/pdfreport.py
sbrunner: `mapserv_url` should be the same as `request.get_route('mapserv_proxy')`
`print_url` already exists can we reuse it ?

Otherwise that looks good :-)

asaunier: > You use layers namespace but it's normally reserved for layers view!
I think that should be 'pdfreport' or 'pdfreport.layers'

That's already the case, see the config example above. In the service we have
```
self.config = self.request.registry.settings.get('pdfreport', {})
```

> print_url already exists can we reuse it ?

Not really because we have a project that use MF Print 3 for this pdfreport tool but still use MF Print 2 for the standard printing.

c2cgeoportal/views/pdfreport.py
asaunier: I don't know, this is a part added by @gilliM

c2cgeoportal/views/pdfreport.py
asaunier: Actually in some projects the map would not be used. So I think that one should be able to pass the whole ``spec`` in the config.

c2cgeoportal/views/pdfreport.py
sbrunner: OK, I miss that :-)

c2cgeoportal/views/pdfreport.py
sbrunner: I agree, but it will be not easy to do ...

c2cgeoportal/views/pdfreport.py
asaunier: > I think that the MapFish Print support the GML, isn't it the case ?

I don't know. If I replace
```
"geoJson": request_url
```
by
```
"gml": request_url
```
(with request_url being the WFS request with no output format - default is GML), the printing fails saying that:
```
Error while processing request:
org.mapfish.print.ExtraPropertyException: Extra properties were found in the request data at: spec.attributes.map.layers[0]: 
Extra Properties: 

	* gml

All allowed properties are: 

	* style
	* name
	* opacity
	* renderAsSvg
	* geoJson
```

tsauerwein: mfp 3 supports GML, you'll have to set the layer type to `type: gml` and then give an `url`.

asaunier: @tsauerwein Thanks for the tip, it indeed works perfectly :)

sbrunner: Cool :-)

asaunier:
I have added a documentation and implemented the suggested improvements.
Please note that the full JSON spec may now be provided in the configuration if needed.

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,169 @@
...
+                            "opacity": 1,
+                            "type": "WMS",
+                            "serverType": "mapserver",
+                            "layers": [{backgroundlayers}],
+                            "imageFormat": "{imageformat}"
+                        }}]
+                    }}
+                }}
+            }}
+            """
```
sbrunner: :( I relay don't like this template ..., why not using only in the case we specify it in the configuration ...
and with: `loads(dumps(self.config.get("spec")) % (...))` ?

I use the % to avoid issue with `{}`

c2cgeoportal/views/pdfreport.py
```diff
@@ -0,0 +1,169 @@
...
+            raise HTTPForbidden
+
+        params = {
+            'service': 'WFS',
+            'version': '1.0.0',
+            'request': 'GetFeature',
+            'typeName': layername,
+            'featureid': featureid
+        }
+        srs = self.config.get('srs_name')
```
sbrunner: Shouldn't simply be `srs` ?

asaunier: OK

c2cgeoportal/views/pdfreport.py
asaunier: >  why not using only in the case we specify it in the configuration 

In order to use the same consistent mechanism whereas a spec_template is provided or not: if it works for default template, it should work for a param template.

sbrunner: Never mind do it as structured config in booth side and call every tine the command I purpose ...

asaunier:
* Fixed commits up
* Used JSON config as suggested by @sbrunner 
* Updated doc
* Added ``# pragma: no cover`` to disable testing because PDF output is difficult to test (as for the printproxy service)

sbrunner:
Excellent, that looks good to me :-)

asaunier:
OK, thanks for reviewing Steph!

PR #1225, opened 2015-02-12
Remove support of Mapserver 6.x and PostGis 1.5

sbrunner:


asaunier:
No objection.
Should you also update the requirements here (for postgis 1.5):
http://docs.camptocamp.net/c2cgeoportal/master/integrator/requirements.html
?

sbrunner:
Yes of course ...

PR #1232, opened 2015-02-13
Add cookie timeout

sbrunner:

PR #1233, opened 2015-02-16
Fix Admin interface

### LINKED ISSUE #1186: layer in multiple groups
A group can now be part of only one group.
If I add a layer to a new group, the layer will be removed from the its old group.

sbrunner:
Depends on #1232 
Fix #1186 
Also requires https://github.com/Pylons/pyramid/issues/1580


ochriste:
looks ok to me

PR #1234, opened 2015-02-17
Add missing static views for Ngeo

sbrunner:


c2cgeoportal/__init__.py
```diff
@@ -266,6 +266,9 @@ def new_f(root, request):
         renderer=renderer
     )
 
+    config.add_static_view('node_modules', config.get_settings().get('node_modules_path'))
+    config.add_static_view('closure', config.get_settings().get('closure_library_path'))
```
elemoine: For non-ngeo projects, `get_settings().get('node_modules_path')` will return `None`. Did you make sure that this will not make `add_static_view` fail?

elemoine:
Please merge when https://github.com/camptocamp/c2cgeoportal/pull/1234#discussion_r25150919 is answered.

c2cgeoportal/__init__.py
sbrunner: We are in `add_interface_ngeo` function than if we (from the project) call it for a non-ngeo application it's a mistake :-)

PR #1237, opened 2015-02-17
Fix Postgres package version command

sbrunner:


c2cgeoportal/scaffolds/update/CONST_packages.yaml
```diff
@@ -8,7 +8,7 @@ main:
         cmd: /usr/bin/python --version 2>&1 | awk '{{print $2}}'
         version: 2.7
     postgres:
-        cmd: sudo -u postgres /usr/lib/postgresql/9.1/bin/postmaster --version | awk '{{print $3}}'
+        cmd: sudo -u postgres /usr/lib/postgresql/*/bin/postmaster --version | awk '{{print $3}}'
         version: 9.1
```
asaunier: If you remove the version in the "cmd" part, shouldn't you drop this line too?
What if several versions of postgres are available?

sbrunner: As I test it should works but Travis complain about it, than I change it :-)

asaunier:
Looks good to me.

PR #1242, opened 2015-02-18
Add missing 'pragma: nocover' for admin interface

sbrunner:


asaunier:
No objection from my perspective.

PR #1243, opened 2015-02-18
Fix make dependencies date issue between submobules

sbrunner:
When we change the head of CGXP the GCXP submodules shouldn't be fetch


asaunier:
OK for me.

PR #1251, opened 2015-02-20
Fix sencha touch dev page and dependencies

### LINKED ISSUE #956: The mobile_dev view don't work
For example at http://mapfish-geoportal.demo-camptocamp.com/demo/wsgi/mobile_dev/ we have the error:
```
Error: [Ext.Loader] Failed loading 'touch/src/fx/animation/Pop.js', please verify that the file exists
throw new Error("[Ext.Loader] " + errorMessage);
```
and the page don't load ....

sbrunner:
Fix #956


asaunier:
OK for me.

PR #1253, opened 2015-02-23
Report restriction area sub select fix by @tsauerwein

sbrunner:


sbrunner:
up ...

asaunier:
Looks good to me!

PR #1255, opened 2015-02-23
Add the header Vary:Accept-Language

sbrunner:
on pages that changed with the language
remove the lang attibute on pages that don't change with the language


asaunier:
OK for me.

PR #1258, opened 2015-02-23
Fix initial check in RestrictionArea Layer link

sbrunner:


c2cgeoportal/forms.py
```diff
@@ -292,7 +295,7 @@ def render_item(self, item, depth):
             # don't save them twice (=> integrity error).
             'name': self.name + ("-second" if final_item.id in self._rendered_id else ""),
             'value': final_item.id,
-            'add': ' checked="checked"' if self._is_checked(item.id) else "",
+            'add': ' checked="checked"' if self.is_checked(item, final_item) else "",
```
asaunier: Why not directly using ``self._is_checked(final_item.id)``?

sbrunner: To be able to overwrite it at line 341 :-)

asaunier: OK, no other objection!

PR #1260, opened 2015-02-24
Fix typo in CONST_Makefile

elemoine:
Because of this typo the js code is not linted on make checks.


sbrunner:
Thanks @elemoine, merge :-)

PR #1267, opened 2015-02-25
Fix theme editing don't work

sbrunner:
The admin interface isn't configured on new projects


asaunier:
The title of the first commit is not really correct (you should reformulate). Else OK for me.

PR #1268, opened 2015-02-25
Add Mapfish Print V3 support

sbrunner:
Use an attr for all functional tests
Refractor the proxies

First step to use MapFish Print V3, the pull request on GeoExt and on GGXP will follow the next weeks.


asaunier:
Wow, huge work, thanks. No objection from my point of view.

PR #1273, opened 2015-02-26
Change pyramid_closure config for camptocamp/pyramid_closure#50

elemoine:
This changes the `pyramid_closure` config in `CONST_vars.yaml` to camptocamp/pyramid_closure#50 in the pyramid_closure package. This makes the debug mode work again, with and with the cache buster enabled.


elemoine:
Patch updated, as discussed with @sbrunner.

sbrunner:
Looks good, I'm happy to don't see those '../' anymore :-)
Thanks for the doc :-)

PR #1279, opened 2015-03-02
Add static Alembic chain, move user on static schema

sbrunner:


asaunier:
Commit 5c84500cd4128fae1fc0ef5167a73d6092bcf652 contains a "pyo" file: c2cgeoportal/scaffolds/update/CONST_alembic/env.pyo
Is it really wanted?

sbrunner:
No, thanks ...

sbrunner:
Finally yes, I remove the binary file :-)

asaunier:
No other "objection". OK for me when Travis passes.

rbovard:
@sbrunner Why did you removed the password replication?

sbrunner:
It will be done because we have the user table in main schema and then when we deploy the application we lost the changed password. Now the user table is in the static schema than we don't have anymore this issue and we don't need anymore this functionality.

rbovard:
And for the people who deploy all the DB... ?

sbrunner:
We have two schema to clearly separate the data that are modified on the client application than it's a relay bad idea to deploy all the database ...

And from now isn't a no good ides to deploy all the database, you will lost all the editing layers (that's possible that you haven't) but you also lost your short URL ...

rbovard:
Mmm ok... I know it's not the better way to deploy all the DB but it works. I'll change my procedures, no problems.

Please just don't forget to mention that in changelog for people who using it.

sbrunner:
Effectively :-)

PR #1280, opened 2015-03-03
Use asset override to get URLs

### LINKED ISSUE #48: settings['project'] should go away
The ```entry``` view defines this function:

```
    def _getIconPath(self, icon):
        if not icon:
            return None
        icon = str(icon)
        # test full URL
        if not icon[0:4] == 'http':
            try:
                # test full resource ref
                if icon.find(':') < 0:
                    if icon[0] is not '/':
                        icon = '/' + icon
                    icon = self.request.static_url(self.settings['project'] + ':static' + icon)
                else:
                    icon = self.request.static_url(icon)
            except ValueError:
                log.exception('can\'t generate url for icon: %s' % icon)
                return None
        return icon
```

So It uses ```static_url(self.settings['project'] + ':static' + icon)``` to get URLs for static resources served by the project's static view.

For this to work the project's ```.ini``` file defines a ```project``` variable in the ```app``` section:

```
[app:${package}]
use = egg:${package}
project = ${package}
```

We should change that, and rely on a Pyramid's [assets overrides](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#overriding-assets). So the project's ```__init__.py``` file would define an asset override with something like this:

```
config.override_asset(to_override='c2cgeoportal:static/',
                                        override_with='${package}:static/')
```

and the c2cgeoportal would use calls like ```static_url('c2cgeoportal:static/icon.png')```.

See this [discussion](https://groups.google.com/group/pylons-discuss/msg/4c8e525f488a2230) for example.

sbrunner:
fix #48

### DIFF
--- a/c2cgeoportal/__init__.py
+++ b/c2cgeoportal/__init__.py
@@ -282,7 +282,11 @@ def add_admin_interface(config):
 
 def add_static_view(config):
     """ Add the project static view """
-    _add_static_view(config, 'proj', '%s:static' % config.get_settings()["package"])
+    _add_static_view(config, "proj", "%s:static" % config.get_settings()["package"])
+    config.override_asset(
+        to_override="c2cgeoportal:project/",
+        override_with="%s:static/" % config.get_settings()["package"]
+    )
 
 
 def _add_static_view(config, name, path):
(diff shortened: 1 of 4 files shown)


asaunier:
OK for me.

PR #1281, opened 2015-03-04
Group mapserver proxy settings

### LINKED ISSUE #1054: Group mapserver proxy settings
There are several settings parameters used in mapserverproxy https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/views/mapserverproxy.py

They are dispatched in several places in config.yaml. What about grouping them together in a "mapserverproxy" object in config.yaml?

sbrunner:
fix #1054

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -159,10 +159,18 @@ vars:
     # this application is a child application in a parent/child
     # architecture.
     external_themes_url:
-    external_mapserv_url:
-    # If the WFS service is different from the WMS service, uncomment and set the
-    # external_mapserv_wfs_url parameter
-    #external_mapserv_wfs_url:
+
+    mapserverproxy:
+        # URL to internal MapServer instance
+        mapserv_url: http://localhost/{instanceid}/mapserv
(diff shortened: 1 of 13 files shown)


c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -159,10 +159,18 @@ vars:
     # this application is a child application in a parent/child
     # architecture.
     external_themes_url:
-    external_mapserv_url:
-    # If the WFS service is different from the WMS service, uncomment and set the
-    # external_mapserv_wfs_url parameter
-    #external_mapserv_wfs_url:
+
+    mapserverproxy:
```
asaunier: Don't we need to add an entry to the changelog to report this config change?

sbrunner: For me it's included in this point
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl#L7-L24
but I just see thet the `vars_project.yaml` is missing ...

asaunier: OK for me when the missing vars_project.yaml is added.

PR #1284, opened 2015-03-05
Fix issue when we want to use a specific egg version

sbrunner:


c2cgeoportal/scaffolds/create/setup.py_tmpl
```diff
@@ -15,7 +15,7 @@ setup(
     author_email='info@camptocamp.com',
     url='http://www.camptocamp.com/geospatial-solutions',
     install_requires=[
-        'c2cgeoportal',
+        'c2cgeoportali>=1.6.0.dev',
```
asaunier: What is "c2cgeoportali"? :P

rbovard: A little c2cgeoportal?

sbrunner: ;vi)

PR #1290, opened 2015-03-10
Activate geometry_validation by default, Group layers config

sbrunner:


asaunier:
Don't you need to add an entry in the changelog since the PR changes some config, doesn't it?

sbrunner:
No because it's already included in the first point :
https://github.com/camptocamp/c2cgeoportal/blob/editing/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl#L7-L26
:-)

asaunier:
OK, no other objection.

sbrunner:
Thanks :-)

PR #1293, opened 2015-03-11
Many litle fix shown on a project update

sbrunner:


asaunier:
I only see a few commits titles that are confusing, else it looks good to me.

PR #1301, opened 2015-03-20
Add customisable apache entry point

sbrunner:


asaunier:
Is it really customizable, since both files are CONST files?

sbrunner:
Yes, You add `APACHE_ENTRY_POINT = /` in your Makefile and because we have a `?= ` instance of `=` the value will not be overwritten :-)

asaunier:
OK
Is is documented somewhere?

sbrunner:
Good point :-)

sbrunner:
Updated :-)

doc/integrator/make.rst
```diff
@@ -64,6 +64,8 @@ The following variables may be set in the makefiles:
 * ``PRE_RULES``: predefine some build rules, default is empty.
 * ``POST_RULES``: postdefine some build rules, default is empty.
 * ``DISABLE_BUILD_RULES``: List of rules we want to disable, default is empty.
+* ``INSTANCE_ID``: The WSGI instance id (should bu unique on a server).
```
asaunier: bu => be

asaunier:
Small typo. Else OK for me.

PR #1312, opened 2015-03-27
Restart Tomcat on each deploy

sbrunner:
See: https://github.com/mapfish/mapfish-print/issues/221#issuecomment-86845418


asaunier:
Looks good to me

PR #1317, opened 2015-04-08
Internationalise distance in profile

### LINKED ISSUE #1263: translate word "distance" in profile.py
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/views/profile.py#L60

Can we make this word translatable?

fr: distance
en: distance
de: Distanz

sbrunner:
fix #1263


asaunier:
Looks good

PR #1319, opened 2015-04-09
Remove dupplicated war

sbrunner:


asaunier:
Arf, you're right, it is already available at https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/print/print-servlet.war

Please merge!

PR #1327, opened 2015-04-17
Add proxy for TinyOWS

tsauerwein:
This PR adds a proxy for TinyOWS which checks if a user has the rights to access a layer. A user has the right to read or write a layer, if the layer belongs to one of the user's restrictions areas and the `readwrite` flag of that restriction area is set to `True`.

It is important to note that the proxy does not take into account the area of a restriction area, contrary to the MapServer proxy which makes sure that only the features inside the defined area are returned. This is because the restriction extent would have to be hard-coded in the TinyOWS configuration file (see attribute [geobbox](http://mapserver.org/tinyows/configfile.html#layer-element)). So, if the access should restricted to a certain area, this area would have to be set in the configuration file.

This PR is not yet finished, but reviews are already welcome! Still to do:

* [x] Update documentation
* [x] Update demo

### DIFF
--- a/c2cgeoportal/views/tinyowsproxy.py
+++ b/c2cgeoportal/views/tinyowsproxy.py
@@ -51,12 +51,8 @@ class TinyOWSProxy(Proxy):
         Proxy.__init__(self, request)
         self.settings = request.registry.settings.get("tinyowsproxy", {})
 
-        assert self.settings.get("tinyows_url", ""), \
+        assert self.settings.get("tinyows_url", "") != "", \
             "tinyowsproxy.tinyows_url must be set"
-        assert self.settings.get("online_resource", ""), \
-            "tinyowsproxy.online_resource must be set"
-        assert self.settings.get("proxy_online_resource", ""), \
-            "tinyowsproxy.proxy_online_resource must be set"
 
     def _get_wfs_url(self):
         return self.settings.get("tinyows_url")
@@ -134,7 +130,7 @@ class TinyOWSProxy(Proxy):
 
     def _get_headers(self):
         headers = Proxy._get_headers(self)
-        if self.settings.get("tinyows_host", "") != "":
+        if "tinyows_host" in self.settings:
             headers["Host"] = self.settings.get("tinyows_host")
         return headers
 
@@ -151,8 +147,8 @@ class TinyOWSProxy(Proxy):
 
         content = self._filter_urls(
             content,
-            self.settings.get("online_resource"),
-            self.settings.get("proxy_online_resource"))
+            self.settings.get("online_resource", ""),
+            self.settings.get("proxy_online_resource", ""))
 
         headers = {
             "Content-Type": resp["content-type"],
@@ -162,7 +158,10 @@ class TinyOWSProxy(Proxy):
         return self._build_response(resp, content, cache, "tinyows", headers)
 
     def _filter_urls(self, content, online_resource, proxy_online_resource):
-        return content.replace(online_resource, proxy_online_resource)
+        if online_resource != "" and proxy_online_resource != "":
+            return content.replace(online_resource, proxy_online_resource)
+        else:
+            return content
 
     def _parse_body(self, body):
         """
--- a/c2cgeoportal/views/tinyowsproxy.py
+++ b/c2cgeoportal/views/tinyowsproxy.py
@@ -132,6 +132,12 @@ class TinyOWSProxy(Proxy):
         writable_layers = get_writable_layers(self.role_id)
         return typenames.issubset(writable_layers)
 
+    def _get_headers(self):
+        headers = Proxy._get_headers(self)
+        if self.settings.get("tinyows_host", "") != "":
+            headers['Host'] = self.settings.get("tinyows_host")
+        return headers
+
     def _proxy_callback(self, operation, role_id, *args, **kwargs):
         cache = kwargs.get('cache', False)
         resp, content = self._proxy(*args, **kwargs)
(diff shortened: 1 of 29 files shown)


sbrunner:
Looks good, when documentation is done and Travis tests pass you can merge :-)

c2cgeoportal/views/tinyowsproxy.py
```diff
@@ -0,0 +1,190 @@
...
+        Checks if the current user has the rights to access the given
+        type-names.
+        """
+        writable_layers = get_writable_layers(self.role_id)
+        return typenames.issubset(writable_layers)
+
+    def _get_headers(self):
+        headers = Proxy._get_headers(self)
+        if self.settings.get("tinyows_host", "") != "":
+            headers["Host"] = self.settings.get("tinyows_host")
```
sbrunner: I don't understand why you need this ...

c2cgeoportal/views/tinyowsproxy.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Checks if the current user has the rights to access the given
+        type-names.
+        """
+        writable_layers = get_writable_layers(self.role_id)
+        return typenames.issubset(writable_layers)
+
+    def _get_headers(self):
+        headers = Proxy._get_headers(self)
+        if self.settings.get("tinyows_host", "") != "":
```
sbrunner: `if "tinyows_host" in self.settings:`
isn't better ?

c2cgeoportal/views/tinyowsproxy.py
tsauerwein: This is only for the debug mode (`make serve`).

c2cgeoportal/views/tinyowsproxy.py
tsauerwein: `self.settings.get("tinyows_host", "") != ""` checks two things: If there is an entry with the key `tinyows_host` and also if a possible value is a non-empty string. So it's not the same as `"tinyows_host" in self.settings`.

tsauerwein:
I updated the documentation and demo. Please review.

ping @yjacolin

c2cgeoportal/views/tinyowsproxy.py
sbrunner: Yes but settings `tinyows_host` to "" should be en error not a disable ...

sbrunner:
two untested lines:
https://coveralls.io/builds/2393880/source?filename=c2cgeoportal%2Fviews%2Ftinyowsproxy.py#L134
https://coveralls.io/builds/2393880/source?filename=c2cgeoportal%2Fviews%2Ftinyowsproxy.py#L164
And I should know why you requires this `tinyows_host`

c2cgeoportal/views/tinyowsproxy.py
tsauerwein: `tinyows_host` is optional. If it's not given in the settings, it's not set on the header.

sbrunner: Then `tinyows_host` shouldn't be in the settings ...

tsauerwein: Where else?

I know you are not using `make serve` but others are. And when you are using `make serve`, you explicitly have to set the host. Then it would be consistent to set the host together with all the other settings for TinyOWS. And like it is [documented](https://github.com/tsauerwein/c2cgeoportal/blob/73484ab87d6061013dde5bd980b15ec25205cfd2/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl#L186-189), it is not required to set the host.

sbrunner: Thanks for the second response, but set `tinyows_host` to "" steal haven't any sense ...

sbrunner:
Thanks for the tests :-)

c2cgeoportal/views/tinyowsproxy.py
tsauerwein: It is not set to `""`.

    if self.settings.get("tinyows_host", "") != "":

is the same as:

    if self.settings.get("tinyows_host") is not None and self.settings.get("tinyows_host") != "":

sbrunner: Yes I know it but why testing that `tinyows_host` is not `""` is un nessery ... it's yhy I largely prefer:
`if "tinyows_host" in self.settings:`

tsauerwein: Why is it unnecessary? I only want to change the host if a valid host name is given. That is a non-empty string.

Because this is only used for debugging, it's not so important. But I don't see your point.

sbrunner: More readable syntax ...

tsauerwein: > More readable syntax ...

Readable syntax, yes. But it's not doing the same thing. If you want to check if there is a non-empty string in the settings, IMHO `self.settings.get("tinyows_host", "") != ""` is an elegant way to do so.

sbrunner: I say again:
1. We don't need the second test
2. It's not elegant

tsauerwein: > I say again:
1. We don't need the second test

Thanks for repeating this, again. But you are again missing an argument why it's not needed. Like I said before, I am testing for a non-empty string, because it doesn't make sense to set an empty string as host name.

But hey, because it's only used in debug mode, I made the change. For you. :heart: 

> 2. It's not elegant

This is your opinion. I am happy to learn a more elegant way to test if there is a non-empty string for a key in the settings.

sbrunner:
Many thanks @tsauerwein, If yous think that's ready you can merge :-)

tsauerwein:
Thanks for the review.

PR #1344, opened 2015-05-11
Pillow and other wheels in Windows

kalbermattenm:
I have the same problem with Pillow as we had with Psycopg2 and Shapely.

It has to be installed using a wheel containing the already compiled dll and binaries.

Furthermore, we cannot keep these `http://www.lfd.uci.edu/~gohlke/pythonlibs/sjdmq9dh/Shapely-1.5.7-cp27-none-win32.whl` kind of URLs, because they are only valid a day... (the `sjdmq9dh` code changes each day...)

I also tried to explain that in the doc...

Please review

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_requirements.txt
+++ b/c2cgeoportal/scaffolds/update/CONST_requirements.txt
@@ -3,6 +3,7 @@
 -r CONST_versions.txt
 psycopg2==2.6
 Shapely==1.5.6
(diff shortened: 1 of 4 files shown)


sbrunner:
@asaunier can you have a look on the documentation ?
for me that looks good :-)

asaunier:
Looks good to me. I merge.

kalbermattenm:
Thanks a lot

PR #1341, opened 2015-05-08
First steps to improve the documentation for Windows

kalbermattenm:
This PR rewrites the documentation for Windows.

There are still things to add like downloading wheels, which pip cannot install because of binaries and dll, but this will be part of another PR (which I hope to propose soon...)

Please review

PR #1347, opened 2015-05-13
Doc to avoid permission problems in Cygwin

kalbermattenm:
Sorry, I did not know about that when I submitted #1341...


sbrunner:
Nice :-)

kalbermattenm:
Thanks !

PR #1355, opened 2015-05-18
Backport print and Tomcat service on Windows

kalbermattenm:
If using Mapfish print v3, then there are some commands which are run to start/stop the Tomcat service.

On Windows, the syntax of these ones is not the same then on Linux.

This PR:
* Defines a new print variable in `CONST_Makefile_tmpl` which is null by default (thus not altering the current build process).
* Adds some doc in the Windows specific section

If anybody has a better idea, then it is welcomed...

Please review

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -214,6 +214,7 @@ PRINT_REQUIREMENT += \
 	$(PRINT_BASE_DIR)/WEB-INF/lib/postgresql-9.3-1102.jdbc41.jar \
 	$(PRINT_BASE_DIR)/WEB-INF/classes/logback.xml \
 	$(PRINT_BASE_DIR)/WEB-INF/classes/mapfish-spring-application-context-override.xml
+PRINT_TMP ?= /tmp
 TOMCAT_SERVICE_COMMAND ?= sudo /etc/init.d/tomcat-tomcat1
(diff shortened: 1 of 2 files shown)


asaunier:
Thanks @kalbermattenm, looks good to me.
@sbrunner please merge if you think it is OK as well.

sbrunner:
Why don't you simply add in your project makfile:
```
TOMCAT_START_COMMAND = net START Tomcat7
TOMCAT_STOP_COMMAND = net STOP Tomcat7)
```
I think that much simpler ...

kalbermattenm:
So that way to it would overwrite https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl#L219-L220, right ?

If yes, then this would be a better solution...

sbrunner:
Yes we use a `?=` then it's possible overwrite them :-)

kalbermattenm:
OK, thanks a lot, I will modify this PR accordingly (just add some doc...)

kalbermattenm:
OK....

I modified the doc and then I realized that @sbrunner and myself forgot to backport a Windows "speciality"... Indeed, in Windows, you cannot use a `tmp` folder which does not already exist when running a `cp` command.

Thus with @sbrunner, we had defined a new variable in the makefile to be able to overwrite it for Windows..

I rebased the whole thing...

@sbrunner, could you review that once again ?

kalbermattenm:
Just updated my commit to make Travis happy...

sbrunner:
Thanks :-)

PR #1362, opened 2015-05-26
Backport from Luxembourg

sbrunner:
Geoportail-Luxembourg/geoportailv3#362
Geoportail-Luxembourg/geoportailv3#340


asaunier:
Looks good, but shouldn't you use a more explicit commit name?

sbrunner:
Update the commit message with the original pullrequests

asaunier:
I meant something telling explicitely what the commit is about. But if you think that's enough like this, please merge.

PR #1363, opened 2015-05-26
Add print cancel, specify reqest method everywhere

sbrunner:
Related to https://github.com/Geoportail-Luxembourg/geoportailv3/issues/478

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /c2cgeoportal/tests/testegg/production.ini
 /c2cgeoportal/locale/c2cgeoportal.pot
 /c2cgeoportal/locale/*/LC_MESSAGES/c2cgeoportal.mo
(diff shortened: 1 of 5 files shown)


asaunier:
What the link between GET request and printing cancellation?

sbrunner:
Witch GET request ? the request method ?

asaunier:
The one from
https://github.com/camptocamp/c2cgeoportal/commit/6e48432d743eb1b80ed62f571d92beb43d8304d9

sbrunner:
Just that I see that when I add the print cancel :-)

asaunier:
OK :)
Perhaps creating separated PR would have make more sense, but no big deal.

It looks good to me. Do you think adding ``request_method="GET"`` in project-specific routes are needed as well? If yes, shouldn't you add a note in the changelog?

sbrunner:
No it's not a requirement it's just better :-)

PR #1376, opened 2015-06-04
Fix request method for "loginchange" route

tsauerwein:
Must be `POST` not `GET`.


sbrunner:
Thanks @tsauerwein , merge :-)

PR #1383, opened 2015-06-05
Prepare the RC1

sbrunner:


kalbermattenm:
With that you force the integrators to use a certain version...

Before I was able to specify a version through `setup.py` by specifying `c2cgeoportal-win==xxxx`.

Is that still possible ? Because c2ctool downloads the latest `CONST_requirements.txt`...

So for now, I cannot test the whole thing, because the rc1 is not out yet... Personally, I think it would be good to leave the c2cgeoportal version choice to the integrators, don't you think ?

sbrunner:
I like that If we do an upgrade to a specific version that we force to have this version, and it's not new things, see:
https://github.com/camptocamp/c2cgeoportal/blob/1.5/c2cgeoportal/scaffolds/create/versions.cfg#L3
I just remove it during the development because it's not usable.

kalbermattenm:
Yes, but this is on branch 1.5, not on master like this PR...

sbrunner:
Yes it's because we enter in stabilisation mode, when we will branch the version 1.6, to start the version 1.7 or 2.0 we will remove it :-)

PR #1393, opened 2015-06-08
Makefile print section & yaml directory variable

kalbermattenm:
This PR does implement two things:
* Print section for Windows: this is needed because Cygwin handles the existing bash commands in the Windows drive definition, thus it is "not" working. I opted for a specific Windows section, because the behavior differs quite a lot between Linux and Windows systems.
* directory variable: if this variable is defined through bash (in Windows...), then `${directory}` is substituted by `/cygdrive/c/...` everywhere. This does not work when calling the app through Apache (or even pserve) as this one expects to get values like `C:/...` on Windows. Defining `directory` using Python resolves this issue

Please review

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -672,14 +672,26 @@ $(VENV_BIN)/flake8: .build/dev-requirements.timestamp
 print: $(PRINT_OUTPUT)/$(PRINT_WAR)
 
 $(PRINT_OUTPUT)/$(PRINT_WAR): $(PRINT_REQUIREMENT)
+ifeq ($(VENV_BIN), .build/venv/bin)
 	cp $(PRINT_BASE_DIR)/$(PRINT_BASE_WAR) $(PRINT_TMP)/$(PRINT_WAR)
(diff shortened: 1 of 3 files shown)


kalbermattenm:
It looks like Travis is trying to use what I wrote... I don't really get why it goes into the ifeq condition (it shouldn't, no ? Isn't it using a Linux architecture...) ?

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -672,14 +672,26 @@ $(VENV_BIN)/flake8: .build/dev-requirements.timestamp
 print: $(PRINT_OUTPUT)/$(PRINT_WAR)
 
 $(PRINT_OUTPUT)/$(PRINT_WAR): $(PRINT_REQUIREMENT)
+ifeq (VENV_BIN, .build/venv/bin)
```
sbrunner: `VENV_BIN` should be `$(VENV_BIN)` :-)

And you should do a rebase to make the tests working :-)

kalbermattenm:
Thanks for the hint @sbrunner ! Travis is now happy...

sbrunner:
Thanks @kalbermattenm :-)

PR #1399, opened 2015-06-09
Replace : by = in SQL query

kalbermattenm:
I do not think that `:` is the good relational equal operator in SQL. In my opinion, it should be `=` (and this is working in my project...)

Or did I miss something concerning new SQL syntax ?


sbrunner:
Many tanks :-)

PR #1401, opened 2015-06-09
Move the EN translations in the code

sbrunner:


c2cgeoportal/models.py
```diff
@@ -194,7 +194,7 @@ class User(Base):
     id = Column(Integer, primary_key=True)
     username = Column(
         Unicode, unique=True, nullable=False,
-        label=_(u"Username")
+        label=_(u"User name")
```
asaunier: Username is a single word

sbrunner: Fixed, thanks :-)

asaunier:
Just curious why you do this change since the translating is done anyway?

sbrunner:
Because in Transifex we see the key and we should see the source language :-)

PR #1408, opened 2015-06-10
Add some documentation about MapFishPrint v3

### LINKED ISSUE #1345: MapFish print 3
Is there somewhere a "howto" for migrating from MapFish print v2 to v3 ?

The only thing I found was <a href="https://github.com/camptocamp/c2cgeoportal/tree/master/c2cgeoportal/scaffolds/create/print" target=_blank>here</a>. I also saw that CGXP should be ready for MapFish print v3...

I could try to do something, because the proposed template is quite "light", but as anybody already done something more elaborated ?

sbrunner:
Will fix #1345 ?

### DIFF
--- a/doc/integrator/print.rst
+++ b/doc/integrator/print.rst
@@ -19,8 +19,51 @@ list of options the plugin can receive.
 
 *The main viewer includes a Print plugin by default.*
 
-MapFish Print configuration
----------------------------
+Using MapFish Print v3
+----------------------
+
+Migration from the v2
+~~~~~~~~~~~~~~~~~~~~~
+
+Start the migration by instantiating the default template:
+
+.. prompt:: bash
+
+   cd <project_root>
+   cp -r print ~
+   git rm -r print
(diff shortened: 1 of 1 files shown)


doc/integrator/print.rst
```diff
@@ -19,8 +19,38 @@ list of options the plugin can receive.
 
 *The main viewer includes a Print plugin by default.*
 
-MapFish Print configuration
----------------------------
+Start with MapFish Print v3
```
asaunier: Is this title relevent? Why "start"? What about something like "Using MapFish Print v3"?

doc/integrator/print.rst
```diff
@@ -19,8 +19,38 @@ list of options the plugin can receive.
...
+----------------------------
+
+Start the migration by instanciating the default template:
+
+.. prompt::
+
+   cp -r print ~
```
asaunier: Where does the "print" dir come from?

sbrunner: From the project ...

asaunier:
> From the project ...

Yes but adding something like 
```
cd <project root dir>
```
would be more explict.

Else looks good.

PR #1410, opened 2015-06-10
Fix installed pyramid version

sbrunner:
Actually after a cleanall / build we have the latest stable version
instance of the development version.


asaunier:
Please merge

PR #1435, opened 2015-07-09
Use full virtualenv path for more robustness

elemoine:
I use https://github.com/brainsik/virtualenv-burrito (on my laptop and on dev servers). I use this to install recent versions of virtualenv and virtualenvwrapper in my homedir. This also adds `/home/elemoine/.venvburrito/bin` to my `PATH`, meaning that the `virtualenv` command is `/home/elemoine/.venvburrito/bin/virtualenv` in my case.

And building c2cgeoportal projects doesn't work with recent versions of virtualenv (and pip). So, for more robustness, this PR suggests using `/usr/bin/virtualenv` instead of `virtualenv` in `Makefile` and `CONST_Makefile`.

Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ C2C_TEMPLATE_CMD = .build/venv/bin/c2c-template --vars $(VARS_FILE)
 
 DEVELOPPEMENT ?= FALSE
 
+VIRTUALENV_CMD ?= /usr/bin/virtualenv
 PIP_CMD ?= .build/venv/bin/pip
 PIP_INSTALL_ARGS += install --trusted-host pypi.camptocamp.net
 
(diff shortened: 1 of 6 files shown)


pgiraud:
Is there any reason why c2cgeoportal doesn't work with recent versions of virtualenv?
Travis doesn't seem to be happy with this modification. `/usr/bin/virtualenv` doesn't seem to be valid there.

elemoine:
> Is there any reason why c2cgeoportal doesn't work with recent versions of virtualenv?

I get an error saying that `zc.recipe.egg` is not a valid wheel. So it seems we have dependencies that cannot be installed with recent versions of virtualenv.

> Travis doesn't seem to be happy with this modification. /usr/bin/virtualenv doesn't seem to be valid there.

Sh^W. I am going to remove the last commit and see what happens…

elemoine:
It does not work. I'll see what I can do…

elemoine:
Ok, I got the build to pass. But there is one change I can't make:

```diff
diff --git a/c2cgeoportal/scaffolds/update/CONST_packages.yaml b/c2cgeoportal/scaffolds/update/CONST_packages.yaml
index db2b378..5c5c1c7 100644
--- a/c2cgeoportal/scaffolds/update/CONST_packages.yaml
+++ b/c2cgeoportal/scaffolds/update/CONST_packages.yaml
@@ -24,7 +24,7 @@ main:
         version: 6.0
     libproj0: 4.7
     virtualenv:
-        cmd: virtualenv --version
+        cmd: /usr/bin/virtualenv --version
         version: 1.7
     httpd:
         cmd: /usr/sbin/apache2ctl -v 2> /dev/null | head -n 1 | awk '{{print $3}}' | awk -F / '{{print $2}}'
```

With this change the build fails on Travis when `travis/build-new-project.sh` is run.

So I think I am just going to abandon this PR for now.

PR #1436, opened 2015-07-09
remove POST limitation in route login

ochriste:
so the login can use other authentication system


elemoine:
Looks good to me. Please merge.

sbrunner:
Why other authentication system can't do a POST request ?

ochriste:
because we dont always have the hand on how it works,
for example the tequila auth system from EPFL return a get

PR #1437, opened 2015-07-09
Add a note about the Tomcat heap size to the docs

tsauerwein:
... and fix the wording a bit.


elemoine:
Please merge if you're lucky enough to get the Travis build to pass!

tsauerwein:
Thanks for the review.

sbrunner:
This is only for GeoMapfish 2.0 ?

doc/integrator/requirements.rst
```diff
@@ -55,6 +55,17 @@ default is: ``http://localhost:8080/print-c2cgeoportal-{instanceid}/pdf/``.
...
 to change it set the ``PRINT_OUTPUT`` value in the makefile.
 
+For MapFish Print to function properly it is recommended to increase the default
+Java Heap Size for Tomcat (parameter ``-Xmx`` and ``-Xms``). On servers managed
+by Camptocamp check the size in the file ``/srv/tomcat/tomcat1/bin/setenv-local.sh``:
+
+  .. code:: make
+
+    export JAVA_XMX="1G"
+    export ADD_JAVA_OPTS="-Xms1G"
```
sbrunner: This should be always done ? 
Do we relay need 1G, I think that 512M is enough...
Why do you add Xms ?

tsauerwein: > Do we relay need 1G, I think that 512M is enough...

@jesseeichar recommended 1GB.

> Why do you add Xms ?

See http://www.ibm.com/developerworks/websphere/techjournal/0909_blythe/0909_blythe.html#minmax

> As mentioned earlier, it is generally recommended that the minimum and maximum heap sizes are set to the same value to maximize run time performance. This prevents the JVM from compacting, which dynamically resizes the heap. This is a very expensive operation and can be as much as 50% of garbage collection pause times. Setting the min heap size equal to the max heap size also makes garbage collection analysis easier by fixing a key parameter to a constant value.

sbrunner: OK, thanks for your comment :-)

doc/integrator/requirements.rst
```diff
@@ -55,6 +55,17 @@ default is: ``http://localhost:8080/print-c2cgeoportal-{instanceid}/pdf/``.
 And by default the 'webapps' folder is ``/srv/tomcat/tomcat1/webapps``,
 to change it set the ``PRINT_OUTPUT`` value in the makefile.
 
+For MapFish Print to function properly it is recommended to increase the default
+Java Heap Size for Tomcat (parameter ``-Xmx`` and ``-Xms``). On servers managed
+by Camptocamp check the size in the file ``/srv/tomcat/tomcat1/bin/setenv-local.sh``:
```
sbrunner: This:

> On servers managed by Camptocamp check the size in the file ``/srv/tomcat/tomcat1/bin/setenv-local.sh``:

is wrong, it should be done by the sysadmins (as you correctly ask) but the sysadmins will do it in the `/srv/tomcat/tomcat1/bin/setenv.sh` file ...

PR #1440, opened 2015-07-10
Make make clean remove the main Apache conf file

elemoine:
Currently, running `make clean` breaks `apache2ctl configtest` (and `apache2ctl graceful`) for every developer on the server. This commit fixes the problem by making `make clean` also remove the main Apache config file for the project.


sbrunner:
Nice, but it shouldn't be on the 1.6 branch ?

sbrunner:
Done by an other pull request, closing

PR #1446, opened 2015-07-22
Remove the no more needed documentation

### LINKED ISSUE #1434: Deprecated commands in the documentation
Some commands are not longer correct in the 1.6 documentation.

In the [internationalization] (http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/internationalization.html?highlight=install%20po2mo) file, the command ``make -f <user>.mk build install po2mo`` must be replaced by ``make -f <user>.mk build``.

If you search "graceful" in the documentation, you'll find 4 results. Three are just after a ````make build````. This command already calls an Apache's graceful. So, this is useless and needs correction.

In the file [install_application] (http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/install_application.html?highlight=apache/*.conf) we also must remove the instructions to create manually the Apache's .conf file. ``Make build`` already does it.

Thanks to correct that. I don't propose a pull request because I'm soon two weeks in holidays :sunglasses:

sbrunner:
fix #1434

### DIFF
--- a/doc/integrator/install_application.rst
+++ b/doc/integrator/install_application.rst
@@ -246,9 +246,9 @@ user, which is the Apache user. In RHEL 6, there's no user ``www-data``, and
 the Apache user is ``apache``. To accomodate that edit ``vars_<project>.yaml`` and
 set ``modwsgi_user`` to ``apache`` in the ``[vars]`` section::
 
-    [vars]
-    ...
-    modwsgi_user = apache
+    vars:
+        ...
+        modwsgi_user: apache
 
 
 Also, by default, the path to Tomcat's ``webapps`` directory is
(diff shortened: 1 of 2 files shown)


doc/integrator/install_application.rst
```diff
@@ -272,9 +272,9 @@ apache2ctl
 ~~~~~~~~~~
 
 On RedHat the commands hasn't the '2'!
```
asaunier: This sentence doesn't look correct (plural subject with singular verb).

sbrunner: Thanks, fixed :-)

asaunier:
Please merge.

PR #1447, opened 2015-07-22
Use Scala better

sbrunner:


sbrunner:
Scala specific - Gatling scripts => merging

PR #1450, opened 2015-07-23
Use export_local_property_definitions Closure Compiler flag

fredj:
Needed to use the `@export` jsdoc annotation


sbrunner:
Looks good :-)

Issue #1400, opened 2015-06-09
To backport


sbrunner:
* [x] https://github.com/Geoportail-Luxembourg/geoportailv3/pull/465
 * [x] https://github.com/Geoportail-Luxembourg/geoportailv3/pull/524
 * [x] https://github.com/Geoportail-Luxembourg/geoportailv3/pull/541
 * [x] https://github.com/Geoportail-Luxembourg/geoportailv3/pull/653
 * [x] https://github.com/Geoportail-Luxembourg/geoportailv3/pull/697 (remove angular-animate)

PR #1456, opened 2015-07-23
Update angular to v1.4.3

fredj:
See #1400


sbrunner:
Looks good, thanks :-)

PR #1463, opened 2015-07-27
Create static-ngeo directory

fredj:


sbrunner:
As asked a new pyramid static view is required

sbrunner:
Should we also move the output dir ?

asaunier:
Just curious: what's the use of this nes static-ngeo dir?

sbrunner:
Clearly separate the ngeo file to the cgxp files :-)

asaunier:
OK thanks
I didn't paid attention that GMF could be used whether with CGXP or ngeo.

fredj:
> Should we also move the output dir ?

yes: builded files should be saved into `static-ngeo/build`

fredj:
> As asked a new pyramid static view is required

https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/__init__.py#L82

fredj:
@sbrunner do you agree to merge this PR as it ?

fredj:
Merging as it; will do more in other pull requests ...

c2cgeoportal/scaffolds/create/+package+/templates/ngeo.html_tmpl
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="\{\{mainCtrl.lang\}\}" ng-app="app" ng-controller="MainController as mainCtrl">
```
elemoine: Why are the backslashes necessary? I am asking the question because we don't have backslashes in geoportailv3. https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/templates/index.html

fredj: the backslashes are necessary otherwise the scaffolding code complains that `mainCtrl.lang` is not defined when generating `ngeo.html` from `ngeo.html_tmpl`

elemoine: Understood. And `package` is effectively a scaffold variable. Thanks for the answer.

PR #1466, opened 2015-07-29
Simplify multi project instance by removing many makefile creation

sbrunner:


doc/integrator/create_application.rst
```diff
@@ -489,11 +482,20 @@ Create the project
    +dest = /var/www/vhosts/<project>/conf/${instance}.conf
    +content = Include /var/www/vhosts/<project>/private/${instance}/apache/*.conf
 
-12. In ``.gitignore`` add the following lines:
+12. Update the deploy configuration as follow:
```
asaunier: as follows

asaunier:
Just curious why the _jinja extension are no longer needed?

sbrunner:
Because we no longer create a template of makefile that create a set of make file, we change it to have a generic makefile and select the right project with an environment variable.

asaunier:
OK thanks for the explanation.
LGTM

PR #1468, opened 2015-07-29
Fix gxp repository url

fredj:


sbrunner:
Thanks, but is should be on the 1.6 branch ... :-)

PR #1469, opened 2015-07-29
Fix gxp repository url and demo urls

fredj:


sbrunner:
Thanks merge when Travis tests pass :-)

PR #1472, opened 2015-07-30
Remove groups when all child layers have been stripped out of getcap

asaunier:
As for now, the filter_capabilities tool removes not allowed layers from the WMS getcapabilities response. On the other hand if the removed layers are contained in a mapserver group, the group remains even though no embedded layers are shown.

This PR makes sure the group is removed as well if all contained layers have been removed.

Most of the code is from @sbrunner

### DIFF
--- a/c2cgeoportal/lib/filter_capabilities.py
+++ b/c2cgeoportal/lib/filter_capabilities.py
@@ -184,6 +184,8 @@ class _layer:
         self.accumul = []
         self.hidden = True
         self.self_hidden = self_hidden
+        self.has_children = False
+        self.children_nb = 0
 
 
 class _capabilities_filter(XMLFilterBase):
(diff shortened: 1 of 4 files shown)


sbrunner:
Looks good but the tests should be updated

asaunier:
It seems that the changes broke the WFS filtering. Commit 0a22be3 should fix this.

For the tests I am going to add the case of a group with only private layers.

asaunier:
Commits have been merged/rebased, tests added and pass. The changes have been tested for real with BL's GeoWMS.

sbrunner:
Looks good for me, merge :-)

asaunier:
OK merging.
Thanks @sbrunner for reviewing and your help to write the code.

PR #1478, opened 2015-08-06
Use git://github.com/ in package.json

sbrunner:


fredj:
+1

asaunier:
Shouldn't we use https URLs instead for users that block the ssh port on their server?

sbrunner:
Good question, I think that we should keep it in head for the moment we have the issue.

PR #1490, opened 2015-08-07
Upgrade the python packages versions

sbrunner:


elemoine:
LGTM.

fredj:
@sbrunner Pylons was not removed from `CONST_dev-requirements.txt`: is it correct?

sbrunner:
Pylons => Pyramid :-)

the master egg will be removed from CONST_dev-requirements.txt
But the alpha version is added in the CONST_versions.txt : `pyramid==1.6a2`

fredj:
ok, thanks for the explanation

PR #1491, opened 2015-08-07
Add missing dependency

sbrunner:


elemoine:
Could you please provide a quick explanation on why this dependency is required? Please merge when done.

sbrunner:
I type on a clean project:
``make -f ... .build/build.js``
and the build failed because the mako-render commends don't exists ...

elemoine:
Obviously. Thanks.

PR #1497, opened 2015-08-11
do not inject role in all generated urls

ochriste:
if role is injected in all the url, it becomes impossible to change it afterward (for example as in project epfl_geoportail)m, because Openlayers only merge params if they dont exist. 
if the parameter already, as with "role" in this case, OL skip the parameter inclusion, meaning the parameter value will always be the one included server-side.

this PR propose to change the unique cache id by removing the auto injection of role in all generated urls.

this is only a proposal, open for discussion.


sbrunner:
For me this will don't change anything ...

ochriste:
fixed with sbrunner

PR #1502, opened 2015-08-13
deploy is a custom c2c package, it shouldnt be checked by default

ochriste:
should the deploy package really be checked at that level?
because this will only work on a c2c server, since deploy is a c2c specific package


sbrunner:
I don't agree with it, deploy should be put on a separate section as it's done for `mobile`, `tilecloud-chain` and `ngeo`.

PR #1504, opened 2015-08-14
Typos fix

asaunier:
* Concerning "seam" => "seem", I have also noticed the problem in the GMF 1.6 migration tool:
```
(...)
.build/venv/bin/c2c-versions CONST_packages.yaml main
(...)
postgis doesn't seam to be installed (required version: 2.0).
```
but have not found where this error message was built...

* Concerning the DB migration doc, shouldn't we replace "<schema_name>" by "main"? It's the standard name, isn't it?

* Concerning the "advanced version" fix, what does "advanced version" means here? It think it's confusing.

### DIFF
--- a/doc/integrator/upgrade_application.rst
+++ b/doc/integrator/upgrade_application.rst
@@ -365,10 +365,10 @@ First rename all the conflicting items:
 
    .. code:: sql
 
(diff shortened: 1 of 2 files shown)


sbrunner:
Looks good, merge :-)

PR #1506, opened 2015-08-14
Fix confusing sentence about root sections in jsbuild/app.cfg.mako

asaunier:
Else one may think updating only the root section of the first file (app.js) is enough.


sbrunner:
Looks good, merge :-)

PR #1481, opened 2015-08-06
Some issue found during an upgrade

sbrunner:
See commits for details

PR #1508, opened 2015-08-14
PRINT_TMP is also needed when building print V2

ochriste:


ochriste:
already done in https://github.com/camptocamp/c2cgeoportal/pull/1481

PR #1514, opened 2015-08-17
Fix Travis error with alembic

sbrunner:


tsauerwein:
Travis passes, so I guess it's ok.

sbrunner:
Thanks :-)

PR #1520, opened 2015-08-18
Fix the default mobile theme

### LINKED ISSUE #1496: Default mobile theme does not work
Version 1.6.1 of c2cgeoportal

The default mobile theme doesn't work. It appears, but empty, at the first load of the application, but is not available in the theme list. Can be reproduced here (it has to contain 4 layers) : http://preprodnew.cartoriviera.ch/mobile

sbrunner:
fix #1496

See result at: http://preprodnew.cartoriviera.ch/sbrunner/mobile

### DIFF
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -1031,7 +1031,10 @@ class Entry(object):
         user = self.request.user
 
         role_id = None if user is None else user.role.id
-        themes, errors = self._themes(role_id, interface, False)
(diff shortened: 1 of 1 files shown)


asaunier:
> See result at: http://preprodnew.cartoriviera.ch/sbrunner/mobile

I see
```
NameError: global name 'response' is not defined
```
:P

sbrunner:
Effectively I just do some tests ..., => fixed :-)

asaunier:
Looks good to me

PR #1530, opened 2015-08-19
Remove workaround for previous nose version

sbrunner:


asaunier:
Commit message should be "Remove workaround for previous nose version". Else no objection.

PR #1531, opened 2015-08-20
Use wheel instance of source egg when it's possible

sbrunner:


asaunier:
No objection from my perspective.

PR #1532, opened 2015-08-20
Add the theme in the Full-Text Search

sbrunner:
Add some columns in the Full-Text Search table
Add a script to fill the Full-Text Search table with the theme
Update the Full-Text Search view to take care on the new columns

Also see others cmmits


asaunier:
No more objections, looks good to me.

PR #1548, opened 2015-08-25
Add difference geometry operation

sbrunner:


doc/developer/webservices.rst
```diff
@@ -240,3 +289,26 @@ Result
 
 The 'Content-Type' header is 'text/csv',
 and the data contains the given 'csv' data.
+
+Geometry operations
+===================
+
+This service gives geometry operations (actually only one)
```
asaunier: actually => currently

asaunier:
Looks good

c2cgeoportal/tests/functional/test_geometry_operation.py
```diff
@@ -0,0 +1,58 @@
...
+    tear_down_common as tearDownModule,
+    set_up_common as setUpModule,
+    create_dummy_request
+)
+
+
+@attr(functional=True)
+class TestGeometryOperation(TestCase):
+
+    def test_defference(self):
```
pgiraud: typo e -> i

doc/developer/webservices.rst
pgiraud: gives => provides

doc/developer/webservices.rst
```diff
@@ -240,3 +289,26 @@ Result
...
+
+Method: ``POST``
+
+Data:
+
+.. code:: json
+
+   {
+       "geom": <The geom to extrude>
+       "diff": <The geom to extrude with>
```
pgiraud: What about geomA & geomB?

pgiraud:
According to postGIS doc, ST_Difference is a processing, not an operation.

doc/developer/webservices.rst
sbrunner: For me geomA & geomB mean nothing, then I try to use better names :-)

sbrunner:
Review applyed :-)

c2cgeoportal/tests/functional/test_geometry_processing.py
```diff
@@ -0,0 +1,58 @@
...
+
+from c2cgeoportal.tests.functional import (  # noqa
+    tear_down_common as tearDownModule,
+    set_up_common as setUpModule,
+    create_dummy_request
+)
+
+
+@attr(functional=True)
+class TestGeometryOperation(TestCase):
```
pgiraud: Please change Operation to Processing.

doc/developer/webservices.rst
pgiraud: But it's compliant with the PostGIS doc. http://postgis.net/docs/ST_Difference.html
Also, we often see schemas where A & B names are used.
text4247

Even if not perfect, I find geomA and geomB easier to understand. I would be interested to know what others think.

c2cgeoportal/tests/functional/test_geometry_processing.py
sbrunner: Oups :-)

doc/developer/webservices.rst
sbrunner: on Sqlite it's geom1, geom2
on argis, geometry1, geometry2
on OGC: g1, g2 (http://portal.opengeospatial.org/files/?artifact_id=25354)

then I can put g1, g2 event if I think that's worth ...

sbrunner: Finally I think that the different names are so bad that I can have something like this:
``` json
{
    arguments: [<geomA>, <geomB>]
}
```
What do you think ?

pgiraud: What about "geometries" or "geoms" instead of "arguments".

sbrunner: OK pour `"geometries"` :-)

sbrunner:
Updated :-)

PR #1553, opened 2015-08-26
Convert METADATA wms_srs to uppercase

arnaud-morvan:
Change "wms_srs" "epsg:..." to "wms_srs" "EPSG:..." in c2cgeoportal.map.mako.
Tell to do the same in all mapfiles in the changelog.

### DIFF
--- a/c2cgeoportal/scaffolds/create/mapserver/c2cgeoportal.map.mako
+++ b/c2cgeoportal/scaffolds/create/mapserver/c2cgeoportal.map.mako
@@ -71,7 +71,7 @@ MAP
             "wms_title" "changeme"
             "wms_abstract" "changeme"
             "wms_onlineresource" "http://${host}/${instanceid}/wsgi/mapserv_proxy"
-            "wms_srs" "epsg:21781"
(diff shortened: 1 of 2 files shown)


asaunier:
Looks good. But maybe you should explain in the changelog why epsg code must be uppercased?

arnaud-morvan:
I've seached the web, but not found real explanation.
It is wrote uppercase everywhere (in mapserver documentation, WMS specification) but it is not explained or clearly specified.
I've added link to http://www.mapserver.org/ogc/wms_server.html#setup-a-mapfile-for-your-wms

asaunier:
I don't really see any explanations or advices about the case of EPSG codes in that page :P 
I would rather write that using lowercase codes may cause incompatibility problems with some client tools including WMSBrowser.

arnaud-morvan:
I've pasted your explanation in the changelog.
Thanks for review.

asaunier:
Thanks for the changes. Looks good to me!

sbrunner:
For me also :-)

PR #1558, opened 2015-08-26
Allow to call mapserver without any params

sbrunner:


asaunier:
What the purpose of calling mapserver with no parameter?

sbrunner:
Yes

sbrunner:
Nothing, but its better to have the mapserver message then a crash ...

asaunier:
Looks good then

PR #1560, opened 2015-08-27
Changelog updates

sbrunner:


sbrunner:
Thanks, updated :-)

asaunier:
Looks good. Just wondering if "parameter" wouldn't fit more than "variable"?

sbrunner:
The vars files contains variables, for me parameters are for query string, function, ...

asaunier:
Makes sense!

PR #1563, opened 2015-08-27
Remove deprecated part of the changelog file

asaunier:
Since upgrading to 1.6 requires to have the very last version of 1.5, one doesn't really need to have all the old changelog instructions for GMF 1.5 and lower.


sbrunner:
Looks good, merge :-)

PR #1566, opened 2015-08-27
Rebuild the mako when we modify the vars

sbrunner:


asaunier:
Isn't it already the case since
```
VARS_DEPENDS += $(VARS_FILES) .build/node_modules.timestamp
```
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl#L283

sbrunner:
You point the project CONST_Makefile but I modify the c2cgeoportal Makefile :-)

asaunier:
OK then

sbrunner:
Thanks :-)

PR #1571, opened 2015-08-31
Fix "CPU on 150%" issue with print

tsauerwein:
Updates MFP to https://github.com/mapfish/mapfish-print/commit/8a69f61e48b352a71ec78fa2b313f9455f3d2871 to include the fix of PR https://github.com/mapfish/mapfish-print/pull/306


sbrunner:
Thanks, merge :-)

PR #1572, opened 2015-09-01
Fix typo and incorrect filename in CHANGELOG

asaunier:
By the way what are the "required changes" mentioned at point 2 of 1.6.3?


sbrunner:
Many thanks, merge when Travis tests pass :-)

asaunier:
What about the required changes to do in print/config.yaml.mako?

sbrunner:
As I remember: changing the path because we moved the file

asaunier:
Right:
```
<%include file="/A4_portrait.mako" />
```
should be transformed to
```
<%include file="templates/A4_portrait.mako" />
```
I think it is not explained in the changelog. I will add a line in the changelog.

sbrunner:
+1

asaunier:
> I will add a line in the changelog.

Done

sbrunner:
Thanks :-)

PR #1576, opened 2015-09-02
Add admin_interface key in update_paths

arnaud-morvan:


sbrunner:
Looks good, merge :-)

PR #1578, opened 2015-09-03
Fix typo in AddOutputFilterByType

fredj:
See https://github.com/mapserver/mapserver/blob/master/mapwms.c#L91


asaunier:
Makes sense. But what about branch 1.6?

fredj:
See #1581

PR #1579, opened 2015-09-03
Fix typo in comment

fredj:


asaunier:
Makes sense. But what about branch 1.6?

fredj:
See #1582

PR #1584, opened 2015-09-07
disabling make built-in rules to speed up execution

ochriste:


c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -294,6 +294,162 @@ ENVIRONMENT_VARS += INSTANCE_ID=${INSTANCE_ID} \
 C2C_TEMPLATE_CMD = $(ENVIRONMENT_VARS) $(VENV_BIN)/c2c-template --vars $(VARS_FILE)
 MAKE_FILES = $(shell ls -1 *.mk) CONST_Makefile
 
+# Disabling Make Implicit Rules to speed up execution time
+.SUFFIXES:
+
+# Disabling Make Implicit Rules to speed up execution time
```
sbrunner: Implicit => Built-in

sbrunner:
Just a small comment, otherwise that looks good :-)

fredj:
According to the documentation `.SUFFIXES:` deletes the default suffixes but why `SUFFIXES :=` is sets just after that ?

ochriste:
we made some tests with sbrunner and the execution time was even shorter when overriding the list.

but it seems this has broken the variable replacement system
we need to make some more tests

ochriste:
removed SUFFIXES := as it was causing issue with some tasks execution

sbrunner:
But is steal not `Implicit Rules` but `built-in rules`

sbrunner:
Excellent, merge :-)

PR #1587, opened 2015-09-08
Make git modules folder overridable in Makefile

asaunier:
Useful for projects with an intermediate "main/" folder. Such projects must have in their ``<project>.mk``:
```
GIT_MODULES_FOLDER ?= ../.git/modules/main/
```

Ported from a @sbrunner patch.


fredj:
+1

sbrunner:
We shouldn’t do it like this, we should directly merge the 1.6 branch to master :-(

PR #1592, opened 2015-09-10
add missing capabilities filtering call

ochriste:


sbrunner:
Effectively, it will work better :-), merge!

PR #1593, opened 2015-09-10
Fix the print_url to be able to put the report in a print-app

sbrunner:


asaunier:
remport? Do you mean report?
At first sight, no objection.

sbrunner:
Effectively :-)

PR #1594, opened 2015-09-10
msgmerge improvements 

fredj:


fredj:
sort-by-file option is added to avoid this kind of change:
```diff
diff --git a/epfl_authgeoportail/locale/fr/LC_MESSAGES/epfl_authgeoportail-server.po b/epfl_authgeoportail/locale/fr/LC_MESSAGES/epfl_authgeoportail-server.po
index d3a3d75..9805b7b 100644
--- a/epfl_authgeoportail/locale/fr/LC_MESSAGES/epfl_authgeoportail-server.po
+++ b/epfl_authgeoportail/locale/fr/LC_MESSAGES/epfl_authgeoportail-server.po
@@ -1,11 +1,11 @@
 #: epfl_authgeoportail/templates/edit.html:60
-#: epfl_authgeoportail/templates/routing.html:60
 #: epfl_authgeoportail/templates/index.html:72
+#: epfl_authgeoportail/templates/routing.html:60
 msgid "Loading message"
 msgstr ""
 
 #: epfl_authgeoportail/templates/edit.html:74
-#: epfl_authgeoportail/templates/routing.html:74
 #: epfl_authgeoportail/templates/index.html:91
+#: epfl_authgeoportail/templates/routing.html:74
 msgid "search tip message"
 msgstr ""
```

Where only the order changes

ochriste:
+1

c2cgeoportal/scaffolds/create/+dot+gitignore_tmpl
```diff
@@ -1,4 +1,3 @@
-*~
```
sbrunner: Why removing this ?

sbrunner:
Also +1 :-)

c2cgeoportal/scaffolds/create/+dot+gitignore_tmpl
fredj: because the backup files from msgmerge are named `xxx.po~`

sbrunner: Then why don't we want to ignore them?

fredj: because the wont be created anymore by using the `backup=none` option

sbrunner: OK, thanks for the full explain :-)

PR #1601, opened 2015-09-14
Stop using deprecated pyramid.i18n.get_locale_name

fredj:
See http://docs.pylonsproject.org/projects/pyramid//en/latest/api/i18n.html#pyramid.i18n.get_locale_name


asaunier:
Makes sense

PR #1602, opened 2015-09-15
Use IE11 document mode

ger-benjamin:
~~Use IE11 document mode when available (currently IE11 uses IE10 document mode to render the geoportal).~~

~~After discussion with Stephane, we won't use the notation: ``<meta http-equiv="X-UA-Compatible" content="IE=EmulateIE7; IE=EmulateIE9; IE=edge">`` because this is less understandable.~~

Use IE11 document mode instead of IE10 document mode to avoid some bugs with ext. IE10 will use IE9 document mode.


sbrunner:
Event if I don't like to manipulate the meta X-UA-Compatible I think that will be a good solution ... merge :-)

PR #1604, opened 2015-09-15
Load less file in debug mode

fredj:


sbrunner:
Looks good :-)

fredj:
the path to `less.min.js` is probably wrong (`node_modules/less/dist/less.min.js`). `request.static_url` should be used

sbrunner:
Right, probably with node_modules (https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/__init__.py#L278)

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -498,10 +497,6 @@ $(APP_OUTPUT_DIR)/build.min.css: $(LESS_FILES) .build/node_modules.timestamp
 	mkdir -p $(dir $@)
 	./node_modules/.bin/lessc --clean-css $(PACKAGE)/static-ngeo/less/$(PACKAGE).less $@
 
-$(APP_OUTPUT_DIR)/build.css: $(LESS_FILES) .build/node_modules.timestamp
-	mkdir -p $(dir $@)
-	./node_modules/.bin/lessc $(PACKAGE)/static-ngeo/less/$(PACKAGE).less $@
-
```
sbrunner: Just a remark, to be as we are with `cgxp` we shouldn't use the `--clean-css` when `DEVELOPMENT` is set to `TRUE`...

fredj:
`${request.static_url('epfl_authgeoportail:node_modules/less/dist/less.js')}` does not work

sbrunner:
I think that should be:
`${request.static_url('node_modules:less/dist/less.js')}`

sbrunner:
no I'm wrong ...

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
fredj: Fixed with #1614

fredj:
It's working now ...

sbrunner:
Looks good for me :-)

PR #1605, opened 2015-09-16
Update the PDF report

sbrunner:
* Update the doc doc make migration.
* Replace the `show_map` by a `map` config level.
* Use the Proxy class.
* Add the ability to use the non map attributes in the custom `spec`.
* Add the ability to have report with multiple maps with different
  background layers (can be combine with single map).

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -78,6 +78,9 @@ Version 1.6.3
 
 7. Remove the `lang_files` from the `vars/checker` in the file `vars_{{package}}.yaml`.
 
+8. In the PDF report config the show_map attribute will be removed and a map is added in the config hierarchy.
+   See the doc to update your config: http://docs.camptocamp.net/c2cgeoportal/master/integrator/pdfreport.html
+
 
 Version 1.6.2
 =============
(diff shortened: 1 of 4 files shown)


c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -78,6 +78,9 @@ Version 1.6.3
 
 7. Remove the `lang_files` from the `vars/checker` in the file `vars_{{package}}.yaml`.
 
+8. In the PDF report config the show_map attribute will be removed and a map is added in the config hierarchy.
+   See the doc to update your config http://docs.camptocamp.net/c2cgeoportal/master/integrator/pdfreport.html.
```
asaunier: http://docs.camptocamp.net/c2cgeoportal/master/integrator/pdfreport.html is broken as for now.
I would remove the "." at the end of the URL to ease copy-pasting it.

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -78,6 +78,9 @@ Version 1.6.3
 
 7. Remove the `lang_files` from the `vars/checker` in the file `vars_{{package}}.yaml`.
 
+8. In the PDF report config the show_map attribute will be removed and a map is added in the config hierarchy.
```
asaunier: Does this mean that to have no map in the report, one should simply omit the map attribute or set it to None?

asaunier:
Looks good to me.

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
sbrunner: Yes :-)

sbrunner: or having a spec.

PR #1611, opened 2015-09-16
Cleanup admin interface config

sbrunner:


asaunier:
What if you have other available functionalities than the 3 types mentioned? You must keep the config as is?

sbrunner:
Yes :-)

asaunier:
So perhaps you should be a little clearer and say that those are the default settings and that it is no use overridding them unless a custom config is required?

sbrunner:
Updated :-)

asaunier:
Thanks, please merge.

PR #1612, opened 2015-09-17
Add the full-text search to the config vars

### LINKED ISSUE #1597: [Upgrade to 1.6] FTS and unaccent
After upgrade to 1.6, the FTS with special chars doesn't work anymore:
* http://nyon-geoportail-testnew.prod.sig.cloud.camptocamp.net/dev/wsgi/fulltextsearch?limit=100&partitionlimit=30&query=ch%C3%A2teau __KO__
* http://map.nyon.ch/prod/wsgi/fulltextsearch?limit=100&partitionlimit=30&query=ch%C3%A2teau OK

I checked if unaccent extension and dictionary are installed on the new server and seems to be the case:

````
nyon_dev=# CREATE EXTENSION unaccent;
ERROR:  function "unaccent" already exists with same argument types

nyon_dev=# CREATE TEXT SEARCH CONFIGURATION fr (COPY = french);
ERROR:  duplicate key value violates unique constraint "pg_ts_config_cfgname_index"
DETAIL:  Key (cfgname, cfgnamespace)=(fr, 2200) already exists.
````

And I added to my `vars.yaml`:

`````
fulltextsearch:
    languages:
        fr: fr
````

The fts table is the same as in `1.5`.

Is there something else to do?

sbrunner:
fix #1597


fredj:
LGTM

PR #1613, opened 2015-09-17
Print config to use the right DPI from the print to mapserver

sbrunner:
Related to https://github.com/camptocamp/cgxp/pull/1011


asaunier:
OK

PR #1620, opened 2015-09-18
Wsgi no site package

sbrunner:


kalbermattenm:
Sure that this will not work on Windows... ;-)

sbrunner:
@kalbermattenm on Windows it will not work but it will not have side effect  :-)

asaunier:
No objection

PR #1622, opened 2015-09-18
Postgis 2 migration documentation for editable tables

ochriste:
added documentation how to migrate geometry column to Postgis 2 new Typmod, needed when using edition

### DIFF
--- a/doc/integrator/upgrade_application.rst
+++ b/doc/integrator/upgrade_application.rst
@@ -406,6 +406,51 @@ Once restored, set the original names back:
       ALTER TABLE <schema_name>.layer ADD CONSTRAINT <schema_name>.layer_id_fkey FOREIGN KEY (id) REFERENCES <schema_name>.treeitem(id);
       ALTER TABLE <schema_name>.layer DROP CONSTRAINT <schema_name>.layertmp_id_fkey;
 
+Edition
++++++++
+
+If you migrate editable tables, you need to modify the geometry data types to
+match Postgis 2 new Typmod.
+
+Example for a layer with ``Point`` geometries and a 21781 projection:
+
+   .. code:: sql
+
+      ALTER TABLE my_table ALTER COLUMN geom SET DATA TYPE geometry(Point, 21781);
+
+To help doing it on several tables at once, here is a function and an example
+of usage:
+
+   .. code:: sql
+
+      CREATE OR REPLACE FUNCTION migrategeomtopostgis2(_tablename text, _geomcolumn text, _geomtype text, _srid int) RETURNS void AS $$
+      DECLARE
+      _cleangeomtype text;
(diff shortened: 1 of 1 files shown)


doc/integrator/upgrade_application.rst
```diff
@@ -406,6 +406,36 @@ Once restored, set the original names back:
...
+Edition
++++++++
+
+If you migrate editable tables, you need to modify the geometry data types to match Postgis 2 new Typmod.
+
+Example for a layer with ``Point`` geometries and a 21781 projection:
+
+   .. code:: sql
+
+      ALTER TABLE my_table ALTER COLUMN geom SET DATA TYPE geometry(Point, 21781)
```
ger-benjamin: Last ';' is missing.

doc/integrator/upgrade_application.rst
```diff
@@ -406,6 +406,36 @@ Once restored, set the original names back:
...
+To help doing it on several tables at once, here is a function and an example of usage:
+
+   .. code:: sql
+
+      CREATE OR REPLACE FUNCTION migrategeomtopostgis2(_tablename text, _geomcolumn text, _geomtype text, _srid int) RETURNS void AS $$
+      DECLARE
+      _cleangeomtype text;
+      BEGIN
+      _cleangeomtype := initcap(lower(_geomtype));
+      EXECUTE ' ALTER TABLE ' || _tablename || ' ALTER COLUMN ' || _geomcolumn || ' SET DATA TYPE geometry(' || _cleangeomtype || ', ' || _srid || ')';
```
sbrunner: Shouldn’t we add the number of dimensions?

sbrunner:
Shouldn’t we also needs to clean-up the constrains?

doc/integrator/upgrade_application.rst
ochriste: afaik the method doesnt accept dimension as parameter, the dimensions constraint are automatically set based on the geometry type
and I assume the geometry type is correct, otherwise it would be too complicated (and Im not going to support weird cases)

ochriste:
the existing constraint cause no problem

sbrunner:
No but they are not needed anymore then they can be removed ...

ochriste:
can the constraint name be recovered automatically ?

sbrunner:
When we create a geometry column the constrains has an name based on the table and column name, than we can know the default name, if it's not the case we will not do anything.

doc/integrator/upgrade_application.rst
sbrunner: Ok :-)

ochriste:
added the constraints suppression

sbrunner:
Many thanks, for me that's looks good to be merged :-)

PR #1625, opened 2015-09-22
dont use schema name in renamed values

ochriste:
postgres complain otherwise

morges_migrate=# ALTER INDEX main.layer_pkey RENAME TO main.layertmp_pkey;
ERROR:  syntax error at or near "."


sbrunner:
Thanks, merge :-)

PR #1629, opened 2015-09-23
User on ngeo interfaces

ger-benjamin:


fredj:
LGTM

sbrunner:
Why do you need that?

I don't like to add the user, it can brake the cash system, But I think that ok here :-)

fredj:
The gcxp also has the user value

ger-benjamin:
Same needs as in the viewer.js. To do things like https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/create/%2Bpackage%2B/templates/viewer.js_tmpl#L307

But if you have a better idea, it is welcome :-) (that can be temporary, for dev purpose...)

sbrunner:
OK, thanks, merge :-)

PR #1631, opened 2015-09-24
dont use schema name in renamed values

ochriste:
same as previous PR, but this time after restoring the db (sorry, I forgot to do it in my previous PR)


sbrunner:
Thanks, merge :-)

PR #1632, opened 2015-09-24
Add validate geometry documentation

sbrunner:


doc/integrator/editing.rst
```diff
@@ -34,3 +34,16 @@ Limitations
 -----------
 
 At the moment only fields of type ``datetime`` and ``date`` are supported. ``time`` is NOT supported.
+
+Validation
+----------
+
+It's possible to validate the validity of committed geometry, to activate that you
```
asaunier: I think that in a written doc, it is better to avoid spoek forms such as "it's". I would rather write:

Checking the validity of a (commited => really needed?) geometry may be enabled using the following configuration in ``vars_<project>.yaml``:

sbrunner: Nice, thanks :-)

asaunier:
Please merge

PR #1633, opened 2015-09-24
Copy the user database before remove it

sbrunner:
* Move the script from post-restore-database to pre-restore-database
* Get the Apache stop from the default script
* Cleanup

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -131,6 +131,18 @@ Version 1.6.3
 
      configfile = "${directory}/${'development' if development == 'TRUE' else 'production'}.ini"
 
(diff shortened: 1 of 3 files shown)


c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -131,6 +131,13 @@ Version 1.6.3
 
      configfile = "${directory}/${'development' if development == 'TRUE' else 'production'}.ini"
 
+12. On deploy the post-restore-code will be moved to pre-restore-code and deeply modified.
+    It you don't modify it just get the new version:
```
elemoine: "If you did not modify it you can just copy the new version with the following:"

sbrunner: Thanks :-)

ger-benjamin:
Not tested but syntactically correct.

PR #1635, opened 2015-09-24
Set click pixel tolerance to 5px, be able to accept click and longpress

### LINKED ISSUE #1201: Difficulty to query with Chrome
After some tests with Chrome on an Android Tablet, I noticed that it's very hard to make a GetFeature, even after update to `1.5.3dev-20150126`.

In fact, when "clicking" on the map, the map moves so the GetFeature is not fired. I tested with the native Android browser, the map doesn't move when clicking on it so it's more much simple to make a query.

I don't know if it's related to C2Cgeoportal or OpenLayers.

@kalbermattenm @gnerred Did you noticed the same behaviour?

sbrunner:
Fix #1201

The in the `<package>/static/mobile/config.js` file `App.queryMode` should be set to "click" or "both".

### DIFF
--- a/c2cgeoportal/scaffolds/create/+package+/static/mobile/config.js_tmpl
+++ b/c2cgeoportal/scaffolds/create/+package+/static/mobile/config.js_tmpl
@@ -22,7 +22,7 @@ try {
 
     App.WFSTypes = jsonFormat.read('${dumps(wfs_types) | n}');
 
-    // Query mode. Can be either 'click' or 'longpress'
+    // Query mode. Can be either "click", "longpress", "both" or "none"
     App.queryMode = 'longpress';
 
     App.tilesURL = jsonFormat.read('${dumps(request.registry.settings["tiles_url"]) | n}');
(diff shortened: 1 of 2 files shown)


c2cgeoportal/scaffolds/create/+package+/static/mobile/config.js_tmpl
```diff
@@ -22,7 +22,7 @@ try {
 
     App.WFSTypes = jsonFormat.read('${dumps(wfs_types) | n}');
 
-    // Query mode. Can be either 'click' or 'longpress'
+    // Query mode. Can be either "click", "longpress", "booth" or "none"
```
elemoine: What's "booth"?

ochriste: both

sbrunner: :-)

sbrunner: Fixed...

elemoine:
Looks good to me.

PR #1647, opened 2015-09-25
Use Angular templatecache also in debug mode

elemoine:
This is a backport of https://github.com/Geoportail-Luxembourg/geoportailv3/pull/848, where we decided to use the Angular template cache in non-debug mode as well as in debug mode.


elemoine:
I added another commit, which fixes a bug when partials include non-ascii characters.

fredj:
please merge

elemoine:
Thanks @fredj.

PR #1648, opened 2015-09-28
Some upgrade improvement

sbrunner:
Already done by make
Template files can be useful to debug the build and the application
Add upgrade from 1.5 to 2.0 note

PR #1649, opened 2015-09-28
Prepare to upgrade to 2.0

sbrunner:
Add missing UPGRADE_ARGS
Add arguments --index-url and --c2cgeoportal-url to be able to upgrade to version 2.0

### DIFF
--- a/c2cgeoportal/scripts/c2ctool.py
+++ b/c2cgeoportal/scripts/c2ctool.py
@@ -126,6 +126,17 @@ def _fill_arguments(command):
         help="Specify the remote branch",
         default="origin",
     )
+    parser.add_argument(
+        "--index-url",
+        help="An alternate camptocamp python package index URL",
+        default="http://pypi.camptocamp.net/internal-pypi/index/c2cgeoportal"
+    )
+    parser.add_argument(
+        "--c2cgeoportal-url",
+        help="An alternate c2cgeoportal egg/wheel URL (with arguments %(package) %(version)s)",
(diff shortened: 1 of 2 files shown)


c2cgeoportal/scripts/c2ctool.py
```diff
@@ -126,6 +126,17 @@ def _fill_arguments(command):
...
+    parser.add_argument(
+        "--index-url",
+        help="An alternate camptocamp python package index URL",
+        default="http://pypi.camptocamp.net/internal-pypi/index/c2cgeoportal"
+    )
+    parser.add_argument(
+        "--c2cgeoportal-url",
+        help="An alternate c2cgeoportal egg/wheel URL (with arguments %(package) %(version)s)",
+        default="http://pypi.camptocamp.net/internal-pypi/index/%(package)s-%(version)s.tar.gz"
+    )
```
elemoine: Why are `--index-url` and `--c2cgeoportal-url` useful?

elemoine: I know it's related to upgrading to 2.0 but I fail to understand why users would need to use an "alternate Camptocamp python package index". I don't even know what an "alternate Camptocamp python package index" is to be honest :)

elemoine: Oh I see from https://github.com/camptocamp/c2cgeoportal/pull/1648 that we store c2cgeoportal wheels in the gh-pages branch. Are we sure that's a good idea? Each c2cgeportal wheel is about 85 MB, so won't the Git repo quickly become huge because of this?

ger-benjamin:
Not tested but syntactically correct.

PR #1675, opened 2015-10-01
ngeo-related changes for the 1.6 branch

elemoine:
This PR targets the 1.6 branch. It basically includes three changes:

* `build.json` is now a template (`build.json_tmpl`), and `{{package}}` is used instead of `geoportailv3` in the file.
* https://github.com/elemoine/c2cgeoportal/commit/6b6769a9641b3d88924ab6dad5327995daaab17a is backported to use the Angular template cache in debug mode.
* Switch from Angular v1.3 to Angular v1.4, v1.4 being what ngeo relies on (although ngeo also works with 1.3), and v1.4 being what the Luxembourg project already uses.


sbrunner:
Looks good, merge :-)

PR #1684, opened 2015-10-02
Add auth that support GeoServer

sbrunner:


c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -176,6 +177,9 @@ vars:
         # external_mapserv_wfs_url parameter
         #external_mapserv_wfs_url:
 
+        # Is an geoserver backend ?
```
asaunier: Using Geoserver as backend?

sbrunner: Thanks, fixed :-)

c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -43,6 +43,7 @@ vars:
     mapfile_data_subselect: 'SELECT ST_Collect(ra.area) FROM main.restrictionarea AS ra, main.role_restrictionarea AS rra, main.layer_restrictionarea AS lra, main.treeitem AS la WHERE rra.role_id = %role_id% AND rra.restrictionarea_id = ra.id AND lra.restrictionarea_id = ra.id AND lra.layer_id = la.id AND la.name = '
     # mapserver filter on role id only
     mapfile_data_noarea_subselect: 'SELECT rra.role_id FROM {mapserver_join_tables} WHERE rra.restrictionarea_id = ra.id AND lra.restrictionarea_id = ra.id AND lra.layer_id = la.id AND la.name = '
+    geoserver: false
```
elemoine: Any reason for having two `geoserver` variable in `CONS_vars.yaml`? Isn't the one on `mapservproxy` sufficient?

c2cgeoportal/views/mapserverproxy.py
```diff
@@ -158,10 +158,17 @@ def proxy(self):
 
         role_id = None if self.user is None else \
             self.user.parent_role.id if self.external else self.user.role.id
+
+        headers = self._get_headers()
+        # Add headers for Geoserver
+        if self.settings["geoserver"] and self.user is not None:
+            headers["sec-username"] = self.user.username
+            headers["sec-roles"] = self.user.role.name
```
elemoine: There is no need to take `parent_role` into account here (as done for `role_id` above)?

elemoine:
Shouldn't we add tests with GeoServer?

c2cgeoportal/views/mapserverproxy.py
sbrunner: Good question I will have a look...

sbrunner:
I will try :-)

elemoine:
Can you confirm that the goal of this PR is to be able to use GeoServer for internal WMS layers?

sbrunner:
Yes :-)

elemoine:
At some point we mentioned being able to have internal layers from both MapServer and GeoServer (and QGIS Server) in an application. This would require a per-layer setting/flag. Would this be too complex?

c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
sbrunner: Thanks for the notice, it's effectively an error :-)

sbrunner:
It's not too complex but it is not what we need here, it will be a next step

sbrunner:
I just see that I'm on wrong branch ... => closing

elemoine:
> It's not too complex but it is not what we need here, it will be a next step

But this change adds a global variable `geoserver`, which we will need to remove when we will change to a per-layer setting. So if the extra complexity is not too high why don't we implement the general case right now?

sbrunner:
Because we needs this that don't change every things for SBB in the 1.6 branch :-)

PR #1685, opened 2015-10-02
Deploy on GitHub releases

sbrunner:
On each commit on master and x.y branches a new wheel will be added on a GitHub release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,10 +79,27 @@ after_failure:
 - .build/venv/bin/python setup.py nosetests
 
 after_success:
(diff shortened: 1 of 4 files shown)


asaunier:
No objection

elemoine:
I see duplications in the wheel name on github.io. See http://camptocamp.github.io/c2cgeoportal/index/c2cgeoportal/index.html. Is this expected?

sbrunner:
No but it's done by an old version the last one shouldn't do that :-)

elemoine:
Ok thanks.

So with this new approach we create one tag/release per push to the master branch on GitHub. We will soon end up with tons of tags/releases. I was hoping that when pushing to the master branch on GitHub we would new wheels to a specific "dev" release or something. But maybe that's not possible.

sbrunner:
Yes but for that we should create an arbitrary tag and the sources download will give something not related to the release :-(

PR #1686, opened 2015-10-02
Fix Python utf8 error on scaffolds

### LINKED ISSUE #1679: The upgrade is not accentuated letters proof
When running the upgrade using c2ctool, I encountered issues with accents.
For example, an accentuated letter in `project.yaml.mako` for `mobile_application_title` raises encoding/codec issues.

sbrunner:
fix #1679


asaunier:
Makes sense

PR #1687, opened 2015-10-05
Use jquery v2.1.4

fredj:
We don't need to use the 1.x branch because we don't support IE < 9

### DIFF
--- a/c2cgeoportal/scaffolds/create/package.json_tmpl
+++ b/c2cgeoportal/scaffolds/create/package.json_tmpl
@@ -13,7 +13,7 @@
     "angular-gettext-tools": "2.1.2",
     "bootstrap": "^3.3.0",
     "font-awesome": "4.4.0",
-    "jquery": "1.11.3",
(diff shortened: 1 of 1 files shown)


elemoine:
What about bootstrap and typeahead? Do they work with jQuery 2?

fredj:
jQuery 2.x has the same API as jQuery 1.x but does not support IE < 9

elemoine:
Cam we use the 1.9 externs then?

fredj:
Yes, see: https://github.com/google/closure-compiler/blob/master/contrib/externs/jquery-1.9.js#L18-L22

sbrunner:
For me it's OK to merge

fredj:
Thanks for the review

elemoine:
I created an ngeo PR that suggests using jQuery 2 in the ngeo examples as well. See https://github.com/camptocamp/ngeo/pull/338. In this way, there's more chance that components that work in ngeo examples also work in c2cgeoportal applications.

PR #1697, opened 2015-10-09
Security improvement

sbrunner:


tsauerwein:
Thanks, looks good!

elemoine:
And here's a nice article about protecting cookies with `HttpOnly`.

PR #1718, opened 2015-10-19
Encode only string types

sbrunner:


c2cgeoportal/scaffolds/__init__.py
```diff
@@ -162,6 +163,7 @@ def pre(self, command, output_dir, vars):
             project = load(file("project.yaml", "r"))
             if "template_vars" in project:
                 for key, value in project["template_vars"].items():
-                    vars[key] = value.encode("utf-8")
+                    if sinstance(value, string_types):
```
elemoine: `isinstance` :)

sbrunner: fixed :-)

elemoine:
Please merge.

pgiraud:
I checked what the `six` package is. It seems to be related to python 2 & 3 compatibility.
Is c2cgeoportal supposed to be run with python3?
If not, the following test would be more relevant:
```python
if isinstance(value, basestring):
```

sbrunner:
Actually no but we should prepare the migration

PR #1720, opened 2015-10-20
Better TileCloud chain integration

sbrunner:
The files `apache/mapcache.xml` and `apache/tiles.conf` will be generated
automatically.

### DIFF
--- a/doc/integrator/make.rst
+++ b/doc/integrator/make.rst
@@ -74,29 +74,31 @@ Custom rules
 ------------
 
 In the ``<package>.mk`` file we can create some other rules.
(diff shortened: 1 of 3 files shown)


doc/integrator/make.rst
```diff
@@ -79,24 +79,25 @@ and removes it on the `clean` rule:
```
asaunier: Does this section really still use mapcache.xml as an example as said in the above sentence? All the following commands are about WMTSCapabilities.xml...

sbrunner: Thanks, fixed :-)

asaunier: Looks good to me.

PR #1721, opened 2015-10-20
Use GML3 for vector layer in PDF report service

tsauerwein:
The vector layer was requested in GML2, which caused problems with MFP (polygons were parsed as empty geometry). This PR proposes to use GML3 which is parsed correctly.


sbrunner:
Thanks, merge when Travis tests pass

PR #1722, opened 2015-10-20
One printserver in a set of site

ger-benjamin:


sbrunner:
Can you add a notes in the documentation to indicates that works for the version 2 and it should be adapted for the version 3 ?

ger-benjamin:
Somethings like that ? :
````
+
+.. note::
+
+   This system works for print v2 but must be adapted for
+   print v3 (although that's the same idea).
````

sbrunner:
yes :-)

ger-benjamin:
Thanks @sbrunner

Issue #1706, opened 2015-10-19
Headers CORS theme


sbrunner:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/views/entry.py#L1180
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/lib/caching.py#L105-L107

See also if there is set on all services that don't return private information.

PR #1723, opened 2015-10-20
Add CORS headers

ger-benjamin:
In response to #1706

### DIFF
--- a/c2cgeoportal/lib/caching.py
+++ b/c2cgeoportal/lib/caching.py
@@ -29,10 +29,12 @@
 
 
 import inspect
+import logging
 
 from dogpile.cache import compat
 from dogpile.cache.region import make_region
 
+log = logging.getLogger(__name__)
 _regions = {}
 
 
@@ -101,10 +103,65 @@ NO_CACHE = 0
 PUBLIC_CACHE = 1
 PRIVATE_CACHE = 2
 
+CORS_METHODS = "HEAD, GET, POST, PUT, DELETE"
+
+
+def set_cors_headers(service_headers_settings, request, service_name,
+                     credentials, response):
+    """
+    Handle CORS requests, as specified in http://www.w3.org/TR/cors/
+    """
+    if "Origin" not in request.headers:
+        return  # Not a CORS request if this header is missing
+    origin = request.headers["Origin"]
+
+    if not service_headers_settings or \
+            "access_control_allow_origin" not in service_headers_settings:
+        log.warning("CORS query not configured for service=%s", service_name)
+        return
+
+    preflight = request.method == "OPTIONS"
+    if preflight and "Access-Control-Request-Method" not in request.headers:
+        log.warning("CORS preflight query missing the " +
+                    "Access-Control-Request-Method header", service_name)
+        return
+
+    allowed_origins = service_headers_settings["access_control_allow_origin"]
+    if origin not in allowed_origins:
+        if "*" in allowed_origins:
+            origin = "*"
+            credentials = False  # Force no credentials
+        else:
+            log.warning("CORS query not allowed for origin=%s, service=%s", origin,
+                        service_name)
+            return
+
+    response.headers.update({
+        "Access-Control-Allow-Origin": origin,
+        "Access-Control-Allow-Methods": CORS_METHODS
+    })
+
+    if preflight:
+        max_age = service_headers_settings.get("access_control_max_age", 3600)
+        response.headers["Access-Control-Max-Age"] = str(max_age)
+
+    if credentials:
+        response.headers["Access-Control-Allow-Credentials"] = "true"
+
+    requested_headers = \
+        request.headers.get("Access-Control-Request-Headers", False)
+    if requested_headers:
+        # For the moment, we allow all requested headers
+        response.headers["Access-Control-Allow-Headers"] = requested_headers
+
+    # If we start using headers in responses, we'll have to add
+    # Access-Control-Expose-Headers
+
 
 def set_common_headers(
         request, service_name, cache,
-        response=None, add_cors=False, vary=False, content_type=None):
+        response=None, credentials=True, add_cors=False, vary=False,
+        content_type=None):
     if response is None:
         response = request.response
 
@@ -136,13 +193,9 @@ def set_common_headers(
             else:
                 response.cache_control.no_cache = True
 
-        if add_cors and service_headers_settings and \
-                "access_control_allow_origin" in service_headers_settings:
-            cors = service_headers_settings["access_control_allow_origin"]
-            response.headers.update({
-                "Access-Control-Allow-Origin": cors,
-                "Access-Control-Allow-Headers": "X-Requested-With, Content-Type"
-            })
+        if add_cors:
+            set_cors_headers(service_headers_settings, request, service_name,
+                             credentials, response)
 
     if vary:
         response.headers["Vary"] = "Accept-Language"
--- a/c2cgeoportal/lib/caching.py
+++ b/c2cgeoportal/lib/caching.py
@@ -120,23 +120,29 @@ def set_common_headers(
     else:  # pragma: nocover
         raise "Invalid cache type"
 
-    if cache != NO_CACHE:
-        max_age = request.registry.settings["default_max_age"]
+    if hasattr(request, "registry"):
+        headers_settings = request.registry.settings.get("headers", {})
+        service_headers_settings = headers_settings.get(service_name)
 
-        settings = request.registry.settings.get("cache_control", {})
-        if service_name in settings and "max_age" in settings[service_name]:
-            max_age = settings[service_name]["max_age"]
+        if cache != NO_CACHE:
+            max_age = request.registry.settings["default_max_age"]
 
-        if max_age != 0:
-            response.cache_control.max_age = max_age
-        else:
-            response.cache_control.no_cache = True
+            if service_headers_settings and \
+                    "cache_control_max_age" in service_headers_settings:
+                max_age = service_headers_settings["cache_control_max_age"]
 
-    if add_cors:
-        response.headers.update({
-            "Access-Control-Allow-Origin": "*",
-            "Access-Control-Allow-Headers": "X-Requested-With, Content-Type"
-        })
+            if max_age != 0:
+                response.cache_control.max_age = max_age
+            else:
+                response.cache_control.no_cache = True
+
+        if add_cors and service_headers_settings and \
+                "access_control_allow_origin" in service_headers_settings:
+            cors = service_headers_settings["access_control_allow_origin"]
+            response.headers.update({
+                "Access-Control-Allow-Origin": cors,
+                "Access-Control-Allow-Headers": "X-Requested-With, Content-Type"
+            })
 
     if vary:
         response.headers["Vary"] = "Accept-Language"
--- a/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
@@ -115,6 +115,10 @@ vars:
             The GeoMapfish team
         smtp_server: smtp.example.com
 
+    headers:
+        mapserver:
+            access_control_allow_origin: "*"
+
 update_paths:
 - authtkt
 - checker
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -966,7 +966,7 @@ class Entry(object):
         return d
 
     def get_ngeo_index_vars(self, vars={}):
-        set_common_headers(self.request, "ngeo_index", NO_CACHE)
+        set_common_headers(self.request, "ngeo_index", NO_CACHE, add_cors=True)
 
         vars.update({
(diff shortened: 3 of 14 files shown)


c2cgeoportal/views/entry.py
```diff
@@ -876,7 +876,7 @@ def get_cgxp_index_vars(self, templates_params={}):
         except:  # pragma: no cover
             d["mobile_url"] = None
 
-        set_common_headers(self.request, "cgxp_index", NO_CACHE)
+        set_common_headers(self.request, "cgxp_index", NO_CACHE, add_cors=True)
```
sbrunner: no CORS on index page ...

c2cgeoportal/views/entry.py
```diff
@@ -932,14 +932,14 @@ def get_cgxp_viewer_vars(self):
             d["permalink_themes"] = json.dumps(permalink_themes.split(","))
 
         set_common_headers(
-            self.request, "cgxp_viewer", PRIVATE_CACHE,
+            self.request, "cgxp_viewer", PRIVATE_CACHE, add_cors=True,
```
sbrunner: no cors on viewer.js page

c2cgeoportal/views/entry.py
```diff
@@ -962,7 +962,9 @@ def mobile(self):
         """
         View callable for the mobile application's index.html file.
         """
-        set_common_headers(self.request, "sencha_index", NO_CACHE)
+        set_common_headers(
+            self.request, "sencha_index", NO_CACHE, add_cors=True
```
sbrunner: no cors on mobile page

c2cgeoportal/views/entry.py
```diff
@@ -1080,7 +1082,7 @@ def mobileconfig(self):
                 })
 
         set_common_headers(
-            self.request, "sencha_config", PRIVATE_CACHE,
+            self.request, "sencha_config", PRIVATE_CACHE, add_cors=True,
```
sbrunner: no cors on mobile page...

c2cgeoportal/views/entry.py
```diff
@@ -1141,7 +1143,7 @@ def xapijs(self):
 
     @view_config(route_name="apihelp", renderer="api/apihelp.html")
     def apihelp(self):
-        set_common_headers(self.request, "apihelp", NO_CACHE)
+        set_common_headers(self.request, "apihelp", NO_CACHE, add_cors=True)
```
sbrunner: no cors on api help pages (also xapi)

c2cgeoportal/views/entry.py
```diff
@@ -1219,7 +1221,9 @@ def loginform403(self):
         if self.request.authenticated_userid:
             return HTTPForbidden()  # pragma: no cover
 
-        set_common_headers(self.request, "loginform403", NO_CACHE)
+        set_common_headers(
+            self.request, "loginform403", NO_CACHE, add_cors=True
+        )
```
sbrunner: no cors on html pages, also next one

c2cgeoportal/views/exportcsv.py
```diff
@@ -62,6 +62,6 @@ def echo(request):
         name.replace(" ", "_"), csv_extension
     )
     return set_common_headers(
-        request, "csvecho", NO_CACHE,
+        request, "csvecho", NO_CACHE, add_cors=True,
```
sbrunner: echo like cors is not useful ..

ger-benjamin:
(please, wait before review)

sbrunner:
Looks good for me :-)

elemoine:
How did you decide where to use CORS and where not to use CORS?

ger-benjamin:
I always use CORE except:
- In case of the service return privates informations.
- If this is "html services", or services used near-obviously only by the same server.

elemoine:
I am a little unsure (and nervous) about enabling CORS for so many things. Are we clear on the implications? Are we sure we're not opening breaches?

c2cgeoportal/views/entry.py
```diff
@@ -1276,11 +1277,12 @@ def logout(self):
         return set_common_headers(
             self.request, "logout", NO_CACHE,
             response=Response("true", headers=headers),
+            add_cors=True
         )
 
     @view_config(route_name="loginchange", renderer="json")
     def loginchange(self):
-        set_common_headers(self.request, "loginchange", NO_CACHE)
+        set_common_headers(self.request, "loginchange", NO_CACHE, add_cors=True)
```
elemoine: This means a CSRF could now be used to change the user's password. Same thing for loginresetpassword.

ger-benjamin:
I understand and I'm not competent in the domain of the security. But I really don't see problems to open services like "themes", "fulltextsearch" or even "cgxp_index".

But yeah, I'm a little bit nervous too with the "login-related" services...

c2cgeoportal/views/entry.py
sbrunner: I don't understand the relation between CORS and CSRF to change the password eventually about reading the result but not change the password.
For me we already have an issue...
And if we don't do that we cant do an ngeo example to demo server ... (for the concerned functionalities)
I am wrong?

elemoine: Maybe CSRF is not the correct term to use. I am not sure. BUT I do think enabling CORS creates a breach here. This makes it possible to create a page that can change the user's password by sending an Ajax request to GeoMapFish's loginchange web service. No?

elemoine:
> But I really don't see problems to open services like "themes", "fulltextsearch" or even "cgxp_index".

I am not a security expert either. But I think enabling CORS makes it possible for an attacker to make cross-site Ajax requests to the "themes" or "fulltextsearch" web services with the user's credentials, and therefore get access to unauthorized data.

c2cgeoportal/views/entry.py
sbrunner: For me the break is already open with or without CORS...

sbrunner:
As I understand this is effectively correct ...

c2cgeoportal/views/entry.py
elemoine: Yep, you may be right.

sbrunner:
A solution can be that we list a set of domains that is allows to read the services ...
To fill the `Access-Control-Allow-Origin` with the list instance of with a `*`.

elemoine:
> A solution can be that we list a set of domains that is allows to read the services ...
> To fill the Access-Control-Allow-Origin with the list instance of with a *.

Yes exactly. We should be more restrictive.

ger-benjamin:
Ok, thanks @elemoine, @sbrunner I'll work on that.

ger-benjamin:
Something like my latest commit ? If the principle suits you, I'll add some documentation here: http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/security.html

c2cgeoportal/lib/caching.py
```diff
@@ -120,21 +120,25 @@ def set_common_headers(
     else:  # pragma: nocover
         raise "Invalid cache type"
 
+    settings = request.registry.settings.get("headers", {})
```
sbrunner: Shouldn't you add `.get(service_name, {})` and remove it later?

sbrunner:
You should also add `headers` here https://github.com/camptocamp/c2cgeoportal/blob/add_cors_headers/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl#L300

c2cgeoportal/lib/caching.py
```diff
@@ -120,21 +120,25 @@ def set_common_headers(
...
+            max_age = settings[service_name]["cache_control_max_age"]
 
         if max_age != 0:
             response.cache_control.max_age = max_age
         else:
             response.cache_control.no_cache = True
 
-    if add_cors:
+    if add_cors and service_name in settings and \
+        "cors" in settings[service_name]:
```
elemoine: I think we should use `access_control_allow_origin` as the setting name.

sbrunner:
+1 for the doc in security

c2cgeoportal/lib/caching.py
elemoine: I'd rename this to `headers_setting`.

c2cgeoportal/lib/caching.py
```diff
@@ -120,21 +120,25 @@ def set_common_headers(
...
+
     if cache != NO_CACHE:
         max_age = request.registry.settings["default_max_age"]
 
-        settings = request.registry.settings.get("cache_control", {})
-        if service_name in settings and "max_age" in settings[service_name]:
-            max_age = settings[service_name]["max_age"]
+        if service_name in settings and \
+            "cache_control_max_age" in settings[service_name]:
+            max_age = settings[service_name]["cache_control_max_age"]
```
elemoine: Will this change require notes in the changelog?

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
```diff
@@ -115,6 +115,10 @@ vars:
             The GeoMapfish team
         smtp_server: smtp.example.com
 
+    headers:
+        mapserver:
+            cors: *
```
elemoine: Do we really need `*` by default for mapserver?

c2cgeoportal/lib/caching.py
ger-benjamin: Right ! I'll do that

c2cgeoportal/lib/caching.py
elemoine: Does the `add_cors` argument still make sense? It seems silly to add `add_cors=True` everywhere.

sbrunner: Actually yes, If we want to remove it we should add a `sub_service_name`argument in the function to have a finer granularity.
An than if we do it we should also be able to remove the 
`cache`, `vary` and `content_type` arguments.

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
sbrunner: Do be able to use it in the `WMSBrowser` of an other site without using the `OGCproxy` who is also a potential hole.

elemoine: > Do be able to use it in the WMSBrowser of an other site

What's the real use-case? Are we talking about API here? Something else?

sbrunner: Use from the Cartoriviera portal with the WMSBrowser the WMS from the Vaud portal without the OGSProxy...

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
```diff
@@ -115,6 +115,10 @@ vars:
             The GeoMapfish team
         smtp_server: smtp.example.com
 
+    headers_settings:
```
elemoine: I did not mean to change this to `headers_settings`. Just the variable in the Python code.

doc/integrator/security.rst
```diff
@@ -41,3 +41,17 @@ In the ``viewer.js`` files you should also remove the ``OpenLayers.ProxyHost`` c
...
+By default, only localhost can access c2c's services.
+To permit access to a specific service by an external server, you must set headers CORS (Access-Control-Allow-Origin) in your ``vars_<project>.yaml`` file.
+
+Add or modify the structure like that:
+
+.. code:: yaml
+
+    headers_settings:
+        <service_name>:
+            Access-Control-Allow-Origin: "<domain1>, <domain2>, ..."
```
elemoine: Access-Control-Allow-Origin -> access_control_allow_origin

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
elemoine: Use from the Cartoriviera portal with the WMSBrowser the WMS from the Vaud
> portal without the OGSProxy...
>


So this is a specific case. I just think that "*" should not be the default
value.


-- 
Eric Lemoine

Camptocamp France SAS
Savoie Technolac, BP 352
73377 Le Bourget du Lac, Cedex

Tel : +33 (0)4 58 48 20 34
Mail : eric.lemoine@camptocamp.com
http://www.camptocamp.com

doc/integrator/security.rst
sbrunner: +1

c2cgeoportal/lib/caching.py
```diff
@@ -120,21 +120,24 @@ def set_common_headers(
     else:  # pragma: nocover
         raise "Invalid cache type"
 
+    headers_settings = request.registry.settings.get("headers_settings", {})
+    service_headers_settings = headers_settings.get(service_name)
```
elemoine: Don't you want to use `.get(service_name, {})` here?

sbrunner:
I think that you should do a rebase for the tests...

ger-benjamin:
(The next week I'll be in holiday (1 week). Please merge or modify if necessary).

sbrunner:
Looks good for me, thanks :-)

PR #1726, opened 2015-10-20
Fix the Alembic branches after merge 1.6 to master

sbrunner:


asaunier:
No objection

elemoine:
Don't you want to remove the "debug" commit?

sbrunner:
@elemoine effectively also for the 'Use the right c2cgeoportal version' we shouldn't need it :-)

PR #1727, opened 2015-10-21
Convert debug print attribute to boolean

sbrunner:


elemoine:
Did you mean to change `en.js` and `fr.js` as part of this PR?

elemoine:
Does this change require notes in the changelog file?

sbrunner:
For me no because it's related to the current print configuration...

elemoine:
> For me no because it's related to the current print configuration...

Ok, just wanted to make sure.

How my questions related to the changes to the lang files?

sbrunner:
Not directly , and I see an error ..
I just separate the commit, fix the error, and update the pull request :-)

asaunier:
No objection from my perspective. @elemoine is it OK for you?

PR #1731, opened 2015-10-21
The use_security_metadata is no more used in the code

sbrunner:


asaunier:
Makes sense, please merge.

PR #1740, opened 2015-10-27
Make sure mapcache.xml is generated before creating tiles.conf

asaunier:


sbrunner:
What's the issue ?

asaunier:
See https://github.com/camptocamp/jura_c2cgeoportal/issues/23#issuecomment-151168072 and the following comments.

sbrunner:
To generate only one file they will do `make ... apache/tiles.conf` it's a strange behaviour for me ...

Then why not but then `apache/mapcache.xml` should be removed here: https://github.com/camptocamp/c2cgeoportal/blob/mapcache-tilesconf/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl#L41

asaunier:
They don't want to generate a single file. They run the standard upgrade commands. But the script fails because mapcache.xml is not existing when trying to generate tiles.conf.
That's why I suggest to add mapcache.xml as the prerequisites for generating tiles.conf.

sbrunner:
It don't fail on generating the `tiles.conf` but on doing the Apache graceful, but never-mind I understand this dependence, but now we don't need to add  `apache/mapcache.xml` here: https://github.com/camptocamp/c2cgeoportal/blob/mapcache-tilesconf/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl#L41 ...

asaunier:
You're right, the problem is that Apache is restarted before creating mapcache.xml.

> now we don't need to add apache/mapcache.xml here

Indeed, I have updated the PR accordingly.

sbrunner:
Looks good, merge :-)

PR #1743, opened 2015-10-28
Fixed box query for GeoServer

pvalsecc:
Geoserver uses namespaces for representing the different workspaces.
Because of that, one cannot remove it.

I took the opportunity to start the GeoServer backend documentation.

Now based on 1.6, as requested.


sbrunner:
Thanks, merge :-)

PR #1748, opened 2015-10-28
Fix typo in the documentation

sbrunner:


asaunier:
"See doc deference" => deference?
Else no objection.

PR #1752, opened 2015-10-29
Interprete closure_library_path

arnaud-morvan:


arnaud-morvan:
c2c-template --vars vars_test.yaml --engine mako --files jsbuild/app.cfg.mako

Raise an error:

    Error: Cannot find module 'openlayers/node_modules/closure-util'

npm install does'nt seems to be called during build.

sbrunner:
Yes it's normal the current test don't use the ngeo interface, I should do a fix in `c2c.tempate` to be able to don't have an error.

c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -303,3 +303,7 @@ interpreted:
     - instanceid
     - apache_entry_point
     - development
+    node:
+        vars:
+        - closure_library_path
+        cmd: ["node", "-e"]
```
sbrunner: You should add:
`ignore_error: True`
and do a rebase :-)

ger-benjamin:
ping @arnaud-morvan 
What about this PR ?

arnaud-morvan:
We use node modules in aeroportdelyon with GMF 1.6, so I'd like to have it in coming releases.
I've added ignore_errors and rebased on 1.6.

sbrunner:
Looks good for me :-)

PR #1753, opened 2015-10-29
Add missing primary key on layergroup_treeitem

### LINKED ISSUE #1738: [Upgrade to 1.6] Missing PK for layergroup_treeitem
I think the column `id` should be PK for table `main.layergroup_treeitem`

sbrunner:
fix #1738


asaunier:
Makes sense

PR #1754, opened 2015-10-29
Test the accent in the template generation

sbrunner:


asaunier:
You mean that having accent in the mobile application title doesn't make the scaffolding crash?

sbrunner:
Yes, fixed in #1686

asaunier:
OK, looks good to me

PR #1758, opened 2015-11-02
Add an "export GPX/KML" view

elemoine:
This PR adds a view for exporting GPX and KML documents. This view acts as an "echo" service, setting the `Content-Disposition` header on the response to make the browser open the "save as" window.

On the client side a form with a standard submission (w/o Ajax) is to be used.

The exportgpxkml and exportcsv views being quite similar I placed them in the same view file, namely `export.py`.


elemoine:
coreralls reports that coverage decreased. I added tests for the code I added. What else could I do? Maybe I should add tests for all the `HTTPBadRequest` cases in the `exportgpxkml` view, but that would be a bit silly.

elemoine:
And coveralls reports that coverage changed in `c2cgeoportal/__init__.py`, which is 100% covered before and after my changes :) So I guess we can just bypass coveralls in that case.

c2cgeoportal/views/export.py
```diff
@@ -40,7 +40,7 @@
 
 
 @view_config(route_name="csvecho")
-def echo(request):
+def exportcsv(request):
```
sbrunner: I don't like to use different name for function and route.
Can we put everywhere `exportcsv`?

sbrunner:
> Maybe I should add tests for all the HTTPBadRequest cases in the exportgpxkml view, but that would be a bit silly.
For me too, than you can add a `# pragma: nocover` on the if :-)

elemoine:
> For me too, than you can add a # pragma: nocover on the if :-)

That doesn't sound better to me.

elemoine:
I've finally decided to add more tests. I'll merge if Travis likes it.

elemoine:
coveralls is happier now.

c2cgeoportal/views/export.py
elemoine: I missed that comment. Sorry.

I didn't want to change the ```route_name``` to avoid breaking existing applications that use `${request.route_url('csvecho')}` in mako templates.

Please tell me if you'd like to reconsider this.

sbrunner: Good point :-)

PR #1760, opened 2015-11-03
Fix the querier documentation

asaunier:


doc/integrator/querier.rst
```diff
@@ -25,16 +25,16 @@ The web service configuration is done in the ``vars_<project>.yaml`` file:
                         separator: ","
 
 The ``dbsession: "<session name>"`` in the ``enum`` is just a shortcut
-if almost of the session are similar in the layers. If the dbsession isn't
-defined we use the main DB session.
+if almost of the session are similar in the layers. If ``dbsession`` is not
```
asaunier: I don't understand what "if almost if the session..." means!

doc/integrator/querier.rst
```diff
@@ -25,16 +25,16 @@ The web service configuration is done in the ``vars_<project>.yaml`` file:
...
+defined, the main DB session is used.
 
 The ``table: "<[schema.]table name>"`` in the layer is just a shortcut
-if almost of the attribute of the layer are in the same table.
-Each attributes requires a table.
+if almost of the attributes of the layer are in the same table.
```
asaunier: Not clear here as well what "if almost of the attributes" means...

doc/integrator/querier.rst
```diff
@@ -25,16 +25,16 @@ The web service configuration is done in the ``vars_<project>.yaml`` file:
...
+Each attribute requires a table.
 
-If the ``column_name`` isn't defined we use the attribute name.
+If the ``column_name`` is not defined, the attribute name is used.
 
-If the ``separator`` is defined we consider the column as a list of values.
+If the ``separator`` is defined, the column is considered as a list of values.
```
asaunier: What is "the column"? Do you mean ":" (colon)? https://en.wikipedia.org/wiki/Colon_(punctuation)

doc/integrator/querier.rst
asaunier: Maybe something like:

``table: "<[schema.]table name>"`` may be used at the layer level as a the default table where following attributes may be found. It can be overridden at the attribute level.

doc/integrator/querier.rst
asaunier: I guess that this means that one may specify the dbsession (but with what value??) as the default session and override it for each layer if needed.

sbrunner: we can fill the Sessions define here:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/models.py#L69-L71
and it's used here:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/views/layers.py#L397

doc/integrator/querier.rst
sbrunner: +1

doc/integrator/querier.rst
sbrunner: «the column» => «the column, values» isn't better ?

... separate with the `separator` ?

sbrunner:
You add a trailing space :-)
https://travis-ci.org/camptocamp/c2cgeoportal/jobs/89022464#L1613

asaunier:
> You add a trailing space :-)

Thanks, updated.

sbrunner:
Looks good :-)

PR #1761, opened 2015-11-04
Security fixes

pvalsecc:
Added a protection against CSRF attacks.
Validate the query referer against a list of authorized referers.

The best solution would be to add a token only the main page
knows when the user logs in (via a cookie). Adding a header to all the requests
would mean a lot of changes:

* Extjs: ``Ext.Ajax.setDefaultHeaders``
* OL: ``OpenLayers.Request.DEFAULT_CONFIG.headers``
* JQuery: ``$.ajaxSetup({headers: {...}})``
* Angular: ``$httpProvider.defaults.headers``

Fixed details in edit.js and view.js    
The edit page was failing because:
* EVENTS was undefined
* WMS and WMTS where undefined
    
The view page was always showing the restricted warning even after
logging in because:
* view.js was cached
* non existant default theme

### DIFF
--- a/doc/integrator/security.rst
+++ b/doc/integrator/security.rst
@@ -58,3 +58,17 @@ Available services are:
 - profile
 - raster
 - error
+
+Authorized referers
+-------------------
+
+To mitigate `CSRF <https://en.wikipedia.org/wiki/Cross-site_request_forgery>`_
+attacks, the server validates the referer against a list of authorized referers.
+
+By default, only the pages coming from the server are allowed. You can change
+that list by adding an ``authorized_referers`` list in your
+``vars_<project>.yaml`` file.
+
+This solution is not the most secure (some people have browser extensions that
+reset the referer), but that is the easiest to implement with all our different
+JS frameworks.
--- a/c2cgeoportal/__init__.py
+++ b/c2cgeoportal/__init__.py
@@ -353,25 +353,51 @@ def locale_negotiator(request):
     return lang
 
 
-def get_user_from_request(request):
-    """ Return the User object for the request.
-
-    Return ``None`` if user is anonymous or if it does not
-    exist in the database.
+def _match_url_start(ref, val):
     """
-    from c2cgeoportal.models import DBSession, User
+    Checks that the val URL starts like the ref URL.
+    """
+    ref_parts = ref.rstrip("/").split("/")
+    val_parts = val.rstrip("/").split("/")[0:len(ref_parts)]
+    return ref_parts == val_parts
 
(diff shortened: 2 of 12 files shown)


tsauerwein:
Thanks for working on this. This is definitely better than nothing.

doc/integrator/security.rst
```diff
@@ -31,3 +31,26 @@ In the ``viewer.js`` files you should also remove the ``OpenLayers.ProxyHost`` c
...
+
+To mitigate `CSRF <https://en.wikipedia.org/wiki/Cross-site_request_forgery>`_
+attacks, the server validates the referer against a list of authorized referers.
+
+By default, only the pages coming from the server are allowed. You can change
+that list by adding an ``authorized_referers`` list in your
+``vars_<project>.yaml`` file.
+
+This solution is not the most secure (some people have browser extensions that
+resets the referer), but that is the easiest to implement with all our different
```
asaunier: resets => reset

pvalsecc:
There is a problem with Travis:
```
$ travis/deploy.sh
== Build the doc ==
Warning: Permanently added the RSA host key for IP address '192.30.252.128' to the list of known hosts.
Enter passphrase for key '/home/travis/.ssh/id_rsa': 
```

How do we fix this passphrase problem?

elemoine:
> There is a problem with Travis:

Did you find a way to fix that issue?

pvalsecc:
I don't know, since a few hours I have another issue that happens earlier. lingua (just upgraded) chokes on theme2fts.py. I have created a pull request for that:
https://github.com/wichert/lingua/pull/70

In the mean time, I'm considering removing that file from the list of files processed by lingua (it doesn't contain any internationalized string) or change a bit the file to work around the bug. What do you recommend?

wichert:
I'll make a new lingua release today to unblock you.

pvalsecc:
Thanks @wichert for the quick merge and release.

Now, I have another problem with Travis. Looks like build-new-project.sh is a bit mad. The first thing I don't understand is that it downloads another version of c2cgeoportal from the web instead of taking the local one. What's the point? Then I have this error:
```
ERROR [alembic.util] Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads

  FAILED: Multiple head revisions are present for given argument 'head'; please
```

ochriste:
about the alembic error, it should be fixed in the git master.
but you will have this error with a released egg.
which version of the code is being used ?

sbrunner:
Looks good, thanks

PR #1778, opened 2015-11-06
Fix Travis when building a new project.

pvalsecc:
The new project Travis was creating was not using the current c2cgeoportal
version, but an older one. This is because c2cgeoportal as been removed from
CONST_versions.txt.


ochriste:
seems ok (afaik)

sbrunner:
Thanks @pvalsecc

PR #1779, opened 2015-11-06
Fix 1.6 to master merge

sbrunner:


asaunier:
Makes sense indeed :)

pvalsecc:
It would be saner if you were on the chat. I was working on fixing the Alembic merge as well  :-(

sbrunner:
Witch chat ?

ochriste:
jabber

sbrunner:
My network is not stable an I will be disconnected :-(

pvalsecc:
Ship it!

sbrunner:
Thanks :-)

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -286,7 +286,6 @@ TEMPLATES_FILES = $(shell find $(PACKAGE)/templates -type f -print)
 
 # Templates
 TEMPLATE_EXCLUDE += .build print/templates \
-	templatecache.js.mako \
```
elemoine: Is this wanted?

sbrunner: Yes, if the file is named `templatecache.mako.js` as in 1.6, it's not needed :-)

elemoine: I don't understand. I thought `TEMPLATE_EXCLUDE` was a list a files that shouldn't be passed to `c2c.template`. And we don't want to pass `templatecache.js.mako` to `c2c.template`.

Am I not understanding this correctly?

sbrunner: Yes, if the file is named `templatecache.js.mako` but he is named `templatecache.mako.js` :-)

sbrunner: `.js` and `.mako` inversion!

elemoine: Oh ok. A bit fragile but it works.

Issue #1791, opened 2015-11-11
[Upgrade to 1.6] How to access project specific vars in Python views?


rbovard:
In a custom python view, I reached some variables defined in `config.yaml.in` with `request.registry.settings['targetemail']`

What is the synthax to get them in the new `vars_xxx.yaml`? Because now I have this error `KeyError: 'targetemail'`.

Thanks

kalbermattenm:
I think you have to explicitly define them in your make file:

https://github.com/camptocamp/sitn_c2cgeoportal/blob/1.6/sitn.mk#L12-L37

rbovard:
Thanks a lot @kalbermattenm!

PR #1794, opened 2015-11-12
Add doc about CONFIG_VARS

asaunier:
CONFIG_VARS is already described here http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/make.html#makefile-config-variables but I think we could make it even more obvious, because making config vars available in the templates or python code is a recurrent question.

### DIFF
--- a/doc/integrator/make.rst
+++ b/doc/integrator/make.rst
@@ -44,6 +44,14 @@ Vars files
 The project variables are set in the ``vars_<package>.yaml`` file,
 which extends the default ``CONST_vars.yaml``.
 
+To make such variables available to the python code, for instance using
+
+.. code:: python
+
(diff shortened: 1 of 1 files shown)


rbovard:
+1 :)

asaunier:
@rbovard I read your question at https://github.com/camptocamp/c2cgeoportal/issues/1791 :)
Actually I had the same problem a few weeks ago and though that the doc was not clear enough about that point. I hope this PR will help.

sbrunner:
Excellent, thanks @asaunier

sbrunner:
Excellent, thanks @asaunier

kalbermattenm:
Thanks @asaunier, I thought about adding some doc, but I was too lazy... (bad me...)

PR #1799, opened 2015-11-12
Update the referance to the config.yaml file

sbrunner:


asaunier:
referance => reference

PR #1807, opened 2015-11-16
clarify TILECLOUD_CHAIN disabling documentation

ochriste:
because I didn't really understand the real consequence of the existing comment before being confronted to this problem, I think a little clarification is needed.

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -4,9 +4,13 @@ This file includes migration steps for each release of c2cgeoportal.
 Version 1.6.6
 =============
 
-1. Now the files `apache/mapcache.xml` and `apache/tiles.conf` will be generated automatically,
-   then you can remove your custom build rules. To disable it set in your Makefile:
-   `TILECLOUD_CHAIN ?= FALSE`.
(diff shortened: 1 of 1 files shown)


c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -5,8 +5,12 @@ Version 1.6.6
 =============
 
 1. Now the files `apache/mapcache.xml` and `apache/tiles.conf` will be generated automatically,
```
asaunier: will be generated => are generated

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -5,8 +5,12 @@ Version 1.6.6
...
-   then you can remove your custom build rules. To disable it set in your Makefile:
-   `TILECLOUD_CHAIN ?= FALSE`.
+   then you can remove your custom build rules.
+
+   If you DON'T use tilecloud-chain and you DON'T have a `tilegeneration/config.yaml` configuration
+   file, you MUST disable it by setting the following parameter in your Makefile:
```
asaunier: Looks good. Not sure it is really needed to use capitalized words, standard case is enought is think.

asaunier:
Thanks for the changes. Please merge.

PR #1814, opened 2015-11-17
Fix the install application process

sbrunner:
Move alembic, JSTools, lingua as dev depencency


ger-benjamin:
Not tested but seems all right.

PR #1819, opened 2015-11-24
Add autentication web service documentation

sbrunner:
typo

### DIFF
--- a/doc/developer/webservices.rst
+++ b/doc/developer/webservices.rst
@@ -13,7 +13,7 @@ URL: ``.../themes``
 Parameters
 ----------
 
-* ``version``: ``1`` or ``2 ``, API version, default is ``1``.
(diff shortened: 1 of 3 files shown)


doc/developer/webservices.rst
```diff
@@ -148,6 +148,50 @@ WMTS layer:
...
+=============
+
+Login
+-----
+
+Used to login in the application.
+
+URL: ``.../login``
+
+Parameters (post form or query string):
```
tsauerwein: The password should not be sent as query string because this risks that passwords appear as plain text in log files.

sbrunner: Many thanks @tsauerwein fixed :-)

ger-benjamin:
Ready to merge.

PR #1822, opened 2015-11-25
Fix the post_deploy

sbrunner:


sbrunner:
Deploy specific -> merging

Issue #1815, opened 2015-11-19
cant execute alambic upgrade on pre-restore-database


ochriste:
the "alembic -c alembic_static.ini upgrade head" command from post-restore-database.mako cant not be executed because the project code has not been put into place when the post-restore-database is running.

the execution order is 1) database, 2) files, 3) code

https://github.com/camptocamp/deploy/blob/master/bin/deploy#L282-L295

ochriste:
after discussion with Fred, we could improve the deploy script to allow to set the execution order

sbrunner:
Or we can simply move the alembic call in the post-restore (global) ?

ochriste:
we must update the static schema before writing over the main scheme, hence we absolutely must run the alembic call in pre-restore

PR #1833, opened 2015-11-30
To update the static schema, we need to deploy the code first

ochriste:
because we need the latest alembic upgrade script to be able to do that.

this will only work with a new delpoy package, but will not cause any problem in the meantime.

cf https://github.com/camptocamp/deploy/commit/7af8edb9aaa6043e288215e90719ecf4d2c6d752


ochriste:
also https://github.com/camptocamp/c2cgeoportal/issues/1815

sbrunner:
We should add it in the changelog ...

sbrunner:
Looks good, merge :-)

PR #1836, opened 2015-11-30
Use PIP version 6.x

### LINKED ISSUE #1832: zc.recipe.egg not supported with pip >= 7
pip >= 7 cannot install `zc.recipe.egg` anymore (for the whole story see https://github.com/camptocamp/c2cgeoportal/issues/1494)

`zc.recipe.egg` is a dependency of the dev environment, because of https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/CONST_dev-requirements.txt#L4

If we use a higher version of `c2c.template`, which does not use anymore `zc.recipe.egg`, then we can go around this problem.

Thus the solution would be to use `c2c.template>=1.2.dev` in `CONST_dev-requirements.txt`

This is what was done on the Lux project: https://github.com/Geoportail-Luxembourg/geoportailv3/pull/842

sbrunner:
Install wheel only on deploy

fix #1832


pvalsecc:
LGTM

sbrunner:
Thanks :-)

PR #1838, opened 2015-12-01
Fix encoding issue in httplib when doing print request from IE11.

ochriste:
reported from https://github.com/camptocamp/jura_c2cgeoportal/issues/40


sbrunner:
Looks good, merge :-)

PR #1858, opened 2015-12-10
Update the print and make it more secure

sbrunner:


sbrunner:
@pvalsecc can you have a look?

c2cgeoportal/scaffolds/create/print/print-apps/+package+/config.yaml.mako
```diff
@@ -35,21 +35,47 @@ templates:
...
-        - !configureHttpRequests
+        - !configureHttpRequests &configureHttpRequests
             httpProcessors:
             - !mapUri
+                matchers:
+                - !dnsMatch
+                    host: ${host}
                 mapping:
                     (https?)://${host}/(.*): "$1://127.0.0.1/$2"
-            - !useHttpForHttps
```
pvalsecc: Warning, this guy is not only changing the port, but is changing the protocol as well. Your change made the local URI redirected to https://{host}:80/... apache is going to have only one response to that: meh?

sbrunner: Thanks fixed :-)

PR #1863, opened 2015-12-11
Add gxp local for fr and de translation for print panel.

ochriste:


sbrunner:
Looks good, merge :-)

PR #1868, opened 2015-12-14
Initial docker integration

pvalsecc:
We have three Docker images generated: print, mapserver and wsgi.
We are still missing the db, tilecloud and proxy images.


c2cgeoportal/scaffolds/create/mapserver/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM camptocamp/mapserver:latest
+MAINTAINER Camptocamp "info@camptocamp.com"
+
+ADD include/*.map /etc/mapserver/include/
```
sbrunner: We don't add *.map ?

pvalsecc: The bsae image I'm using want the file to be named locally mapserver.map because of this line:
ONBUILD COPY mapserver.map /etc/mapserver/

That's the reason why the makefile is copying c2cgeoportal.map to mapserver.map before creating the image.

sbrunner: Thanks @pvalsecc :-)

c2cgeoportal/scaffolds/create/mapserver/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM camptocamp/mapserver:latest
```
sbrunner: Shouldn't we create a docker folder =>  c2cgeoportal/scaffolds/create/docker/mapserver/Dockerfile ?

sbrunner:
I think that you can squash all your commit in one...

c2cgeoportal/scaffolds/create/mapserver/Dockerfile
pvalsecc: Nope, the ADD /COPY in the Dockerfile can see only the stuff in the directory where the file is located. Very weird limitation IMHO, but a limitation we cannot work around.

sbrunner:
This line:
https://github.com/camptocamp/c2cgeoportal/blob/master/.coveragerc#L6
should be: 
` c2cgeoportal/scripts/*`
and all the `  # pragma: no cover` can be removed in the script folder...

c2cgeoportal/scaffolds/create/mapserver/Dockerfile
sbrunner: OK ...

sbrunner:
Merge :-)

PR #1869, opened 2015-12-15
Stop using deprecated configuration for print logging

pvalsecc:
We use to have this error message:
WARN in ch.qos.logback.core.ConsoleAppender[standardOut] -
  This appender no longer admits a layout as a sub-component, set an encoder instead.

Upgraded the print to 3ff8727.


sbrunner:
Excellent, merge when Travis tests pass :-)

PR #1870, opened 2015-12-15
Add a new resourceproxy service, allowing to load external resources.

ochriste:
Useful when the project is configured with https and you want to include non-https content. The resources url host must be specifically added in the project configuration to be allowed to be loaded.

Needed in the Morges project.


sbrunner:
It relay a security holl, and as it's configured by default it is...

ochriste:
let me know when you have a moment to discuss about this

sbrunner:
Looks good

sbrunner:
up

PR #1873, opened 2015-12-16
Disable redirect

### LINKED ISSUE #1866: Verify that files are correctly downloaded in Makefile
The Makefile downloads a couple of files using wget (for example several JARs for MFP). If there is a redirect or if the requests fail, the target file is created anyway and the build does not fail.

There should be check that the return code is `200`.

sbrunner:
fix  #1866

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -519,22 +519,22 @@ $(OUTPUT_DIR)/templatecache.js: templatecache.mako.js $(APP_PARTIALS_FILES) .bui
 
 .build/externs/angular-1.4.js:
 	mkdir -p $(dir $@)
-	wget -O $@ https://raw.githubusercontent.com/google/closure-compiler/master/contrib/externs/angular-1.4.js
+	wget --max-redirect=0 -O $@ https://raw.githubusercontent.com/google/closure-compiler/master/contrib/externs/angular-1.4.js
 	touch $@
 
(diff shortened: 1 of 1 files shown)


c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -750,7 +750,7 @@ print/WEB-INF/lib/jasperreports-fonts-5.5.0.jar:
 
 print/WEB-INF/lib/postgresql-9.3-1102.jdbc41.jar:
 	mkdir -p $(dir $@)
-	wget http://jdbc.postgresql.org/download/postgresql-9.3-1102.jdbc41.jar -O $@
+	wget --max-redirect=0 https://jdbc.postgresql.org/download/postgresql-9.3-1102.jdbc41.jar -O $@
```
ger-benjamin: From http to https ? it's correct ?

ger-benjamin:
Read my comment then (correct it and then) LGTM

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
sbrunner: Yes because the http is redirected to the https :-)

tsauerwein:
Sorry, I missed this PR. With this change you are avoiding the redirect, but you will create an empty file and no error will be emitted. Or not?

sbrunner:
wget steal write a file :-/ just to be against Make?

sbrunner:
I will use curl ...

PR #1875, opened 2015-12-16
Use curl instance of wget, remove touch

### LINKED ISSUE #1866: Verify that files are correctly downloaded in Makefile

sbrunner:
fix #1866


pvalsecc:
All good, thanks.

PR #1892, opened 2015-12-21
Some little fix in the login views

sbrunner:
See separate commits :-)

TODO:
 * [x] add an old password field in the "change password" view
 * [x] All the login view should be less verbose => to don't offer informations to a potential hacker


sbrunner:
TODO: add an `old password` field in the "change password" view

ochriste:
lgtm

PR #1898, opened 2015-12-24
fix regexp for local tile access

ochriste:
the regexp doesnt work as expected without the .*$


sbrunner:
Effectively, then we also can remove the '^' and '$'...

ochriste:
tested ok without the '^' and '$'

sbrunner:
Thanks, merge :-)

PR #1900, opened 2016-01-06
Fix change of user with the same role.

pvalsecc:
When doing a logout and login with a different user within the same
group, the "Logged in as" text on the top right corner was still showing
the previous username.

That was because the edit.js, view.js, ... pages were having role= added
as parameter to avoid the cache, not the username. So the page was not
reloaded by the browser and the username sitting in those files was
remaining at the old value.

The fix is to add username= instead of role= to the URLs.

### DIFF
--- a/c2cgeoportal/lib/__init__.py
+++ b/c2cgeoportal/lib/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2011-2015, Camptocamp SA
+# Copyright (c) 2011-2016, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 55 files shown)


c2cgeoportal/lib/__init__.py
```diff
@@ -152,8 +152,11 @@ def __call__(self, request, elements, kw):
             from c2cgeoportal.lib.cacheversion import get_cache_version
             query["cache_version"] = get_cache_version()
 
-        if self.role and request.user and request.user.role:
-            query["role"] = request.user.role.id
+        if self.role and request.user:
```
sbrunner: I think that the `self.role` can also be removed...

pvalsecc: I don't think we should remove it. In some cases (one, actually), we don't want to invalidate the cache if the user changes identity. This one:
```python
    config.add_route(
        "layers_enumerate_attribute_values",
        "/layers/{layer_name}/values/{field_name}",
        request_method="GET",
        pregenerator=C2CPregenerator(),
    )
```

sbrunner:
Thanks, merge :-)

c2cgeoportal/lib/__init__.py
sbrunner: Effectively, thanks :-)

PR #1922, opened 2016-01-13
Disable the referer check for the admin interface

sbrunner:
This not good for security but without this the admin interface don't works ...


pvalsecc:
Ship it!

PR #1925, opened 2016-01-13
Made the JasperReports version configurable.

pvalsecc:
Took the opportunity for switch to the version that matches the print version
we are using.


tsauerwein:
:+1:

PR #1926, opened 2016-01-14
Fix the deploy on documentation change

sbrunner:
and  don't create releases automatically


asaunier:
Changed discussed with @sbrunner 
Looks good to me

PR #1933, opened 2016-01-14
Remove unexisting extern

sbrunner:


fredj:
Not sure; this extern exists: https://github.com/openlayers/ol3/blob/master/externs/closure-compiler.js

sbrunner:
Effectively, thanks @fredj

PR #1934, opened 2016-01-15
Some sourceforge URLs have redirects.

pvalsecc:
For example:
http://sourceforge.net/projects/jasperreports/files/jasperreports/JasperReports%206.1.1/jasperreports-fonts-6.1.1.jar/download


sbrunner:
Thanks, merge :-)

PR #1957, opened 2016-01-25
Upgrade print to 3.3.0.

pvalsecc:
Took the opportinity to fix a few things:
* The Jasper version changed.
* The lib jasperreports-fonts is already included in the print.
* DRY in the makefile for the cleaning of print libs.
* Moved the print Dockerfile from create to update (no reason for the user
  to change it).


sbrunner:
Looks good, merge :-)

PR #1961, opened 2016-01-25
Put the right jasper report version

### LINKED ISSUE #1902: [1.6] Print on Demo instance does not work any more
https://geomapfish-demo.camptocamp.net/1.6/

```
javax.net.ssl.SSLException: hostname in certificate didn't match: <127.0.0.1> !=
```

sbrunner:
Fix #1902


pvalsecc:
Maybe you want to define a variable with the jasper version...
Otherwise ship it.

sbrunner:
It's already don in the master branch, I prefer stay the stable branch like it :-)

PR #1963, opened 2016-01-26
Forward headers for pdfReport's proxy

pvalsecc:
Otherwise, the print won't have the correct hostname.


sbrunner:
Thanks, merge :-)

PR #1975, opened 2016-01-28
Profile CSV export - Check if layer value is defined

kalbermattenm:
This PR checks if a profile value is defined for a given layer. If not, it adds the value `-9999` (which is kind of a no data value in the raster GIS world).

I had to do so because at SITN not all raster layers have the same geographical coverage. So it happens from time to time that a value for a certain layer does not exist in the `point` dict.

This is a proposal, if anyone has a better idea, it is welcomed...

Thanks

(This PR is meant for branch 1.6, but also should be applied to the master branch)


kalbermattenm:
Hmmmm... Don't really get why Coverage failed... ??

sbrunner:
Because the line you added is not tested :-)

To make it pass you should add a test here:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/tests/test_raster.py#L170-L197

Call me if you needs help

sbrunner:
Something like:
```py
        request.params["geom"] = '{"type":"LineString",' \
            '"coordinates":[[1000000,400000],[600000, 200000]]}'
        response = profile.csv()
        self.assertEqual(response.body, """distance,dem,x,y
??? => to be filled
""")
```

kalbermattenm:
Got it...

So if I understood everything, I should add this test after line 197 (https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/tests/test_raster.py#L197) and create a new test case, right ?

sbrunner:
yes :-)

kalbermattenm:
Hmmmm... I do not get the Travis error, because it fails on a line and in a function which I did not change... ??

Are there any interferences between tests, moreover `test_raster` and `test_profile_csv` ?

sbrunner:
What about changing your new dem (dem3), probably a mix in the tests ...

kalbermattenm:
Yeah, that's what I thought... I am going to rename it to `dem4`.... So there can no be any confusion anymore...

kalbermattenm:
Done... To get Travis happy was hell ;-)

Thanks for your help @sbrunner

sbrunner:
Thanks @kalbermattenm :-)

kalbermattenm:
Just a question... @srbunner, will this be backported to master ? (I never got how and when you do that... )

sbrunner:
Yes, I regularly merge the 1.6 branches to master branches :-)

kalbermattenm:
OK, thanks for your answer !

PR #1982, opened 2016-01-29
Add check for the downloaded JARs

pvalsecc:
The places we download from are often returning HTML files instead of the
JARs we request. The leads to very obscure and hard to debug errors in the
print. So checks will serve as early detection of those problems.


ochriste:
lgtm when travis has finished

PR #1986, opened 2016-01-29
Fix the links

sbrunner:


ger-benjamin:
Thanks, merge when Travis agree

(Strange Travys error, I've relauched the build.)

PR #1990, opened 2016-02-01
Fix "alembic upgrade --sql"

pvalsecc:
When in this mode, the DB connection is a MockConnection returning None as
cursor when doing a select. So has_table is not functional.


c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/166ff2dcc48d_create_database.py
```diff
@@ -62,9 +62,9 @@ def upgrade():
     srid = context.get_context().config.get_main_option('srid')
 
     engine = op.get_bind().engine
-    if op.get_context().dialect.has_table(
-        engine, 'functionality', schema=schema
-    ):  # pragma: nocover
+    if type(engine).__name__ != "MockConnection" and \
```
sbrunner: We should ask about this ...

sbrunner:
Thanks, merge :-)

PR #1994, opened 2016-02-01
No error on application without theme

sbrunner:
With theme and layers in the permalink


ger-benjamin:
Seems alright but it is always up to date ?
If so, please merge.

PR #1996, opened 2016-02-02
Add missing attribute in the full-text search

sbrunner:


pvalsecc:
Do we really need a PR for that? Go for it!

sbrunner:
;-)

PR #2001, opened 2016-02-03
Add the coordinate even if computed number of point is 1

rmichaelis:
The problem comes when there is a small distance between most of the vertices of the linestring.
A linestring composed of 800 vertices could return only result for the first vertex.


sbrunner:
The tests should also be updated...

rmichaelis:
Tests are now updated. Travis still failing on a config issue. Not sure if I'm able to solve this problem alone.

sbrunner:
In fact the tests pass, it's a post task that failed, so merging :-)

sbrunner:
Thanks :-)

PR #2007, opened 2016-02-04
Apply the static CORS only for static paths

sbrunner:


c2cgeoportal/tests/test_cachebuster.py
```diff
@@ -57,5 +60,6 @@ def test_noreplace(self):
         request = MyRequest("/test2/123456/build.css")
         response = ctf(request)
         self.assertEqual(request.path_info, "/test2/123456/build.css")
-        self.assertEqual(response.headers["Access-Control-Allow-Origin"], "*")
-        self.assertEqual(response.headers["Access-Control-Allow-Headers"], "X-Requested-With, Content-Type")
+        # test no CORS
+        self.assertFalse("Access-Control-Allow-Origin" in response.headers)
+        self.assertFalse("Access-Control-Allow-Headers" in response.headers)
```
pvalsecc: self.assertNotIn

c2cgeoportal/lib/cacheversion.py
```diff
@@ -61,7 +61,8 @@ def __call__(self, request):
 
         response = self.handler(request)
 
-        response.headers["Access-Control-Allow-Origin"] = "*"
-        response.headers["Access-Control-Allow-Headers"] = "X-Requested-With, Content-Type"
+        if path[1] in CACHE_PATH:
```
pvalsecc: I'll need more explanations, here...

c2cgeoportal/tests/test_cachebuster.py
sbrunner: Tanks :-)

c2cgeoportal/lib/cacheversion.py
sbrunner: Here we fill this global value with the static view name (second path element) https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/__init__.py#L343-L351
Already used 7 lines above :-)

c2cgeoportal/tests/test_cachebuster.py
sbrunner: done

c2cgeoportal/lib/cacheversion.py
pvalsecc: Yes, but I don't understand why you do that only if path[1] in CACHE_PATH?

And, bonus question, why not in the previous if with exactly the same condition?

sbrunner: i.-e.:
the `request.path_info` can have the value: `"/proj-ngeo/5f5e11b65b774343945a44eea2882d5d/build/mobile.css"`
the path will have the value `['', 'proj-ngeo', ...]`
the `CACHE_PATH` will have the value `[u'static', u'project', u'proj', u'proj-ngeo']`
...
For the dynamic view We can ash for user data and credentials, than we needs something dynamic that done here: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/lib/caching.py#L109-L155

Bonus: to have the response :-)

pvalsecc:
OK, ship it

PR #2011, opened 2016-02-08
Add title for steps

rbovard:
@ybolognini It's a suggestion to use `<h3>`, tell me if it's ok for you?


ybolognini:
Thanks for suggestion. Yes, We need titles. But H3s are a taking a bit too much space. Will try something else

ybolognini:
Just commited new version with titles

PR #2018, opened 2016-02-09
Add a /version.json URL

pvalsecc:
That allows to get the version that is actually running. Returns
something like that:
```json
{
  "my_webapp": {
    "git_hash": "09941add0f77e99fdcacf85fab65a2f61db585ee",
    "version": "1.0",
    "git_tag": "None"
  },
  "c2cgeoportal": {
    "git_hash": "b359b4aab3dcbe2414e790ec6671272bd7b5dc43",
    "version": "2.0.0",
    "git_tag": "0.0.9eb138c1268e18fc3c22bc43bb75888a01d829ae-34"
  }
}
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,7 @@ setup(
             "themev1tov2 = c2cgeoportal.scripts.themev1tov2:main",
             "theme2fts = c2cgeoportal.scripts.theme2fts:main",
             "import-ngeo-apps = c2cgeoportal.scripts.import_ngeo_apps:main",
+            "gen-version = c2cgeoportal.scripts.gen_version:main",
         ],
         "pyramid.scaffold": [
             "c2cgeoportal_create = c2cgeoportal.scaffolds:TemplateCreate",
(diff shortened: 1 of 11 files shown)


travis/no-make-error.sh
```diff
@@ -1,19 +1,19 @@
...
-    cd - > /dev/null
-    exit 1
+    echo There was errors in the make:
+    cat make-out
+    rm make-out
+    exit ${RESULT}
+else
+    rm make-out
 fi
-cd - > /dev/null
```
sbrunner: All this change is wrong, this script is or avoid something like:
```make
TEST = $(shell ls unexisting)
```
witch return something in the error log but no error code ...

setup.py
```diff
@@ -113,6 +113,7 @@
             "themev1tov2 = c2cgeoportal.scripts.themev1tov2:main",
             "theme2fts = c2cgeoportal.scripts.theme2fts:main",
             "import-ngeo-apps = c2cgeoportal.scripts.import_ngeo_apps:main",
+            "gen-version = c2cgeoportal.scripts.gen_version:main"
```
sbrunner: I like to have a coma at the end to have omly one line diff on the next add :-)

sbrunner:
Just some comment otherwise that's looks good :-)

pvalsecc:
OK, all comments fixed

sbrunner:
Looks good, merge :-)

PR #2023, opened 2016-02-15
Use correct remplacement variable 

ochriste:
in the reset password message body in the .yaml scaffold and in the related documentation.


sbrunner:
Looks good, merge :-)

PR #2029, opened 2016-02-16
Extract the raster layer in the i18n strings

### LINKED ISSUE #1904: Add raster Layers to i18n strings
Update the Lingua extractors:
https://github.com/camptocamp/c2cgeoportal/blob/master/setup.py#L124-L127

sbrunner:
fix #1904


ger-benjamin:
Seems alright but I've not tested and I'm not enough confident with Makefiles...
But please merge if you have well tested.

PR #2030, opened 2016-02-17
Add full-text search groups extract

sbrunner:
fix https://github.com/camptocamp/ngeo/issues/422


ger-benjamin:
The code is alright but it's hard to say if that works, or just if it does what we expect.
But merge if you have well tested.

PR #2036, opened 2016-02-23
Use hostname instead of facter

### LINKED ISSUE #2035: Use hostname instead of facter
`hostname -f` instead of `facter fqdn`

fredj:
Fixes #2035 

untested


sbrunner:
Looks working but Travis environment variable is broken ...

sbrunner:
Looks good, merge event if Travis failed on deploy ...

PR #2050, opened 2016-03-02
set version of pip not causing error when installing PasteScript

ochriste:
because pip version 8.* is currently not compatible


sbrunner:
Looks good, merge if Travis tests pass :-)

PR #2055, opened 2016-03-02
Add prefix for layers in the admin interface

### LINKED ISSUE #2049: In the admin interface we should display the layer type
When we work with layer v1 and v2 we can have two layers with the same name ...

sbrunner:
fix #2049

### DIFF
--- a/c2cgeoportal/forms.py
+++ b/c2cgeoportal/forms.py
@@ -258,7 +258,9 @@ class SimpleLayerCheckBoxTreeSet(CheckBoxTreeSet):  # pragma: no cover
         result = "<li>"
         if self.auto_check:
             result += '<input type="checkbox"></input>'
-        result += "<label>%(label)s</label>" % {"label": item.name}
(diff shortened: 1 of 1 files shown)


ger-benjamin:
Good improvement, thanks!
Not tested but seems alright. Please merge.

c2cgeoportal/forms.py
```diff
@@ -299,6 +307,7 @@ def render_item(self, item, depth):
             "name": self.name + ("-second" if final_item.id in self._rendered_id else ""),
             "value": final_item.id,
             "add": ' checked="checked"' if self.is_checked(item, final_item) else "",
+            "type": prefixs.get(final_item.item_type, ""),
```
ger-benjamin: Perhaps you can add a comma (",") at the end of line 311 (under the new "type" value). Just to be constant.

sbrunner: Thanks, added :-)

PR #2057, opened 2016-03-02
Print layout name are not correct

yjacolin:
Print layout name are different between print configuration and
vars_+package+.yaml.

(to backport to 1.6)


sbrunner:
> (to backport to 1.6)

Then why on master ?

yjacolin:
because the issue is both in master and 1.6 :) I just thought that you backport some commit to 1.6. Do I need to create another PR for 1.6?

I don't mean to not to merge to master, I meant to backport also this commit to 1.6 :)

sbrunner:
no we do the pull request only on 1.6 and we regulars merge the 1.6 branch into master !!

PR #2061, opened 2016-03-02
Print layout name are not correct

yjacolin:
Print layout name are different between print configuration and
vars_+package+.yaml.


sbrunner:
Thanks, merge :-)

PR #2068, opened 2016-03-04
Host Checker should be ${host} by default

yjacolin:


sbrunner:
Host is localhost then that's wrong :-/

yjacolin:
So that's one is wrong: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/create/vars_%2Bpackage%2B.yaml_tmpl#L8 ?!

sbrunner:
Non because in the application we forward the host header, and when we call the checker from c2ctool we needs to fill it:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scripts/c2ctool.py#L186

Finally the first one is used to build the URL, the second on it to set a header.

yjacolin:
ok, the update process uses this values also and shows something strange:

```
$ .build/venv/bin/c2ctool upgrade yjacolin.mk 1.6.7 --step 3
make: 'project.yaml' is up to date.
================================================================
Checker error:
Open `http://<host>/yjacolin/wsgi/check_collector?` for more informations.
Correct them, then type:
.build/venv/bin/c2ctool upgrade yjacolin.mk 1.6.7 --step 3
```

sbrunner:
The URL used from the server contains localhost,
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scripts/c2ctool.py#L183
and the URL that you should use in your browser contains the host
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scripts/c2ctool.py#L192

This is required when we are behind a reverse-proxy.

PR #2069, opened 2016-03-04
TODAY() is not a native Java method

yjacolin:
Use of new DATE() method instead of TODAY() which is not a native Java
method and throw errors in some configuration.


sbrunner:
Thanks :-)

tsauerwein:
`TODAY()` is part of `jasperreports-functions`. If this fails, there is likely a problem with including the jars of `jasperreports-functions` and joda time. But I am not opposed to your change.

sbrunner:
Good, complement @tsauerwein, I prefer with this change to have less possible issue :-)

Issue #1937, opened 2016-01-19
Infrastructure


ybolognini:
General infrastructure tasks ONLY.
Do not link PRs that should be linked to another issue!

PR #2074, opened 2016-03-10
Fix an incompatibility issue between WebError and pyramid_debugtoolbar

sbrunner:
See also: https://github.com/Pylons/weberror/issues/13

@kalbermattenm this should fix your issue :-)

[#1937]


kalbermattenm:
Yes, it does !

Thanks @sbrunner !

pvalsecc:
Ship it!

PR #2083, opened 2016-03-16
Using git version for transifex-client

ochriste:
To solve bug with windows ssl cert validation.


sbrunner:
Looks good thanks :-)

PR #2091, opened 2016-03-22
Fix locale_negotiator for fulltextsearch

pvalsecc:
This function was relying on Piramid to use default_locale_name if no locale
is found. But fulltextsearch is using this function directly and was returing
status=500 if the browser's languages were not matching the list of available
languages.


sbrunner:
I agree but Travis don't ...

sbrunner:
maven.ibiblio.org looks down ...

sbrunner:
Should be rebased :-)

sbrunner:
Thanks, merge :-)

PR #2096, opened 2016-03-22
Best place for b flag in git submodule add command

yjacolin:
git submodule add -b <version> <repository> <path> is working.

### DIFF
--- a/doc/integrator/create_application.rst
+++ b/doc/integrator/create_application.rst
@@ -206,7 +206,8 @@ Add the CGXP submodule:
 
 .. prompt:: bash
 
-    git submodule add git@github.com:camptocamp/cgxp.git <project>/<package>/static/lib/cgxp -b <version>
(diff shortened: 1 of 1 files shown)


sbrunner:
It don't do the same, what you purpose add the branch in the .gitmodules, and the c2ctool suppose that's not the case ...

yjacolin:
@sbrunner what the original commande line is suppose to do?

I thought that my CGXP submodules will be on 1.6 branch which is not.

sbrunner:
effectively there is something wrong, but i don't like this reference of the original branch in the `.gitmodules` it can be confusing...
then I prefer:
```
git submodule add git@github.com:camptocamp/cgxp.git <project>/<package>/static/lib/cgxp
git submodules foreach git checkout <version>
```

sbrunner:
Thanks that's looks good, merge when Travis tests pass :-)

PR #2095, opened 2016-03-22
Update the joda-time jar URL

sbrunner:

PR #2097, opened 2016-03-22
Use of maven central repository

yjacolin:
It is better to use the official maven repository instead of a custom maven
repository. ibiblio is currently down.


sbrunner:
Duplicated with #2095

PR #2119, opened 2016-03-31
Add dependency to zip

sbrunner:


fredj:
please merge when travis is green

kalbermattenm:
This means that on Windows, the `zip` command has to exist, right ?

sbrunner:
Yes and also unzip :-)
If i's difficult we can find a way to don't need them on windows...

kalbermattenm:
Nope, should be possible (checked it, it is possible to install it through Cygwin)

I would just like to update the documentation (http://docs.camptocamp.net/c2cgeoportal/master/integrator/requirements.html#additional-notes-for-windows-users)...

Thanks for your answer.

PR #2126, opened 2016-04-01
Add layer with restricted access

yjacolin:
* Data source is the same as previous one
* Use of LEADER as a MapServer example
* MapServer 7.0 only (COMPOSITION object is used)

(Don't know if this is really useful for the project but brought protected layer via role_id)


c2cgeoportal/scaffolds/update/mapserver/example.map.mako_tmpl
```diff
@@ -89,3 +89,201 @@ LAYER
...
+LAYER
+    NAME "density"
+    TYPE POLYGON
+    STATUS ON
+    TEMPLATE fooOnlyForWMSGetFeatureInfo # For GetFeatureInfo
+    EXTENT  -31 27 45 71
+    CONNECTIONTYPE postgis
+    PROCESSING "CLOSE_CONNECTION=DEFER" # For performance
+    CONNECTION "${mapserver_connection}"
+    DATA "geom FROM (SELECT geo.*, (pop2005/st_area(geom)/1000000) AS density FROM data.europe_borders AS geo WHERE %role_id% IN (${mapfile_data_noarea_subselect} 'density')) as foo USING UNIQUE gid USING srid=4326"
```
yjacolin: this is wrong, st_area() use srid unit so it won't be in meter.

sbrunner:
Looks good for me

PR #2129, opened 2016-04-04
Forward cookie header in checker and check_collector

arnaud-morvan:
When we apply ACL on views, we need cookie to be forwarded by the checker.
I need this for lyon airport where only login and logout views are public.

### DIFF
--- a/c2cgeoportal/views/checker.py
+++ b/c2cgeoportal/views/checker.py
@@ -57,6 +57,12 @@ def build_url(name, url, request, headers=None):
     else:
         url_ = url
 
+    settings = request.registry.settings.get("checker", {})
+    for header in settings.get("forward_headers", []):
+        value = request.headers.get(header)
+        if value is not None:
+            headers[header] = value
+
     log.info("%s, URL: %s => %s" % (name, url, url_))
     return url_, headers
 
--- a/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -254,6 +254,7 @@ vars:
 
     # Checker configuration
     checker:
+        forward_headers: []
         fulltextsearch: text to search
         lang_files: [cgxp, cgxp-api]
         themes:
--- a/c2cgeoportal/tests/test_checker.py
+++ b/c2cgeoportal/tests/test_checker.py
@@ -90,3 +90,29 @@ class TestExportCSVView(TestCase):
                 }
             )
         )
+
+    def test_build_url_forward_headers(self):
+        request = DummyRequest()
+        request.environ = {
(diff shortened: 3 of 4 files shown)


sbrunner:
??? The check collector is used to be call by a tool like nagios, He shouldn't use a cookie ...

sbrunner:
@ lyon airport you configure the plugin in nagios ?

arnaud-morvan:
No, but they ask me for a first intervention manual, and I was thinking it was a good point to give us a link to ckeck collector. Is it reserved to nagios ?

sbrunner:
Non but we want to have exactly the same comportment, we don't want to have an error in Nagios but not in the browser (or the inverse).

They will not use a tool like Nagios?

ochriste:
@arnaud-morvan cant you access the checker using a localhost url using curl from the server directly, if as I understand it is only to debug/configure the project

arnaud-morvan:
They ask for a list of point to verify that all works correctly.
It seems obvious to give them an url to the check_collector.

If the checker it-self do not provide credential, it will get an 401 forbidden response with the login form, because I've put default permission on all views except login page.

If nagios provide cookies to the checker, they will be forwarded, if do not provide cookies, nothing will be forwarded. From my point of view, i don't see any difference between Nagios and my scenario.

arnaud-morvan:
After discussion with lyon airport, there are currently introducing Nagios in their infrastructure. He's just asked me if we have an interface for it in GMF. But the problem seems the same.

Note that we should use Kerberos Single Sign On on this project. This use also "Authorisation" header.

Maybe we could had a forward_headers in configuration, like for mapfish print ?
I think that these headers (Cookie, Authorisation) could be forwarded by default.

arnaud-morvan:
Travis is happy now, can someone make a review.

c2cgeoportal/views/checker.py
```diff
@@ -57,6 +57,13 @@ def build_url(name, url, request, headers=None):
     else:
         url_ = url
 
+    settings = request.registry.settings.get("checker")
+    if settings:
+        for header in settings.get("forward_headers") or []:
```
sbrunner: Better:
```js
settings = request.registry.settings.get("checker", {})
for header in settings.get("forward_headers", []):
    ...
```

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
```diff
@@ -75,6 +75,7 @@ vars:
 
     # Checker configuration
     checker:
+        forward_headers: []
```
sbrunner: Should be in the `CONST_vars.yaml` :-)

sbrunner:
Just two small comments also that's looks good :-)

sbrunner:
Thanks, merge :-)

PR #2138, opened 2016-04-07
Add more stats (rendering time).

pvalsecc:
Fix a small bug where invalid statsd packet where sent when the key
was containing weird chars.

Fix build when there is no c2cgeoportal installed on the machine


sbrunner:
Looks good :-)

PR #2156, opened 2016-04-14
Ability to log the last update to a feature

fredj:
work in progress

TODO:
 - [x] filter out the two columns; they should not be editable on the client side
 - [x] handle create
 - [x] metadata ui instead of config file

### DIFF
--- a/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
@@ -59,6 +59,8 @@ vars:
 
     layers:
         geometry_validation: True
+        last_update_date_column: False
+        last_update_user_column: False
 
     # For base layers
     tiles_url:
--- a/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
@@ -59,8 +59,6 @@ vars:
 
     layers:
         geometry_validation: True
-        last_update_date_column: False
-        last_update_user_column: False
 
     # For base layers
     tiles_url:
--- a/c2cgeoportal/views/layers.py
+++ b/c2cgeoportal/views/layers.py
@@ -261,7 +261,7 @@ class Layers(object):
         try:
             features = protocol.create(self.request)
             for feature in features.features:
-                self._log_last_update(feature)
+                self._log_last_update(layer, feature)
             return features
         except TopologicalError, e:
             self.request.response.status_int = 400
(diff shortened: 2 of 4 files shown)


c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
```diff
@@ -59,6 +59,8 @@ vars:
 
     layers:
         geometry_validation: True
+        last_update_date_column: False
+        last_update_user_column: False
```
sbrunner: Shouldn't be in the UI metadata?

fredj: yes, it could be moved to the UI metadata but that's not a UI feature, only server side

c2cgeoportal/views/layers.py
```diff
@@ -362,7 +376,16 @@ def metadata(self):
         if not layer.public and self.request.user is None:
             raise HTTPForbidden()
 
-        return self._metadata(str(layer.geo_table), layer.exclude_properties)
+        # exclude the columns used to record the last features update
+        exclude = layer.exclude_properties if layer.exclude_properties else ""
```
sbrunner: => `... if layer.exclude_properties is None ...`

c2cgeoportal/views/layers.py
```diff
@@ -362,7 +376,16 @@ def metadata(self):
...
+        last_update_date = self.settings.get("last_update_date_column", False)
+        if last_update_date:
+            exclude += "," + last_update_date
```
sbrunner: `exclude` can be ',date'?
the start ',' is wrong...
I think that `exclude` should be a list and at the end we should do `",".join(exclude)` ...

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
sbrunner: Yes finally I think that `UI metadata` should be just `Metadata`...
And I think that should be configured by layer...

c2cgeoportal/views/layers.py
fredj: layer.exclude_properties is a string: `a,b,c`
i'm adding an element to this list with `"," + last_update_date` to have `a,b,c,d`

sbrunner: Yes but with the else it can also be en empty string => `,d` ...

fredj: ouch, you're right ... :-)

fredj:
ready to be reviewed

sbrunner:
Can you write a test (probably in this file https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/tests/functional/test_layers.py) ?

sbrunner:
Also that's looks good :-)

fredj:
The travis error will be fixed later, see https://github.com/camptocamp/c2cgeoportal/pull/2269#issuecomment-221835500

sbrunner:
It needs some tests: https://coveralls.io/builds/6383164/source?filename=c2cgeoportal%2Fviews%2Flayers.py#L332
Is it possible to write some tests?

PR #2160, opened 2016-04-14
Use new version of c2c.template

sbrunner:
#1937


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -77,13 +77,13 @@ def __call__(self, filename, options):
         with open(filename) as config_file:
             config = yaml.load(config_file)
             # for application config (.build/config.yaml)
-            if "raster" in config:
+            if "vars" in config:
```
pvalsecc: Not 100% equivalent. Are you sure you don't want that?
```
if "vars" in config and "raster" in config["vars"]:
```

sbrunner: yes it's not 100% equivalent but the latest test is a little bit a hack, I prefer the new one :-)

PR #2174, opened 2016-04-26
Make apache2ctl graceful configurable

### LINKED ISSUE #2170: Hard-coded path for apache 'graceful' action in c2ctool
Just here: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scripts/c2ctool.py#L349

And that doesn't work on RedHat. Can  we use the `APACHE_GRACEFUL` variable from the makefile ?

sbrunner:
fix #2170


fredj:
LGTM

PR #2191, opened 2016-05-03
Note for Windows because export does not exist

kalbermattenm:
Title says it all

Please review (and eventually merge...)


sbrunner:
Thanks :-)

PR #2195, opened 2016-05-03
Fix the vary assignment

### LINKED ISSUE #2193: Vary the cache by Origin for CORS responses
Same as #2112 but for the print

sbrunner:
fix #2193

### DIFF
--- a/c2cgeoportal/lib/caching.py
+++ b/c2cgeoportal/lib/caching.py
@@ -139,8 +139,10 @@ def set_cors_headers(service_headers_settings, request, service_name,
     response.headers.update({
         "Access-Control-Allow-Origin": origin,
         "Access-Control-Allow-Methods": CORS_METHODS,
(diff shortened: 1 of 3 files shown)


c2cgeoportal/lib/caching.py
```diff
@@ -199,7 +200,9 @@ def set_common_headers(
         )
 
     if vary:
-        response.headers["Vary"] = "Accept-Language"
+        response.headers["Vary"].append("Accept-Language")
+
+    response.headers["Vary"] = ", ".join(response.headers["Vary"])
```
fredj: not inside the block (when vary is True) ?

sbrunner: Yes because we also want it for the Origin and Accept-Encoding :-)

sbrunner: in fact the vari variable don't have the right name...

PR #2223, opened 2016-05-12
Allow to have a master/replica DB configuration

pvalsecc:
Basically, this guy switches the DB connection to a replica (readonly) DB if the HTTP method is GET or OPTIONS.

### DIFF
--- a/c2cgeoportal/__init__.py
+++ b/c2cgeoportal/__init__.py
@@ -564,6 +564,15 @@ def includeme(config):
     config.include(pyramid_tm.includeme)
     config.include("pyramid_closure")
 
+    if "sqlalchemy_replica.url" in settings and \
+            settings["sqlalchemy.url"] != settings["sqlalchemy_replica.url"]:  # pragma: nocover
+        # Setup a replica DB connection and add a tween to switch between it and the default one.
+        log.info("Using a replica DB for reading")
+        engine_replica = sqlalchemy.engine_from_config(config.get_settings(), "sqlalchemy_replica.")
+        sqlahelper.add_engine(engine_replica, name="replica")
+        config.add_tween("c2cgeoportal.models.db_chooser_tween_factory",
+                         over="pyramid_tm.tm_tween_factory")
+
     # initialize the dbreflection module
     dbreflection.init(engine)
 
--- a/c2cgeoportal/models.py
+++ b/c2cgeoportal/models.py
@@ -37,6 +37,7 @@ except ImportError:  # pragma: nocover
 
 import sqlahelper
 from papyrus.geo_interface import GeoInterface
+import re
 from sqlalchemy import ForeignKey, Table, event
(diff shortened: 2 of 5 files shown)


tsauerwein:
The session is bound to a request, so binding the connection would work (like you say, assuming that no writes are done in GET requests). You are proposing to use a DNS to support multiple read-only instances. We could also allow to specify multiple read-only connections and then use a simple random load-balancing strategy to avoid the name-resolution step.

Maybe you want to add some documentation on how to configure the actual replication in Postgres?

And on a side node: It would be great if we could avoid a historically racially charged term and maybe use "primary - replica/read-only" instead (for example like Drupal or CouchDB).

pvalsecc:
@tsauerwein, slavery is not necessarily linked to races. Romans were enslaving their own, sometimes. But OK, I'm removing references to our dark past (and present) and I'm using Postgres' other terms: master/replica. And added a link to their doc for how to set it up.

Regarding the DNS or multiple addresses, after discussion with our IT experts, it's not enough. If you want failover, you need to setup two HAProxy:

* one for the master with the functionality to elect a new master and point to it in case of failure
* one the the replicas
 
I've added that to the doc, but without much details. It's out of my league/responsibilities ;-)

pvalsecc:
@fredj, I've heard you've worked on that problem in the past. Any opinion?

tsauerwein:
> @tsauerwein, slavery is not necessarily linked to races. Romans were enslaving their own, sometimes. But OK, I'm removing references to our dark past (and present) and I'm using Postgres' other terms: master/replica. And added a link to their doc for how to set it up.

Thanks! :+1:

fredj:
We created 2 distinct engines (one for reads and one for writes) and we used one or the other depending on the operation.
It was not automatic but the principle is the same

sbrunner:
I don't found the therm replica relay useful, every body know whats a master-slave architecture is and we don't do philosophy...

sbrunner:
https://www.google.ch/trends/explore#q=master%20replica%2C%20master%20slave&cmpt=q&tz=Etc%2FGMT-2

tsauerwein:
> ...like you say, assuming that no writes are done in GET requests.

Here is an example of a GET request that also writes to the database:
https://github.com/camptocamp/c2cgeoportal/blob/94259524d78a032decd5fde3fb69d2e05afcdfc9/c2cgeoportal/views/shortener.py#L55-L64

pvalsecc:
@tsauerwein, you should look at the last version of my patch. I've added the possibility to force the usage of the master or the replica for some URLs. And the shortener is part of it.

tsauerwein:
> I've added the possibility to force the usage of the master or the replica for some URLs. And the shortener is part of it.

Perfect, I missed that.

pvalsecc:
So... what now? Merge or not?
To: @tsauerwein @sbrunner

tsauerwein:
LGTM

PR #2228, opened 2016-05-13
Use node to read the JS to be stronger

sbrunner:
Should fix https://github.com/camptocamp/cartoriviera_geomapfish/issues/9


fredj:
LGTM

PR #2264, opened 2016-05-23
Fix print yaml error

sbrunner:
like
found unknown escape character '.'


pvalsecc:
Maybe even safer if you don't put quotes at all. But works fine for this context. Ship it!

PR #2266, opened 2016-05-24
Explain PostGIS connexion using pg_service.conf

yjacolin:
.pg_service.conf file allow the use of a service name instead of parameters for
to set up a PostGIS connection. So for the same QGIs project, you can easily
change the connection parameters switching the .pg_service.conf file.

In a client <-> server process, you can use the same QGIs project file with
different PostGIS connection.

### DIFF
--- a/doc/integrator/backend_qgis.rst
+++ b/doc/integrator/backend_qgis.rst
@@ -51,7 +51,6 @@ In the ``vars_<project>.yaml`` file, define:
         ...
             mapserverproxy:
                 mapserv_url: http://your.qgis.server/ows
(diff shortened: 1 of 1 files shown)


yjacolin:
cc @eleu @marionb

sbrunner:
What's the status ?
By the way it should be on 2.1...

yjacolin:
I will try to rebase it later. Please don't close it :)

sbrunner:
up ...

yjacolin:
See new PR https://github.com/camptocamp/c2cgeoportal/pull/2903

Issue #2270, opened 2016-05-27
Deploy: error if db doesnt exist


ochriste:
When deploying the db for the first time, we get an error because the https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/create/deploy/hooks/pre-restore-database.mako script tries to do some psql command on a none-existant database.

we should first try to detect if the db existes, and if yes, execute the psql commands.

PR #2271, opened 2016-05-27
Check if database exists before executing psql commands.

ochriste:
fix https://github.com/camptocamp/c2cgeoportal/issues/2270

### DIFF
--- a/c2cgeoportal/scaffolds/create/deploy/hooks/pre-restore-database.mako
+++ b/c2cgeoportal/scaffolds/create/deploy/hooks/pre-restore-database.mako
@@ -11,23 +11,26 @@ DATABASES=$@
 # databases / tables restore.
 sudo apache2ctl stop
 
+if psql --list --tuples-only | cut --delimiter='|' --fields=1 | grep --quiet --fixed-strings $DATABASES; then
 
-# The following line works only with Postgres 9.3 and upper.
-# otherwise the schema should be create manually
-psql -c 'CREATE SCHEMA IF NOT EXISTS ${schema}_static;' ${db}
+    # The following line works only with Postgres 9.3 and upper.
+    # otherwise the schema should be create manually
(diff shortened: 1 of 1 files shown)


c2cgeoportal/scaffolds/create/deploy/hooks/pre-restore-database.mako
```diff
@@ -11,23 +11,26 @@ DATABASES=$@
 # databases / tables restore.
 sudo apache2ctl stop
 
+if psql -lqt | cut -d \| -f 1 | grep -qw $DATABASES; then
```
sbrunner: no `sudo -u postgresql`?
long parameters =>
```
psql --list --quiet --tuples-only | cut --delimiter='|' --fields=1 | grep --quiet --word-regexp $DATABASES
```

why `psql --quiet`?
why `grep --quiet`?
`--word-regexp` should be `--fixed-strings`...

ochriste: we dont do any ´sudo -u postgresql´ in the other command either, why would we ?

if you know how to make the command line shorter, please tell me

I can remove the q option if you dont like it

I can change the --word-regexp

sbrunner: for  ´sudo -u postgresql´ your right :-)

> I can remove the q option if you don't like it

It's not that I don't like it but I prefer to don't have unneeded argument :-)

+1

ochriste: we need the --quiet for grep, otherwise the script will output the result of the grep (string), while we only need to know if it founds it or not (exit status)

sbrunner: OK, thanks for explain :-)

sbrunner:
Looks good for me :-)

PR #2281, opened 2016-06-02
Fix Make dependency on the generated files from mako

sbrunner:


pvalsecc:
IMHO, it would be better to have the print depend on only the files in the print directory and the same for mapserver.

That way, when you do `make print`, it cares only on those files and not the ones from the entire project.

The would make your other PR useless...

sbrunner:
I agree with your comment, but I will do it in an other pull request :-), then merging

PR #2282, opened 2016-06-03
Fixed tsearch query to use bindings

pvalsecc:
Fulltextsearch was not using variable binding for searched terms. This was
leading to a lot of different statistics for those queries, leading to
memory leak if the MemoryBackend was enabled and a lot of data in statsd.


sbrunner:
For me it's cryptic, I hope that's ok :-)

sbrunner:
Great :-), Merge when Travis is happy :-)

PR #2237, opened 2016-05-18
Queryable

sbrunner:
Server part of https://github.com/camptocamp/ngeo/issues/962

PR #2284, opened 2016-06-03
Fix WMS Layer documentation

sbrunner:
see: https://github.com/camptocamp/c2cgeoportal/pull/2237/files#r65681454


tsauerwein:
Thanks! :+1:

PR #2287, opened 2016-06-06
Get the query string in the mobile URL redirect

### LINKED ISSUE #2234: Permalink is not compatible between desktop and mobile
This permalink, generated on desktop GeoMapfish 2.0, cannot be opened on mobile GeoMapfish 2.0 : 

https://preprod.cartoriviera.ch/v2/s/2SX9 

[from @sbrunner] To resume, now:
* [x] the zoom is ok
* [x] the position is ok
* [x] the theme is ok
* [x] the language is ok
* [x] the background is ok
* [ ] tree_group_layers_* does not work on 2.x ...
* [x] the redirect to mobile don't works

sbrunner:
Partially fix #2234


fredj:
LGTM

Issue #2278, opened 2016-06-02
In the theme set the time data at the right place


sbrunner:
* on the leaf for mixed group
* on the leaf on non mixed group with only one time layer
* on the first gtoup for non mixed group with more than one time

Also check if the merge of time informations is correct.

PR #2288, opened 2016-06-06
Put the time information on each layer in mixed group

sbrunner:
Add time informations test

fix: #2278


pvalsecc:
LGTM. I guess you are missing a few nocover...

sbrunner:
It's not me it's fred :-), see: https://github.com/camptocamp/c2cgeoportal/pull/2156#issuecomment-222429812

PR #2295, opened 2016-06-08
Be able to have multiple id values for PDF reports

sbrunner:
Fix: https://github.com/camptocamp/baselstadt_mapbs/issues/104

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -278,6 +278,10 @@ add it in the `.gitignore` (at step 3).
 
 8. It is not allowed anymore to have layers directly in a theme.
 
+13. In the PDF report we now make a list id ids, then we should change:
+    - The `paramID` is replaced by a `ids` and it should be a `!stringArray`.
+    - In the `spec` template the `id` is renamed to `ids` and will be a list.
+
 
 Version 1.6.8
 =============
--- a/c2cgeoportal/views/pdfreport.py
+++ b/c2cgeoportal/views/pdfreport.py
@@ -104,12 +104,14 @@ class PdfReport(OGCProxy):  # pragma: no cover
 
     @view_config(route_name="pdfreport", renderer="json")
     def get_report(self):
-        id_ = self.request.matchdict["id"]
+        ids = self.request.matchdict["ids"].split(",")
         self.layername = self.request.matchdict["layername"]
         layer_config = self.config["layers"].get(self.layername)
 
         if layer_config is None:
             raise HTTPBadRequest("Layer not found")
+
+        features_ids = [self.layername + "." + id_ for id_ in ids]
 
         if layer_config["check_credentials"]:
             # check user credentials
@@ -142,7 +144,7 @@ class PdfReport(OGCProxy):  # pragma: no cover
                 "outputformat": "gml3",
                 "request": "GetFeature",
                 "typeName": self.layername,
-                "featureid": self.layername + "." + id_,
+                "featureid": ",".join(features_ids),
                 "srsName": "epsg:" + str(srs)
             }.items()])
         )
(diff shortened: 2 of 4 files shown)


c2cgeoportal/views/pdfreport.py
```diff
@@ -155,7 +157,7 @@ def get_report(self):
                 "outputformat": "gml3",
                 "request": "GetFeature",
                 "typeName": self.layername,
-                "featureid": self.layername + "." + id,
+                "featureid": ",".join(features_ids),
```
tsauerwein: How will the property `featureid` be used in the jrxml template?

sbrunner: Effectively I miss it, as a table?

sbrunner: or a datasource

tsauerwein: Usually for these kind of reports a datasource with a query in the template is used:
https://github.com/camptocamp/baselstadt_mapbs/blob/master/print/print-apps/mapbs/Hochstammobstbaeume_Baeume.jrxml#L10-L19

We could do the same for multiple feature ids but we have to make sure that there is no SQL injection possible.

sbrunner: Updated :-)

sbrunner: Isn't the role of who write the report that there is no possible injection?
Because we can use the attribute just to display it...

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -106,6 +106,10 @@ replace your sencha touch interface by an ngeo interface.
       </script>
       % endif
 
+13. In the PDF report we now make a list id ids, then we should change:
+    - The `paramID` is replaced by a `datasource` with an `id` and it should be a `!datasource` with a `id: !string`.
```
tsauerwein: The way BS is using that feature is that they do not define a datasource in `config.yaml` but directly in the template.

sbrunner: Have you a link on the BS code?

tsauerwein: https://github.com/camptocamp/baselstadt_mapbs/blob/master/print/print-apps/mapbs/Hochstammobstbaeume_Baeume.jrxml

c2cgeoportal/views/pdfreport.py
sbrunner: Cool :-)

sbrunner:
@eleu what's the status of that?

eleu:
is the feature ready? dev is finished and ready to merge? -> merge and I can tell the customer to test.

sbrunner:
The goal was to tell the customer to test before merging because we have no way to do a good test...

eleu:
as far as I know, the customer cannot test as long as the development is not merged. If this is not the case: Can you give detailed instructions what to do in order to test it?

sbrunner:
Yes: http://docs.camptocamp.net/c2cgeoportal/1.6/developer/server_side.html

ger-benjamin:
@eleu @sbrunner any news about this PR ?
Can we do something to merge and, or close it ?

svamet:
Hi, we just wanted to test this feature but we are still on 2.1.7. 
Do i see it correct, that it will only be available in version 2.2?

sbrunner:
No, it was never merged... is it required for 2.2?

svamet:
Yes, this feature is required for 2.2 
A long time ago you asked us to test the feature in the branch "report-many" (https://github.com/camptocamp/baselstadt_mapbs/issues/104)
But because we are allready on 2.1.7 and we assume that the branch "report-many" is build on 2.0 we inserted the changes from  https://github.com/camptocamp/c2cgeoportal/pull/2295/commits/dfdc85232a212ae4a47d46814aed45fe8b9e4374 manualy. 
But it does not work. We get the following error:
```
ids = self.request.matchdict["ids"].split(",")
KeyError: 'ids'
```
Would it maybe possible to build this branch on 2.1.8 so we could test it properly?

Issue #1825, opened 2015-11-27
Update the travis infrastructure and use docker


sbrunner:
Doc: https://docs.travis-ci.com/user/ci-environment/

Probably by using the config:

``` yaml
sudo: required
services:
- docker
```

And launch the tests in docker

Inject config using sidekick containers

https://docs.travis-ci.com/user/ci-environment/
https://docs.travis-ci.com/user/docker/
- [x] c2cgeoportal and ngeo build (`camptocamp/geomapfish_build_dev`) (we should install node and python, and the npm and python packages ad the right version)
- [ ] Investigate to use docker cp in place of the 3 run script, it needs to separate the generated files from the sources.
- [ ] separate ngeo build container?
- [ ] build (`camptocamp/geomapfish_build`) (`camptocamp/geomapfish_build_dev` + c2cgeoportal)
- [ ] wsgi (project exists, we should create an intermediate container) (`camptocamp/c2cgeoportal_wsgi`)
- [ ] print (exists, we should create an intermediate container or put jars in the standard print, https://hub.docker.com/r/camptocamp/mapfish_print/) (`camptocamp/mapfishprint` if we can add everything what we needs) 
- [ ] mapserver (exists, https://hub.docker.com/r/camptocamp/mapserver/) (`camptocamp/mapserver`)
- [ ] mapcache (`camptocamp/mapcache`)
- [ ] memcached (https://hub.docker.com/_/memcached/) (`_/memcached`)
- [ ] tilecloudchain create (install python and tilecloud-chain) (`camptocamp/tilecloud_chain`)
- [ ] qgis-server (https://hub.docker.com/r/pvalsecc/qgis-server/, https://hub.docker.com/r/kartoza/qgis-server/) (`qgis/qgis-server`)
- [ ] Probably a separated docker for the admin interface
- [ ] nginx (`_/nginx`) or HAProxy(for rancher or SSL endpoint, `_/haproxy`) (on front, Should be managed by the customer)
- [ ] SSH (camptocamp/sshd-shell) ?
- [ ] c2ctool and db model should be in separate eggs.

Separate the project in component, (one folder per Docker container)

Make file organization in the project
- Get the `docker-run` (and `docker-compose-run`) from c2cgeoportal
- Add a Makefile with the specific thing from docker
- Will be run with `./docker-run make ...`

Others things:
- https://github.com/mapserver/mapcache/pull/116
[...]

PR #2299, opened 2016-06-08
Docker integration

sbrunner:
First step for #1825 
- [x] Use a container for the database
- [x] Use a separate container for Mapserver
- [x] Use docker ignore in place of multiple coppy
- [x] Fix the pip version
- [x] Suppress traces on failures
- [x] Fix coverage report send
- [x] Fix the coverage


docker-run.bat
```diff
@@ -0,0 +1,22 @@
+SET SRC_DIR=/c/Users/ochriste/test/c2cgeoportal
```
gberaudo: ochriste?

docker/build/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM debian:stable
```
gberaudo: It may be preferable to depend on a fixed stable release. Would `debian:jessie` work?

sbrunner: Done :-)

PR #2302, opened 2016-06-09
Fix error 500 in fulltextsearch in case of special chars

pvalsecc:
For example, searching "wald (genf)" was leading to an SQL error.


sbrunner:
Looks good, for me :-)

PR #2326, opened 2016-06-17
Fix path to jquery-ui.min.js

gberaudo:


sbrunner:
Thanks, merge when Travis is happy :-)

PR #2328, opened 2016-06-20
The WMTS layer is in the attribute name

sbrunner:


fredj:
LGTM

sbrunner:
Replaced by #2333

PR #2332, opened 2016-06-21
pot creation requires mapserver

sbrunner:


gberaudo:
@sbrunner do you know why it builds OK now?

sbrunner:
I also  surprised, but I cut the apache graceful in two parts, and it seam that a task who create the error don't run again :-)

gberaudo:
OK, then let's merge it.

PR #2335, opened 2016-06-21
Simplify CONST_Makefile

fredj:
No need to remove from `.build/externs`; `.build` is already removed (`rm -rf .build` 2 lines above).


sbrunner:
Thanks, merge :-)

PR #2348, opened 2016-06-28
Match theme with a simple pattern

gberaudo:
The previous pattern was matching all the end of the URL, which was too much. If an Angular partial was erronously using a prefix with 'theme' it bugged the browser in an infinite loop.


fredj:
+1

sbrunner:
Merge :-)

PR #2350, opened 2016-06-28
Docker: add alembic to the WSGI image

pvalsecc:
Now, the WSGI can be used as an Alembic container to upgrade the DB schema.

Fix the Makefile dependencies for Alembic generated SQL files.

### DIFF
--- a/doc/integrator/docker.rst
+++ b/doc/integrator/docker.rst
@@ -131,3 +131,25 @@ something like that:
 
     environment:
       DB_CONNECTION: user=www-data password=toto dbname=geoacordaDev host=db
+
+
+Keep your DB schema up to date
+------------------------------
+
+The WSGI image contains Alembic. You can use it as a start once container and
+add something like that in your composition:
+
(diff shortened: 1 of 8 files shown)


doc/integrator/docker.rst
```diff
@@ -131,3 +131,25 @@ something like that:
...
+------------------------------
+
+The WSGI image contains Alembic. You can use it as a start once container and
+add something like that in your composition:
+
+.. code:: yaml
+
+    alembic:
+      labels:
+        io.rancher.container.start_once: 'true'
```
sbrunner: true as a string?

pvalsecc: Yes, it's what rancher is generating when you use their GUI to edit that. So they must expect a string, I guess.

sbrunner: :-/, thanks :-)

sbrunner:
Looks good, merge when Travis tests pass :-)

PR #2353, opened 2016-06-29
Fix the master

sbrunner:


fredj:
The alambic file doesn't seems to do anything (no instruction in upgrade or downgrade)

sbrunner:
Effectively, just a merge of two revisions :-)
https://github.com/camptocamp/c2cgeoportal/pull/2353/files#diff-14324ee2f9e907ab056489a9fd2b05d4R41

fredj:
LGTM

PR #2381, opened 2016-07-01
Checks cleanup

sbrunner:


.gitattributes
```diff
@@ -5,6 +5,7 @@ CONST_Makefile_tmpl text whitespace=indent-with-non-tab,tabwidth=2
 *.mk text whitespace=indent-with-non-tab,tabwidth=2
 doc/Makefile -whitespace
 *.rst tex conflict-marker-size=100
+*.rst.mako tex conflict-marker-size=100
```
fredj: `text` instead of `tex` ? (the line above is also wrong)

fredj:
LGTM

.gitattributes
sbrunner: +1, thanks :-)

PR #2387, opened 2016-07-06
Docker: better logging configurability

pvalsecc:
Add a way to configure the logging from the composition.

Fixed the production.ini file loading. I have a feeling it
was silently not working before.

### DIFF
--- a/c2cgeoportal/scaffolds/create/apache/application.wsgi.mako
+++ b/c2cgeoportal/scaffolds/create/apache/application.wsgi.mako
@@ -1,15 +1,28 @@
 import site
 import sys
 import re
+import os
+from logging.config import fileConfig
 
+% if docker == 'TRUE':
+root = "/app"
+% else:
 site.addsitedir("${python_path}")
+root = "${directory}"
+% endif
+
 
 # Remove site packages
 regex = re.compile("^/usr/lib/python.\../dist-packages$")
 sys.path = [p for p in sys.path if regex.match(p) is None]
 
-from pyramid.paster import get_app, setup_logging
+from pyramid.paster import get_app
 
-configfile = "${directory}/${'development' if development == 'TRUE' else 'production'}.ini"
-setup_logging(configfile)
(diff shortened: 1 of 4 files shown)


c2cgeoportal/scaffolds/create/apache/application.wsgi.mako
```diff
@@ -1,15 +1,28 @@
...
+root = "${directory}"
+% endif
+
 
 # Remove site packages
 regex = re.compile("^/usr/lib/python.\../dist-packages$")
 sys.path = [p for p in sys.path if regex.match(p) is None]
 
-from pyramid.paster import get_app, setup_logging
+from pyramid.paster import get_app
```
sbrunner: Any reason to have this import here?

sbrunner:
One question also that's looks good :-)

c2cgeoportal/scaffolds/create/apache/application.wsgi.mako
pvalsecc: Apart from the fact that it was there before... I don't know. I've tried to move it up and I've seen no ill effect.
You want me to move it at the top?

sbrunner: Thanks :-)

pvalsecc: Humm, actually, that works with the import at the top only in docker. For a normal install, the import must be after the addsitedir call and sys.path hacking. Otherwise we have this error:
```
Traceback (most recent call last):
   File "/tmp/travis/testgeomapfish/apache/application.wsgi", line 5, in <module>
     from pyramid.paster import get_app
ImportError: No module named pyramid.paster
```

sbrunner: ok :-/, he probably needs the `addsitedir`

sbrunner:
Merge :-)

PR #2394, opened 2016-07-08
Don't add the MultiDomainStaticURLInfo when it's not needed

sbrunner:
Thant's don't works with the debug toolbar
Update the documentation


fredj:
LGTM

PR #2397, opened 2016-07-12
Get wfs types though mapservproxy

arnaud-morvan:
Using this we could show in cgxp_querier only the layers the user have access to : 
```
featureTypes: ${dumps(list(set(loads(WFSTypes))&set(['locaux','cameras']))) | n},
```

### DIFF
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -691,14 +691,8 @@ class Entry(object):
                         children.append(l)
         return children, errors
 
-    def _get_wfs_url(self):
-        if "mapserv_wfs_url" in self.mapserver_settings and \
-                self.mapserver_settings["mapserv_wfs_url"]:
-            return self.mapserver_settings["mapserv_wfs_url"]
-        return self.mapserver_settings["mapserv_url"]
-
     def _internal_wfs_types(self, role_id):
-        return self._wfs_types(self._get_wfs_url(), role_id)
+        return self._wfs_types(self.request.route_url('mapserverproxy'), role_id)
 
     def _get_external_wfs_url(self):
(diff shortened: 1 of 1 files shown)


sbrunner:
To solve that you should use functionalities as it explained here: http://docs.camptocamp.net/c2cgeoportal/1.6/integrator/functionality.html#using-functionalities-list-to-configure-the-layers-in-the-querybuilder

arnaud-morvan:
Functionnalities do not seems very useful for this. I don't see any reason to not use proxy filtering on wfs_types since role_id is taken in cache key.

sbrunner:
- Using cache on the results
- One less query

arnaud-morvan:
Cache is made on a role basis, so this is not a problem.
I prefer to get one more request than having to keep two different system for layer rights.

arnaud-morvan:
I close the pull request since ``functionnalities`` make the job, but I do not understand why we do not use proxy filtering here.

arnaud-morvan:
@sbrunner 
Since wfs_types are caches on a role basis, we get only one request per role, this really do not make sense to not use proxy filtering. I really do not agree with you on this.

sbrunner:
I don't completely desaggree with that, but
- I don't want to have regression on the branch 1.6 than it should be done on 2.0 or master
- To make the cache working you should add a role_id argument to the methode `_wfs_types_cached`
- You should remove all this block: https://github.com/camptocamp/c2cgeoportal/blob/bb848556e9b7d0ee5d7bec1a4f8f46cefd6fc2fb/c2cgeoportal/views/entry.py#L713-L722
- Have one less method (`_wfs_types` became just a pass throw)
- Do the same for `_external_wfs_types`

PR #2398, opened 2016-07-12
Fix the i18n in the theme 2 fts import script

sbrunner:
- Use <package> instead of demo
- Use client instead of server
- Generate the requires mo files

fix: https://github.com/camptocamp/ngeo/issues/1506


fredj:
LGTM

PR #2400, opened 2016-07-12
Update the documentationabout the layer v1 and v2

### LINKED ISSUE #2378: Adminstrating layers
According to https://camptocamp.github.io/c2cgeoportal/2.0/administrator/administrate.html#layers

> In the version 2 we split the layer table in 3 tables: ‘layer_internal_wms’, ‘layer_external_wms’, ‘layer_wmts’, and we copy the previous layer table in ‘layerv1’.

Is it really the case for desktop layers? Or do we need to maintain the two versions?

sbrunner:
fix #2378

### DIFF
--- a/doc/administrator/administrate.rst.mako
+++ b/doc/administrator/administrate.rst.mako
@@ -11,8 +11,12 @@ required.
 Layers
 ------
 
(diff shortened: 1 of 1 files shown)


fredj:
+1

PR #2401, opened 2016-07-12
Remove the --sort-by-file argument, it will make unnecessary changes

### LINKED ISSUE #2377: Modifications in translations file
Is it normal that after some builds the content of PO file is changed (and need to be commited)?

sbrunner:
fix #2377


fredj:
+1

PR #2404, opened 2016-07-13
Don't use the version argument after the step one.

### LINKED ISSUE #2359: Use version only on the first step in the c2ctool to have less upgrade errors

sbrunner:
fix #2359


fredj:
LGTM

PR #2411, opened 2016-07-13
Search less files in more directories

gberaudo:
This change allows having component styles defined in the components subdirectories.


sbrunner:
Looks good, but then shouldn't the `components` folder be out of the `js` folder?

gberaudo:
Indeed that would make sense. I will update the PR.

sbrunner:
APP_JS_FILES shouldn't also be modified?

sbrunner:
and here https://github.com/camptocamp/ngeo/blob/master/buildtools/mako_build.json#L23-L25 we should add a `src_components`.

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -177,12 +177,12 @@ endif
 APP_OUTPUT_DIR = $(PACKAGE)/static-ngeo/build
 OL_JS_FILES = $(shell $(FIND) node_modules/openlayers/src/ol -type f -name '*.js' 2> /dev/null)
 NGEO_JS_FILES = $(shell $(FIND) node_modules/ngeo/src -type f -name '*.js' 2> /dev/null)
-APP_JS_FILES = $(shell $(FIND) $(PACKAGE)/static-ngeo/js -type f -name '*.js')
+APP_JS_FILES = $(shell $(FIND) $(PACKAGE)/static-ngeo/js $(PACKAGE)/static-ngeo/components/ -type f -name '*.js')
```
sbrunner: without en `/` to be coherent :-)

sbrunner:
`APP_DIRECTIVES_PARTIALS_FILES = $(shell ls -1 $(PACKAGE)/static-ngeo/js/components/*/*.html 2> /dev/null)`
=>
`APP_DIRECTIVES_PARTIALS_FILES = $(shell ls -1 $(PACKAGE)/static-ngeo/components/*/*.html 2> /dev/null)`

gberaudo:
Thanks for the review.

PR #2417, opened 2016-07-14
Add the feature type name in the print results

### LINKED ISSUE #2393: The attribute type should be display on the print results

sbrunner:
fix #2393


fredj:
LGTM

PR #2419, opened 2016-07-20
Add the item type in the UI metadata item combo

sbrunner:


c2cgeoportal/models.py
```diff
@@ -377,8 +377,9 @@ def get_metadatas(self, name):  # pragma: nocover
     def __init__(self, name=u""):
         self.name = name
 
-    def __unicode__(self):
-        return self.name or u""  # pragma: nocover
+    def __unicode__(self):  # pragma: no cover
+        type_ = self.__label__ if hasattr(self, "__label__") else self.item_type
+        return "{}: {}".format(type_, self.name or u"")
```
gberaudo: should return unicode

gberaudo:
one comment, otherwise looks good to me.

PR #2421, opened 2016-07-21
Fix utf8 error on GetCapabilities

sbrunner:
Fix error like
⇝ UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 16: ordinal not in range(128)


gberaudo:
+1 if it works

PR #2424, opened 2016-07-22
Install node version 4.x

sbrunner:


gberaudo:
An alternative could be to switch the language to `node_js`: https://docs.travis-ci.com/user/languages/javascript-with-nodejs.

Do we need to use language `python`?

sbrunner:
Probably yes, it a largely harder change :-(

gberaudo:
OK, feel free to merge then.

PR #2432, opened 2016-07-26
Force to use setuptools

sbrunner:


sbrunner:
Wrong way, replaced by #2433

PR #2447, opened 2016-07-29
Fix the template cache for components based structure

### LINKED ISSUE #2446: The template cache doesn't work for components

sbrunner:
Should fix #2446


gberaudo:
Are you sure this change is necessary?

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -858,10 +858,12 @@ $(APP_OUTPUT_DIR)/templatecache.js: .build/gmftemplatecache.js
 	$(PRERULE_CMD)
 	mkdir -p $(dir $@)
 	PYTHONIOENCODING=UTF-8 $(VENV_BIN)/mako-render \
-		--var "app=$(PACKAGE)" \
-		--var "partials=$(addprefix ngeo:,$(NGEO_DIRECTIVES_PARTIALS_FILES)) \
-		$(addprefix gmf:,$(GMF_DIRECTIVES_PARTIALS_FILES)) \
-		$(addprefix $(PACKAGE):,$(APP_DIRECTIVES_PARTIALS_FILES))" $< > $@
+		--var "partials=ngeo:node_modules/ngeo/src/directives/partials \
+		gmf:ode_modules/ngeo/contribs/gmf/src/directives/partials \
```
sbrunner: missing n

sbrunner:
Yes (I see that the title is not so good), this is also to be able to override the gmf template with a template with the same name:
- to be compatible with https://github.com/camptocamp/ngeo/pull/1644.
- to have one prefix per source.

gberaudo:
OK, makes sense.

gberaudo:
LGTM

PR #2448, opened 2016-07-29
Directly gets the editing columns from the DB

### LINKED ISSUE #2445: Collecting string from editable layers dosn't work
* The URL should be logged
* We should be authenticated ...

=> finally we should find an other way to do that ...

sbrunner:
Should fix #2445


sbrunner:
Replaced by #2455

PR #2449, opened 2016-07-29
Fix the theme to full-text search

### LINKED ISSUE #2444: theme2fts don't works anymore
* Config variable are in vars
 * I should be run from the makefile with a dependency on the mo files
 * I should run on all interface

sbrunner:
- Correctly get config
- Run on all interfaces
- Add in the Make file to have the dependencies

Should fix #2444

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -481,6 +481,7 @@ help:
 	@echo
 	@echo  "Deploy branch target:"
 	@echo
+	@echo  "- theme2fts		Inport the theme in the full-text search"
 	@echo  "- deploy-branch		Deploy the curent branch to demos"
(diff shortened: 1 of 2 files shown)


c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -481,6 +481,7 @@ help:
 	@echo
 	@echo  "Deploy branch target:"
 	@echo
+	@echo  "- theme2fts		Inport the theme in the full-text search"
```
gberaudo: Import

c2cgeoportal/scripts/theme2fts.py
```diff
@@ -47,7 +47,7 @@ def main():
 
     parser.add_argument(
         "--interfaces",
-        nargs='+',
+        nargs='*',
```
gberaudo: Also remove the required=True.

gberaudo:
One remark otherwise LGTM.

PR #2454, opened 2016-08-16
Invalidate cache on role restriction areas change

gberaudo:


gberaudo:
I had errors with this patch. Need to do more checking before a merge.

PR #2455, opened 2016-08-17
Directly get the editing column from the database

sbrunner:
fix: #2445

### DIFF
--- a/c2cgeoportal/lib/lingua_extractor.py
+++ b/c2cgeoportal/lib/lingua_extractor.py
@@ -35,15 +35,20 @@ import yaml
 from json import loads
 from urlparse import urlsplit
 from xml.dom.minidom import parseString
-from sqlalchemy.exc import ProgrammingError
+from sqlalchemy.exc import ProgrammingError, NoSuchTableError
+from sqlalchemy.orm.util import class_mapper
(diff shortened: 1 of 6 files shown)


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -157,33 +161,22 @@ def _import_layer_wms(self, layer, messages):
...
-            ):
-                if not element.getAttribute("type").startswith("gml:"):
+            exclude = [] if layer.exclude_properties is None else layer.exclude_properties.split(",")
+            last_update_date = self._get_ui_metadata(layer, "lastUpdateDateColumn")
+            if last_update_date is not None:
+                exclude.append(last_update_date)
+            last_update_user = self._get_ui_metadata(layer, "lastUpdateUserColumn")
+            if last_update_user is not None:
+                exclude.append(last_update_user)
+            cls = get_class(str(layer.geo_table, ",".join(exclude)))
```
gberaudo: Remove join

gberaudo:
LGTM

gberaudo:
> def get_class(tablename, session=None, exclude_properties=[]):
@sbrunner, is it really OK to use `[]` static array as default parameter?

sbrunner:
Yes if we don't modify it :-)

PR #2456, opened 2016-08-17
Rename mobile_default_theme functionality to default_theme

fredj:
The value will now be used ba the desktop and mobile interfaces


sbrunner:
By the way you can also add it here: https://github.com/camptocamp/c2cgeoportal/blob/9544013afc7f68033099fc92f812137d0fe979df/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl#L195

fredj:
@sbrunner you're right, added

sbrunner:
Looks good, merge, when it's OK for Travis :-)

fredj:
Thanks for the review

PR #2458, opened 2016-08-18
Wms dimension

sbrunner:


c2cgeoportal/models.py
```diff
@@ -914,10 +918,10 @@ def __unicode__(self):  # pragma: no cover
 event.listen(UIMetadata, "after_delete", cache_invalidate_cb, propagate=True)
 
 
-class WMTSDimension(Base):
-    __label__ = _(u"WMTS dimension")
-    __plural__ = _(u"WMTS dimensions")
-    __tablename__ = "wmts_dimension"
+class Dimension(Base):
+    __label__ = _(u"dimension")
+    __plural__ = _(u"Dimensions")
```
gberaudo: Shouldn't it be a small D?

c2cgeoportal/models.py
```diff
@@ -573,6 +573,10 @@ def __init__(self, name=u"", public=True):
         self.public = public
 
 
+class DimensionLayer(Layer):
+    __acl__ = [DENY_ALL]
```
gberaudo: Why do you need this line?

c2cgeoportal/views/entry.py
```diff
@@ -70,7 +70,46 @@
...
+        errors = set()
+
+        dimensions = {}
+        for dimension in layer.dimensions:
+            if dimension.value is not None and not self.URL_PART_RE.match(dimension.value):
+                errors.add("The layer '{}' has an unsupported dimension value '{}' ('{}').".format(
+                    layer.name, dimension.value, dimension.name
+                ))
+            elif dimension.name in dimensions:
+                errors.add("The layer '{}' has an duplicated dimension name '{}'.".format(
```
gberaudo: a

gberaudo: Don't we want to accept compatible value?
- allow dimensions with the same values;
- allow dimensions with the None values.

Then after all the merges we can have a test that no dimension has a None default value.

gberaudo: OK, I understand better now:
- in this first step you do a validation of the data coming from the database;
- in the second step below you implement the merge with the compatibility rules.

Maybe you should make the dimensions property private and add the last validation in a getter method.

c2cgeoportal/models.py
sbrunner: Yes to don't be visible in the admin interface...

c2cgeoportal/models.py
sbrunner: No we always have a capital but the label should have a capital D ...

gberaudo: OK, I think I got it, the __label__ should have the capital.

c2cgeoportal/views/entry.py
sbrunner: Why not ...

gberaudo:
@sbrunner, what is the status of this PR? Can it be merged?

sbrunner:
For me it's ready to be merged :-)

gberaudo:
OK, then when Travis is green please merge.

PR #2461, opened 2016-08-19
Fix resolutions hints

### LINKED ISSUE #2459: WMTS queries requires a MXSCALEDENOM in the queried layer
To reproduce the issue, create a WMTS layer (wmts_layer) with a bound layer to query (layer_q).
 
This layer (layer_q) must have a MAXSCALEDENOM attibut in the LAYER. Else, the layer will be not queriable. In the Themes.json object (received by the client), the wmts layer definition will not have any value in the "query_layers" field.

MINSCALEDENOM seems not require.

Found in GMF 1.5.3 but seems always in the newer version.

sbrunner:
fix #2459

### DIFF
--- a/c2cgeoportal/tests/functional/c2cgeoportal_test.map.mako
+++ b/c2cgeoportal/tests/functional/c2cgeoportal_test.map.mako
@@ -515,4 +515,49 @@ MAP
             END
         END
     END
+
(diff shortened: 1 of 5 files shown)


gberaudo:
LGTM, please merge.

PR #2467, opened 2016-08-25
Fix the capabilities content type for QGIS and WMS 1.3

### LINKED ISSUE #2462: QGIS Layers in GeoMapFish 2.0 mobile not working
When one integrates a QGIS layer in GMF 2.0 (and probably also GMF 2.1), the /themes request shows a `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1053: ordinal not in range(128)` error for a _wms_getcap_cached URL get:

image

with the end of the message:
image

Probably because of a encoding problem because there is a <City> tag with an Umlaut in the GetCapabilites ([example](http://78.155.22.98:81/eleu/qgisserv?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities))

Probably reproducible with a QGIS layer in GMF Demo.

sbrunner:
fix  #2462

### DIFF
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -159,7 +159,7 @@ class Entry(object):
         if resp.get("content-type").split(";")[0].strip() not in \
                 ["application/vnd.ogc.wms_xml", "text/xml"]:
             error = "GetCapabilities from URL %s returns a wrong Content-Type: %s\n%s" % \
-                (url, resp.get("content-type"), content)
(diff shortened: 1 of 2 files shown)


ger-benjamin:
Okay but (at least) in case of the header is `text/cml`, `content` can use no-ascii caractere and you must encode it in utf-8. no ?

sbrunner:
To be tested ...

ger-benjamin:
Please merge

PR #2496, opened 2016-09-09
Fix components dependencies

gberaudo:


gberaudo:
Hmm, this is a duplicate of #2495.

PR #2497, opened 2016-09-09
Be able to use the CGXP interface without external server

sbrunner:
@gberaudo it should fix the editing interface


gberaudo:
LGTM

Issue #2499, opened 2016-09-12
themev1tov2 script broken!


ochriste:
running the themev1tov2 script fail:
```
No handlers could be found for logger "c2cgeoportal.pyramid_"
Emptying table main.layer_wmts.
Emptying table main.layer_wms.
Emptying table main.ogc_server.
Traceback (most recent call last):
  File "./.build/venv/bin/themev1tov2", line 11, in <module>
    sys.exit(main())
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/c2cgeoportal/scripts/themev1tov2.py", line 92, in main
    ogc_server(session)
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/c2cgeoportal/scripts/themev1tov2.py", line 144, in ogc_server
    transaction.commit()
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/transaction/_manager.py", line 123, in commit
    return self.get().commit()
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/transaction/_transaction.py", line 280, in commit
    reraise(t, v, tb)
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/transaction/_transaction.py", line 271, in commit
    self._commitResources()
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/transaction/_transaction.py", line 416, in _commitResources
    reraise(t, v, tb)
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/transaction/_transaction.py", line 388, in _commitResources
    rm.tpc_begin(self)
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/zope/sqlalchemy/datamanager.py", line 90, in tpc_begin
    self.session.flush()
  File "/var/www/vhosts/planv4-dev/private/och/epfl_authgeoportail/.build/venv/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2019, in flush
    self._flush(objects)
[...]

PR #2500, opened 2016-09-12
OGCServer url can not be null, OGCServer name must be unique

ochriste:
fix https://github.com/camptocamp/c2cgeoportal/issues/2499

### DIFF
--- a/c2cgeoportal/scripts/themev1tov2.py
+++ b/c2cgeoportal/scripts/themev1tov2.py
@@ -126,6 +126,11 @@ def ogc_server(session):
             image_type = "image/png"
         if url is None:
             url = "config://internal/mapserv"
+            name = u"source for {}".format(image_type)
+        else:
+            name = u"source for {} {}".format(url, image_type)
+        if is_single_tile:
+            name += u" with single_tile"
         identifier = (url, image_type, is_single_tile)
         if identifier not in unique_servers:
             unique_servers.add(identifier)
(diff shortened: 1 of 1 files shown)


sbrunner:
Thanks, that's gooks good :-)

c2cgeoportal/scripts/themev1tov2.py
```diff
@@ -129,17 +130,24 @@ def ogc_server(session):
         if identifier not in unique_servers:
             unique_servers.append(identifier)
             new_ogc_server = OGCServer()
-            new_ogc_server.url = server[0]
+            url = server[0]
+            if url is None:
+                url = 'config://local/mapserv'
```
sbrunner: should be `config://internal/mapserv`

c2cgeoportal/scripts/themev1tov2.py
```diff
@@ -154,8 +162,11 @@ def layer_v1tov2(session, layer):
         image_type = layer.image_type
         if layer.image_type is None:
             image_type = 'image/png'
+        url = layer.url
+        if layer.url is None:
+            url = 'config://local/mapserv'
```
sbrunner: same

ochriste:
corrected

sbrunner:
the name error comes from that the line https://github.com/camptocamp/c2cgeoportal/pull/2500/files#diff-ca696626af2c24c1c56983a6b182e8beR129
is incoherent with:
https://github.com/camptocamp/c2cgeoportal/blob/dbd2ce8cc92be857d421451e212ad969f5172dcc/c2cgeoportal/scripts/themev1tov2.py#L147
...

gberaudo:
LGTM

PR #2526, opened 2016-09-15
Put the fixed versions in a separate file

### LINKED ISSUE #2523: We shouldn't have version fixed in the makefile it should be in a separate file
To be sure that it will be updated during an update

sbrunner:
fix #2523


gberaudo:
LGTM

PR #2530, opened 2016-09-15
Support mako generated js files

### LINKED ISSUE #2457: Introducing variables in mobile.js.make is giving a build error
When GMF is used as different instances, one wants to set the center coordinates of the ngeo-mobile application depending on the vars of the build.

The variable cannot be introduced "normally" but needs the following in the CONST_Makefile (addition marked with +):

```
 .build/%.js: .build/%.json $(OL_JS_FILES) $(NGEO_JS_FILES) $(APP_JS_FILES) \
 +		rudaz/static-ngeo/js/%.js \
```

see also https://github.com/camptocamp/rudaz_geomapfish/commit/be96e005dbf1e016e90e8766a01ecdf0eaab6da3.

Version: GMF 2.0

sbrunner:
fix #2457


gberaudo:
LGTM

Issue #2188, opened 2016-05-03
Updating to 2.0.0rc2


kalbermattenm:
I just ran the update (`make -f michael.mk upgrade`) and the first command:

```
make -f michael.mk upgrade1
```

I got an error...

It seems we need some kind of "new" `srid` variable in some kind of new "mapserver" template (what for...??)

What should I do to get rid of that error ?

Any hint welcomed...??

```

 Recursing into mapserver
Traceback (most recent call last):
  File "C:\Python27\Lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Python27\Lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\git_projects\sitn_c2cgeoportal\.build\venv\Scripts\pcreate.exe\__main__.py", line 9, in <module>
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scripts\pcreate.py", line 18, in main
    return command.run()
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scripts\pcreate.py", line 88, in run
    return self.render_scaffolds()
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scripts\pcreate.py", line 135, in render
_scaffolds
    scaffold.run(self, output_dir, props)
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scaffolds\template.py", line 65, in run
    self.write_files(command, output_dir, vars)
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scaffolds\template.py", line 97, in writ
e_files
    template_renderer=self.render_template,
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scaffolds\copydir.py", line 103, in copy
_dir
    template_renderer=template_renderer, out_=out_)
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scaffolds\copydir.py", line 115, in copy
_dir
    template_renderer=template_renderer
  File "c:\git_projects\sitn_c2cgeoportal\.build\venv\lib\site-packages\pyramid\scaffolds\copydir.py", line 260, in subs
titute_content
    return template_renderer(content, v, filename=filename)
[...]

PR #2579, opened 2016-10-07
Remove http URL for tx host on Windows

kalbermattenm:
See https://github.com/camptocamp/c2cgeoportal/issues/2188#issuecomment-216537865

This is actually also working via https on Windows....

But I think it is OK to keep the `TX_HOST` variable in `CONST_Makefile` so that the URL is only hard coded once (by defining the variable), thus not to completely revert https://github.com/camptocamp/c2cgeoportal/pull/2196

Please review and (eventually) merge


sbrunner:
Thanks :-)

PR #2581, opened 2016-10-07
Fix the fix

gberaudo:
Follow up of https://github.com/camptocamp/c2cgeoportal/pull/2566.


sbrunner:
Thanks @gberaudo just one comment and you can merge :-)

PR #2612, opened 2016-10-24
Fix layer exclude

sbrunner:
Probably introduce by merge from 2.0


fredj:
LGTM

sbrunner:
Thanks @fredj :-)

Issue #2613, opened 2016-10-25
Template print A4 landscape is not valid


ger-benjamin:
Template print A4 landscape is not valid

These kind of comparison can't work: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/create/print/print-apps/%2Bpackage%2B/A4_Landscape.jrxml#L25

`<printWhenExpression><![CDATA[$P{debug} == 1]]></printWhenExpression>`

Must be:

`<printWhenExpression><![CDATA[$P{debug}]]></printWhenExpression>`

(Concern only the _A4 landscape_ template, others are all right)

@sbrunner 
I'll correct that but which branch should I target ? 1.6 ?
I will also do the modification into our demo

PR #2615, opened 2016-10-25
Print template A4L - Don't compare boolean with int

ger-benjamin:
Fix: #2613

Must be reported in 1.6+ branches


ger-benjamin:
No wait !

ger-benjamin:
Now it's correct and like others templates

PR #2629, opened 2016-11-07
Cleanup the tmp directory after npm install

### LINKED ISSUE #2601: clean up npm temporary files after install
after installation, npm leave quite some files in /tmp which are never cleaned (until reboot)

cf 
https://github.com/npm/npm/issues/6761
https://github.com/npm/npm/issues/6855

auto-cleanup in npm is set as "patch welcome", so we can without worry implement some manually clean up of anything npm\* in /tmp after npm has finished installing without error.

sbrunner:
fix  #2601


gberaudo:
Unfortunately I don't think it is a good solution since it would conflict (destroy) with:
- another run from the same user;
- run from other users on the same machine.

ochriste:
it would not cause any problem for other user as the folder owner would be different and the script would not be able to delete someone else folder

gberaudo:
@ochriste, it would still fail parallel builds by the same user...

ochriste:
yes, and what are the probability for that to happen?

PR #2635, opened 2016-11-14
Some small fix

sbrunner:


ger-benjamin:
In the commit "Add current route URL" you add a `current_route_url` function, in the next commit, you remove this function. Is that normal ?

sbrunner:
Oups, thanks @ger-benjamin

PR #2644, opened 2016-11-16
Disable geometry check per layer in 1.6

### LINKED ISSUE #2583: Disable geometry check for a layer (GMF 1.6)
Exclude some layers from the geometry check (st_issimple PostGIS) usually done for editing layers.

The proposition was to configure the excluded layers in the editing plugin c2cgeoportal config (probably in UImetadata):

All layers will be validated with the PostGIS function:

```
geometry_validation: true
```

All layers will be validated with the PostGIS function except the layers layer_a und layer_b:

```
geometry_validation: all_except:
- layer_a
- layer_b   
```
### To Do
- Implementation in GMF 1.6

Developement server: gmf-test.sig.cloud.camptocamp.net

juliensam:
Fix #2583

Added the `geometry_validation` UIMetadata to disable the check ST_IsValid and ST_IsSimple. Simply set it to False to disable it. Added tests and there's only one commit.

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -130,6 +130,7 @@ vars:
         - copy_to
         - lastUpdateDateColumn
         - lastUpdateUserColumn
+        - geometry_validation
 
     # The list of functionalities that can be configured
     # through the admin interface.
--- a/c2cgeoportal/tests/functional/test_layers.py
+++ b/c2cgeoportal/tests/functional/test_layers.py
@@ -105,7 +105,7 @@ class TestLayers(TestCase):
 
     def _create_layer(
             self, public=False, none_area=False, attr_list=False,
-            exclude_properties=False, ui_metadatas=[]):
+            exclude_properties=False, ui_metadatas=[], geom_type=False):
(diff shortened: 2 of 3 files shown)


juliensam:
Ready for review

sbrunner:
Thanks @juliensam :-)

sbrunner:
I think that this line: should be moved https://github.com/juliensam/c2cgeoportal/blob/ecde7f5dc5fe73a2660c90840d127f5d181c5b23/c2cgeoportal/views/layers.py#L326...

To get something like this:
```
            if self._get_ui_metadata(
                    layer, "geometry_validation", str(self.settings.get("geometry_validation", "false"))).lower() != "false":
                self._validate_geometry(spatial_elt)
```
and moved in a method to don't have duplicated code :-)

sbrunner:
In other word the geometry_validation validation settings will be the default value...

juliensam:
I'm thinkig of adding a `_get_validation_setting` method in the Layer model that returns a boolean. How does that sound?

juliensam:
Or should it be in LayerV1?

juliensam:
Oups! no the _get_ui_metadata is in the view. Scratch that.

sbrunner:
Thanks :-)

juliensam:
Ready to merge

PR #2689, opened 2016-11-28
Do not use contracted form

sbrunner:
```
git-grep-sed "what's" 's/\([wW]hat\)'s/\1 is/g'
git-grep-sed "here's" 's/\([hH]ere\)'s/\1 is/g'
git-grep-sed "that's" "s/\([tT]hat\)'s/\1 is/g"
git-grep-sed "it's" "s/\([Ii]t\)'s/\1 is/g"
git-grep-sed "don't" "s/\([dD]o\)n't/\1 not/g"
git-grep-sed "won't" "s/\([wW]\)on't/\1ill not/g"
git-grep-sed "does't" "s/\([dD]\)oes't/\1oes not/g"
git-grep-sed "doesn't" "s/\([dD]\)oesn't/\1oes not/g"
git-grep-sed "isn't" "s/\([iI]\)sn't/\1s not/g"
git-grep-sed "can't" "s/\([cC]\)an't/\1annot/g"
git-grep-sed "shouldn't" "s/\([sS]\)houldn't/\1hould not/g"
```
where `git-grep-sed` is:
```
#!/usr/bin/python

import sys
import subprocess


if len(sys.argv) != 3 or sys.argv[1] in ('-h', '--help'):
    print("Usage: {} GREP SED".format(sys.argv[0]))
    print("")
    print("GREP: The git grep expression.")
    print("SED: The sed expression.")
    exit(0)

try:
    for file_name in subprocess.check_output(["git", "grep", "-i", "-l", "--full-name", sys.argv[1]]).strip().split("\n"):
        subprocess.check_call(["sed", "-i", "-e", sys.argv[2], file_name])
except subprocess.CalledProcessError as e:
    print(e)
    exit(1)
```


rbovard:
You miss some cases like :
* could not > couldn't
* must not > mustn't
* would not > wouldn't

And some non negative cases like:
* have > 've
* had > 'd
* are > 're
* would > 'd

sbrunner:
> You miss some cases like :
> ...

We don't have any case :-)

> And some non negative cases like:
> ...

Thank's, fixed :-)

rbovard:
~You're~ You are welcome :)

sbrunner:
:-)

PR #2695, opened 2016-12-01
On upgrade we don't use /tmp but CONST_create_template

pgiraud:
When we use the "upgrade" command, the create scaffold doesn't write files into /tmp any more.
https://github.com/camptocamp/c2cgeoportal/pull/2396/files
Old paths need to be fixed as well.


sbrunner:
Looks good, il should be merge if the migration from 1.6 to 2.1 is a success :)

PR #2699, opened 2016-12-02
Fix the documentation and upgrade instructions

pgiraud:


c2cgeoportal/scripts/c2ctool.py
```diff
@@ -384,10 +384,10 @@ def step1(self):
         else:
             print("")
             print(self.color_bar)
-            print(
-                "Apply the manual migration steps based on what is in the CONST_CHANGELOG.txt file"
-                " (listed in the `changelog.diff` file)."
-            )
+            text = "Apply the manual migration steps based on what is in the CONST_CHANGELOG.txt file"
+                   " (listed in the `changelog.diff` file)."
+            print(colorize(text, RED)
```
sbrunner: I think that red isn't the better color, shouldn't we use yellow?
Also it should be modified for each steps.

doc/integrator/legacy_cgxp.rst.mako
```diff
@@ -3,107 +3,112 @@
...
-The OGC proxy is deprecated because with modern browser it is not required to have it.
 
-Then you should remove it from the CGXP ``js`` interface files (``<package>/templates/<interface>.js``):
+The OGC proxy is deprecated because with modern browsers it is not required anymore.
+You can simply remove the following line in  ``<package>/templates/desktop.js``.
 
 .. code:: diff
 
-    - OpenLayers.ProxyHost = "${'$'}{request.route_url('ogcproxy') | n}?url=";
+  OpenLayers.ProxyHost = "${'$'}{request.route_url('ogcproxy') | n}?url=";
```
sbrunner: If you remove the `-` you should also teplace `diff` by `javascript`...

doc/integrator/legacy_cgxp.rst.mako
```diff
@@ -3,107 +3,112 @@
...
 
-The ``externalWFSTypes`` do not exist anymore than you should remove the following line
-from the CGXP ``js`` interface files (``<<package>/templates/<interface>.js``)
+``externalWFSTypes`` does not exist anymore so you should remove the following line
+from ``<package>/templates/desktop.js``
 
 .. code:: diff
 
-    - externalWFSTypes: ${'$'}{externalWFSTypes | n},
+  externalWFSTypes: ${'$'}{externalWFSTypes | n},
```
sbrunner: same as above

pgiraud:
> You quote CGXP and ngeo with only one back quote, is it wanted?
> To have codes stale in rst files you should use two back quote...

This is done on purpose. I prefer to use double quote for paths or file names.
screenshot from 2016-12-05 09 50 05

pgiraud:
> You replace in some place <interface> by desktop, I do is because it's valid for any interface ...

In our case, we ask the integrator to replace "index.html" by "desktop.html" and same for the ".js" file. And we don't explain what an "interface" is. This is why I didn't keep "<interface>".

sbrunner:
I don't know for this line
https://github.com/pgiraud/c2cgeoportal/blob/9cdf738a22a4e899974f108b37adb1e62574a366/doc/integrator/legacy_cgxp.rst.mako#L82
But this https://github.com/pgiraud/c2cgeoportal/blob/9cdf738a22a4e899974f108b37adb1e62574a366/doc/integrator/legacy_cgxp.rst.mako#L90 and that https://github.com/pgiraud/c2cgeoportal/blob/9cdf738a22a4e899974f108b37adb1e62574a366/doc/integrator/legacy_cgxp.rst.mako#L97 lines should be done for all CGXP interfaces ...

pgiraud:
Here's what I did
https://github.com/pgiraud/c2cgeoportal/blob/b87a4c542b67e63c971167d5788c757f8b042b6f/doc/integrator/legacy_cgxp.rst.mako#L86-L88

I left this since it's only relevant for the desktop.
https://github.com/pgiraud/c2cgeoportal/blob/b87a4c542b67e63c971167d5788c757f8b042b6f/doc/integrator/legacy_cgxp.rst.mako#L94-L95

sbrunner:
If you have exactly the original template it's true but if not it  can be not ...

PR #2711, opened 2016-12-08
Some changes found during an upgrade...

sbrunner:


c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -61,7 +61,7 @@ You should also add a Layer group named 'background' with the background layers.
     cp CONST_create_template/{{package}}/static-ngeo/less/mobile.less {{package}}/static-ngeo/less
 
 7. The checkers will be refactored, you should do at least the following
-   change in your vars_{{package}}.yaml:
+   change in your `vars_{{package}}.yaml` (be careful if you use legacy components):
```
ger-benjamin: chnage -> chnages (with the 's')

c2cgeoportal/views/entry.py
```diff
@@ -584,7 +584,9 @@ def _fill_internal_wms(self, l, layer, wms, wms_layers, errors):
         else:
             if self.default_ogc_server.type != OGCSERVER_TYPE_GEOSERVER:
                 errors.add(
-                    "The layer '{0!s}' is not defined in WMS capabilities".format(wmslayer)
+                    "The layer '{}' ({}) is not defined in WMS capabilities from '{}'".format(
+                        wmslayer, layer.name, self.default_ogc_server.name
```
ger-benjamin: I think that's egal for a human but `layer.layer` isn't more precise ?

And don't we do the same trick (layer.name -> layer.name) for wmts layers ?

ger-benjamin:
Some comments but that's look good. Thanks :-)

sbrunner:
Thanks @ger-benjamin :-)

c2cgeoportal/views/entry.py
sbrunner: layer.layer is the same as `wmslayer` than it's displayed, and `layer.name` can be helpful when we have more than one layer with the same `layer`...

I don't understand, this is for layerv1, for layerwms and layerwmts it's working with layer(s) and name for a long time :-)

PR #2733, opened 2017-01-04
Admin dimension

### LINKED ISSUE #2632: Strange error with psycopg2
How to reproduce:

 - Go on: https://planv4-dev.epfl.ch/wsgi/admin/Dimension
 - Try to add a dimension on a layer

You'll got this error: 

⇝ 
```
IntegrityError: (psycopg2.IntegrityError) null value in column "layer_id" violates not-null constraint DETAIL: Failing row contains (217, floor, null, null, null). [SQL: 'INSERT INTO main.dimension (name, value, description, layer_id) VALUES (%(name)s, %(value)s, %(description)s, %(layer_id)s) RETURNING main.dimension.id'] [parameters: {'layer_id': None, 'name': u'floor', 'value': None, 'description': None}]
```

sbrunner:
Fix #2632

### DIFF
--- a/c2cgeoportal/models.py
+++ b/c2cgeoportal/models.py
@@ -1017,7 +1017,8 @@ class Dimension(Base):
     def __init__(self, name="", value="", layer=None):
         self.name = name
         self.value = value
-        self.layer = layer
+        if layer is not None:
+            self.layer = layer
 
     def __unicode__(self):  # pragma: no cover
         return self.name or u""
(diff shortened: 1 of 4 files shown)


c2cgeoportal/models.py
```diff
@@ -1017,7 +1017,8 @@ class Dimension(Base):
     def __init__(self, name="", value="", layer=None):
         self.name = name
         self.value = value
-        self.layer = layer
+        if layer is not None:
+            self.layer = layer
```
ger-benjamin: It's really better to have an `undefined` value instead of a `None` value ?

sbrunner: If I set the layer to `None` form-alchemy will not update it with the form data :-(

PR #2738, opened 2017-01-10
Less failing message in i18n generation, non fatal error messages in …

sbrunner:
…yellow


sbrunner:
I don't think that's the goal of i18n generation to block the build...

PR #2739, opened 2017-01-11
karma-phantomjs-launcher is no more an ngeo dependency

sbrunner:


sbrunner:
@gberaudo do you want to have a second look?

Issue #2756, opened 2017-01-17
Upgrade 1.6 to 2.1.0rc7


ger-benjamin:
Upgrade demo 1.6 to 2.1.0rc7 with ngeo for all interfaces and printv3
Current test url: https://testgmf.sig.cloud.camptocamp.net/bge (no background layer for now)

## 1) Minor note: 
- [x] Why ngeo release are named: `2.1.0-rc.7` and in c2c `2.1.0rc7`

## 2) upgrade:
 - No errors
 
## 3) upgrade2:

- [x] **3-A) Warnings**
````
.build/venv/bin/treeitem_uniquename
No handlers could be found for logger "c2cgeoportal.pyramid_"
/home/bgerber/demo_geomapfish/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/sql/sqltypes.py:219: SAWarning: Unicode type received non-unicode bind param value 'osm_time-2'. (this warning may be suppressed after 10 occurrences)
  (util.ellipses_string(value),))
/home/bgerber/demo_geomapfish/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/sql/sqltypes.py:219: SAWarning: Unicode type received non-unicode bind param value 'Enseignement-2'. (this warning may be suppressed after 10 occurrences)
````

**3-B) After to apply the changelog and do a new upgrade2**

- [x] *3-B-1) ``.build/venv/bin/themev1tov2``*
````
sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (psycopg2.IntegrityError) null value in column "image_type" violates not-null constraint
DETAIL:  Failing row contains (134, http://wmts.geo.admin.ch/1.0.0/WMTSCapabilities.xml?lang=fr, ch.astra.ausnahmetransportrouten, ch.astra.ausnahmetransportrouten, 21781_26, null).
[...]

PR #2760, opened 2017-01-18
Use bracket in format function

ger-benjamin:
Partially fix: #2756 
Fix: *4-A) Error when checkers don't pass*


c2cgeoportal/scripts/c2ctool.py
```diff
@@ -200,10 +200,10 @@ def test_checkers(self):
                 }
             )
             if resp.status < 200 or resp.status >= 300:
-                return False, "\n".join(
+                return False, "\n".join([
                     "Checker error:"
```
sbrunner: And add the missing coma :-)

PR #2768, opened 2017-01-20
Use scripts instead of bin for Windows

kalbermattenm:
Just some changes in CONST_CHANGELOG.txt_tmpl (in order for me not to waste everybody's time...)

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -161,6 +161,8 @@ You should also add a Layer group named 'background' with the background layers.
     .build/venv/bin/alembic --config alembic.ini upgrade 7d271f4527cd
     .build/venv/bin/treeitem_uniquename
 
+    If you are using Windows, then please replace `bin` with `scripts` in the paths
+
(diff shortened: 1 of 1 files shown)


c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -161,6 +161,8 @@ You should also add a Layer group named 'background' with the background layers.
     .build/venv/bin/alembic --config alembic.ini upgrade 7d271f4527cd
     .build/venv/bin/treeitem_uniquename
 
+    If you are using Windows, then please replace `bin` with `scripts` in the paths
```
sbrunner: What about adding the `.cmd`?

kalbermattenm:
Actually, it is a Windows executable (`alembic.exe`), thus you do not need to write it using the file extension.

In Windows, running
`> alembic.exe`
or running
`> alembic`
is the same...

sbrunner:
And `make -f <user>.mk .build/venv/scripts/alembic` will also works?

kalbermattenm:
Yes it will

sbrunner:
@kalbermattenm thanks :-)

PR #2772, opened 2017-01-24
Fix the CGXP interface test

sbrunner:


ochriste:
c'est a cause de ca que ca fail?
https://travis-ci.org/camptocamp/c2cgeoportal/builds/194870963#L5234

sbrunner:
if don't exists in make :-)

PR #2791, opened 2017-02-02
Set default encoding to utf-8

ger-benjamin:
@sbrunner


ger-benjamin:
Should be for 2.1 too

PR #2796, opened 2017-02-06
Fix openlayers submodule url

arnaud-morvan:
Openlayers v2 github repository has be renamed.
This update the submodule url.


ochriste:
merci

sbrunner:
@arnaud-morvan it's not good to merge when not all the travis checks pass...

PR #2806, opened 2017-02-07
Use extern for angular 1.6

sbrunner:


sbrunner:
Yes but we want to fix to a stable version who it isn't released ...

PR #2809, opened 2017-02-09
Fix the deploy

sbrunner:


fredj:
looks good but I'm not sure that the `/tmp/np` still works

sbrunner:
I don't know but it don't hurt :-)

PR #2811, opened 2017-02-09
Add simple session creation

sbrunner:


sbrunner:
+1, I will add a test

PR #2818, opened 2017-02-15
Extract enum attributes values for i18n

ger-benjamin:
Fix: #2819

### DIFF
--- a/c2cgeoportal/lib/lingua_extractor.py
+++ b/c2cgeoportal/lib/lingua_extractor.py
@@ -170,31 +170,73 @@ class GeoMapfishConfigExtractor(Extractor):  # pragma: no cover
             config = yaml.load(config_file)
             # for application config (.build/config.yaml)
             if "vars" in config:
-                return [
-                    Message(
-                        None, raster_layer, None, [], u"", u"",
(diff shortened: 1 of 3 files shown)


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -187,13 +187,18 @@ def _collect_app_config(self, config, filename):
             for raster_layer in config["vars"].get("raster", {}).keys()
         ]
         # Collect layers enum values (for filters)
+        settings = get_config(".build/config.yaml")
+        from c2cgeoportal.pyramid import initDBSessions
```
sbrunner: missing _

ger-benjamin:
Tested with the demo. It generate lines like this int the .pot (three layers use the same table's column, so it's normal that we have three description):

````
#: ./.build/config.yaml:/layers/osm_open/values/type/bench                       
#: ./.build/config.yaml:/layers/buildings18/values/type/bench                    
#: ./.build/config.yaml:/layers/landuse18/values/type/bench                      
msgid "bench"                                                                    
msgstr ""
````

sbrunner:
+1 :-)

PR #2820, opened 2017-02-16
Don't use 'directory' on deploy

ger-benjamin:
Because that's not the directory to deploy the code


sbrunner:
This should be on 2.1 :-)

PR #2823, opened 2017-02-17
Code style

sbrunner:


sbrunner:
Because of https://docs.python.org/2/library/functions.html#vars and https://docs.python.org/2/library/functions.html#dir :-)

ger-benjamin:
That's logic. Thanks.

PR #2839, opened 2017-02-21
Rename docker-compose.yaml to docker-compose.yml

fredj:
docker-compose output the following warning:
```
WARNING: Please be aware that .yml is the expected extension in most cases,
and using .yaml can cause compatibility issues in future.
```


sbrunner:
??? docker developers should know that's we are after 1995 ???

PR #2849, opened 2017-02-22
Add timeattribute for query

ger-benjamin:
For specs https://github.com/camptocamp/c2cgeoportal/wiki/Spec-%232652-Time-in-Query


ger-benjamin:
What is the content of this metadata ? A String ?
"timeAttribute" looks like oder "\*Attribute**s**" but these "attributes" are list.
Can't we have a better name ?

sbrunner:
For me the layer is OK, and the sting is effectively the good name.

PR #2851, opened 2017-02-22
Fix the headers configuration fix, ...

sbrunner:


c2cgeoportal/scaffolds/update/CONST_config-schema.yaml
```diff
@@ -262,6 +268,10 @@ mapping:
                     base_url:
                         type: str
                         required: True
+                    allowed_hosts:
+                        type: seq
+                        sequence:
+                        -   type: str
```
ger-benjamin: I'ts an array ? (because of the `-`)

sbrunner: Yes :-)

PR #2860, opened 2017-02-28
Set CGXP_API default to false

### LINKED ISSUE #2832: In new project, CGXP_API should be false by default

fredj:
fixes #2832


sbrunner:
Can you remember me why it should be false by default?

fredj:
Because `CGXP_INTERFACES` is disabled by default.

sbrunner:
But we steal have the CGXP API ...

PR #2872, opened 2017-03-20
Less Windows requirements

kalbermattenm:
Starting with pip >= 9, binary dependencies can now be installed through pip (as long as the wheel enables it...).

I removed the potential dependencies from the specific Linux/Windows requirement files and added their current "Linux" version to the common version file.

I tested all these packages in an independent virtual environment. I did not have any problem running the following commands:

```
Scripts\pip.exe --no-cache-dir install pillow==3.4.2
Scripts\pip.exe --no-cache-dir install psycopg2==2.6.2
Scripts\pip.exe --no-cache-dir install pyproj==1.9.5.1
```

I also updated the documentation.

only Shapely remains problematic...


fredj:
thanks @kalbermattenm !

PR #2874, opened 2017-03-22
Allow to pass permission to add_interface

arnaud-morvan:


sbrunner:
Travis tests should be fixed...

PR #2878, opened 2017-03-23
Use max_age 0 if no cache is asked

ger-benjamin:
It seems that with "response.cache_control.no_cache" to True, the cache is still used...

My fix seems to work for https://github.com/camptocamp/pully_geomapfish/issues/90


sbrunner:
I prefer to change the max_age directly than doing a post modification...

sbrunner:
The tests should also be updated...

ger-benjamin:
> I prefer to change the max_age directly than doing a post modification...

I don't understand or that doesn't wortk with the "no_cache" requests

ger-benjamin:
@sbrunner thanks for the explanation. I've updated my code.

sbrunner:
Thanks :-)

PR #2906, opened 2017-04-21
Generate a on-the-fly path separator

kalbermattenm:
This is needed because in Windows only double backslash might be used.

This is mainly a backport of what we have done with @sbrunner this morning in the SITN project.


sbrunner:
Thanks :-)

PR #2914, opened 2017-05-02
Fix the desktop interface debug mode

sbrunner:


ger-benjamin:
What does \1 means ? (last modified line)

sbrunner:
@ger-benjamin group one :-)

PR #2917, opened 2017-05-02
Lowercase layer to filter them

### LINKED ISSUE #2916: Private uppercased layers are listed in public wms capabilities
On 1.6 (et perhaps upper)

Without login, you can see the wms layers in the getcapabilities file if layers are named with uppercases. That because we normalize (lowercase) the names of the layers from mapserver but not from DB.

(Look also in https://jira.camptocamp.com/browse/GEO-127)

ger-benjamin:
Fix #2916


ger-benjamin:
@sbrunner Now it's look good

Issue #2929, opened 2017-05-11
Secured layers are visible in GetCapabilities


rbovard:
There is a regression in `2.1.4` compared to `1.6.8` because secured layers are visible in GetCapabilities.

* Without login
  * Private layers are in GetCapabilities --> **WRONG**
  * I can add them in QGIS and only the legend is displayed (no data) --> The legend should be protected no?
* With login
  * Private layers are shown in GetCapabilities --> **OK**
  * I can add them in QGIS and the data and the legend are displayed --> **OK**

On `1.6.8` it's fine:
* Without login
  * Private layers are not in GetCapabilities --> **OK**
* With login
  * Private layers are in GetCapabilities --> **OK**
  * I can add them in QGIS and the data and the legend are displayed --> **OK**

Original post: https://github.com/camptocamp/c2cgeoportal/issues/2921#issuecomment-300499084

sbrunner:
Duplicated with #2921

rbovard:
No it's not the same, in my case the secure layers are visible without login and @gnerred has trouble to see secure layers with login.

I've volontary opened this issue to keep those issues distinct.

sbrunner:
> I can add them in QGIS and only the legend is displayed (no data) --> The legend should be protected no?

no, we never do this

rbovard:
Ok, thanks for the info.

BTW could you please reopen this issue, since @ger-benjamin is working on it.

rbovard:
Thanks, it's working.

PR #2932, opened 2017-05-11
Filter WMS capabilities with correct OGC server

ger-benjamin:
Part of #2929 

TODO:
 - [x] Fix DetachedInstanceError: Instance <LayerWMS at 0x7feae2836290> is not bound to a Session; ~~attribute refresh operation cannot proceed -> maybe come frome the `cache_on_argument` It seems to work with a single argument but not with an array  ?~~ -> must to expung objects in session
 - [x] add tests

(I've not tested the things about doing a curl to get capabilities with basic auth.)

### DIFF
--- a/c2cgeoportal/lib/filter_capabilities.py
+++ b/c2cgeoportal/lib/filter_capabilities.py
@@ -123,6 +123,13 @@ def wms_structure(wms_url, host):
         log.exception(error)
         raise HTTPBadGateway(error)
 
+    except SyntaxError:  # pragma: no cover
+        error = "WARNING! an error occured while trying to " \
+            "read the mapfile and recover the themes."
+        error = "{0!s}\nurl: {1!s}\nxml:\n{2!s}".format(error, wms_url, content)
+        log.exception(error)
+        raise HTTPBadGateway(error)
+
 
 def enable_proxies(proxies):  # pragma: no cover
     old_prepare_input_source = saxutils.prepare_input_source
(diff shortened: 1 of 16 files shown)


sbrunner:
You should also add a test case :-)

c2cgeoportal/lib/filter_capabilities.py
```diff
@@ -152,15 +161,19 @@ def caching_prepare_input_source(source, base=None):
     saxutils.prepare_input_source = caching_prepare_input_source
 
 
-def filter_capabilities(content, role_id, wms, wms_url, headers, proxies):
+def filter_capabilities(content, role_id, wms, url, headers, proxies, request, ):
```
ger-benjamin: last `, ` works but looks strange

c2cgeoportal/tests/functional/__init__.py
```diff
@@ -63,65 +64,65 @@
 caching.init_region({"backend": "dogpile.cache.memory"})
 
 
+def cleanup_db():
```
ger-benjamin: That clean really all ?  What about metadata, tsearch... ? Perhaps layer ? (or that comes with treeitem ? but both layerv1 and v2 ?)

sbrunner: TreeItem cover all the tree, FullTextSearch and Shorturl will be added.

c2cgeoportal/lib/filter_capabilities.py
sbrunner: fixed :-)

ger-benjamin:
Can I merge ?

PR #2957, opened 2017-05-22
Fix multi-language mapserverless projects

pvalsecc:
When MAPSERVER=FALSE, we should not depend on mapserver.timestamp
to build the translations.

Pass the list of languages when creating the TX configuration so that
it is not hardcoded to fr and de anymore.

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -272,12 +272,15 @@ I18N_SOURCE_FILES += $(APP_HTML_FILES) \
 	$(APP_JS_FILES) \
 	$(APP_DIRECTIVES_PARTIALS_FILES) \
 	.build/config.yaml
-ifeq ($(DOCKER),FALSE)
+ifeq ($(DOCKER), FALSE)
 # The theme from the database
 I18N_SOURCE_FILES += development.ini
 endif
-I18N_DEPENDENCIES += $(MAP_FILES) \
-	project.yaml
+I18N_DEPENDENCIES += $(VENV_BIN)/pot-create$(PYTHON_BIN_POSTFIX)
+ifeq (${MAPSERVER}, TRUE)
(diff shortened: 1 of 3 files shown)


pvalsecc:
Depends on an upgrade to https://github.com/camptocamp/ngeo/pull/2652 to be effective

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -708,14 +708,17 @@ $(PACKAGE)/locale/$(PACKAGE)-$(L10N_SERVER_POSTFIX).pot: \
 	[ ! -f $@ ] || sed -i '/^"PO-Revision-Date: /d' $@
 	[ ! -f $@ ] || chmod go+r $@
 
-.PRECIOUS: $(PACKAGE)/locale/$(PACKAGE)-$(L10N_CLIENT_POSTFIX).pot
-$(PACKAGE)/locale/$(PACKAGE)-$(L10N_CLIENT_POSTFIX).pot: \
-		lingua-client.cfg \
+TRANSLATION_DEPENDENCIES = lingua-client.cfg \
```
sbrunner: is there a relay deference between TRANSLATION_DEPENDENCIES and I18N_DEPENDENCIES?

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -708,14 +708,17 @@ $(PACKAGE)/locale/$(PACKAGE)-$(L10N_SERVER_POSTFIX).pot: \
...
 		$(I18N_DEPENDENCIES) \
 		$(I18N_SOURCE_FILES) \
 		$(VENV_BIN)/pot-create$(PYTHON_BIN_POSTFIX) \
-		.build/mapserver.timestamp \
 		.build/config.yaml
+ifeq (${MAPSERVER},TRUE)
+TRANSLATION_DEPENDENCIES += .build/mapserver.timestamp
+endif
```
sbrunner: Should be moved at line 180

PR #2972, opened 2017-06-09
Fetch the Git submodules and fix master

sbrunner:

PR #2973, opened 2017-06-09
Add git submodule --init instruction and fix indentation

Vampouille:


sbrunner:
This is generated pages :-/ , see also #2972

PR #2978, opened 2017-06-14
Use the fixed version also in the Docker build package

sbrunner:
To avoid this kind of unwanted update: https://github.com/camptocamp/c2cgeoportal/pull/2977/files

### DIFF
--- a/docker/build/Dockerfile
+++ b/docker/build/Dockerfile
@@ -11,7 +11,7 @@ RUN \
   apt-get clean && \
   rm --recursive --force /var/lib/apt/lists/*
 
-COPY *-requirements.txt /tmp/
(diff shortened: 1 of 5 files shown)


docker/build/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN \
   apt-get clean && \
   rm --recursive --force /var/lib/apt/lists/*
 
-COPY *-requirements.txt /tmp/
+COPY *.txt /tmp/
```
ger-benjamin: It isn't too wide ? What happens if someone create a "readme.txt" or something like this ?

sbrunner: I will change that :-)

sbrunner: Updated :-)

PR #2982, opened 2017-06-15
Fix accent and port in add_url_params

sbrunner:


c2cgeoportal/lib/__init__.py
```diff
@@ -224,25 +222,19 @@ def get_typed(name, value, types, request, errors):
...
-        query.append(spliturl.query)
-    prepared_params = []
-    for param in params.items():
-        if isinstance(param[1], unicode):
-            prepared_params.append([param[0], param[1].encode("utf-8")])
-        else:
-            prepared_params.append(param)
-
-    query.extend([urllib.urlencode(dict([param])) for param in prepared_params])
+    query = dict([(k, v[-1]) for k, v in urlparse.parse_qs(
```
ger-benjamin: You use the same mechanisme 3 time. Perhaps that's more lisible with a lambda function:
`encode = lambda val: val.encode('utf-8') if isinstance(val, unicode) else val`

sbrunner: +1 Thanks :-)

PR #2984, opened 2017-06-15
Some fix for Docker

sbrunner:


docker/test-external-db/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM camptocamp/postgres:latest
+FROM camptocamp/postgis:9.5
```
pvalsecc: camptocamp/postgres is the official one.

docker/test-external-db/10_init.sql
```diff
@@ -1,5 +1,3 @@
-CREATE EXTENSION postgis;
```
pvalsecc: CREATE EXTENSION IF NOT EXISTS postgis;

PR #2985, opened 2017-06-16
An other issue with the Turing apple

sbrunner:


c2cgeoportal/scripts/import_ngeo_apps.py
```diff
@@ -210,6 +210,7 @@ def main():
 
             data = re.sub("app", "{{package}}", data)
             data = re.sub(re.escape("{{package}}lication"), "application", data)
+            data = re.sub(re.escape("{{package}}lication"), "apple", data)
```
pvalsecc: s/lication/le/

PR #2998, opened 2017-06-27
Fix trying to get name of None external_ogc_server.

ochriste:


ochriste:
fix https://github.com/camptocamp/cgxp/issues/1119

PR #3007, opened 2017-07-21
Codacy

sbrunner:


c2cgeoportal/views/mapserverproxy.py
```diff
@@ -47,6 +47,14 @@ class MapservProxy(OGCProxy):
 
     def __init__(self, request):
         OGCProxy.__init__(self, request)
+        self.user = self.request.user
+        self.external = bool(self.request.params.get("EXTERNAL"))
+
+        # params hold the parameters we"re going to send to MapServer
+        params = dict(self.request.params)
```
pvalsecc: This change is not equivalent. You don't delete role_id and user_id anymore...

PR #3011, opened 2017-07-28
Be more strict on i18n file generation

sbrunner:


sbrunner:
Reviewed by @gberaudo :-)

PR #3027, opened 2017-08-11
Upgrade to Python3 and disable the old admin interface

sbrunner:


docker/build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:2
+FROM python:3
```
gberaudo: Why not 3.6?

sbrunner: Because 3.6 is not available on Debian stable, and we want the same version every where :-)

c2cgeoportal/lib/__init__.py
```diff
@@ -220,23 +221,19 @@ def get_typed(name, value, types, request, errors):
 
 
 def add_url_params(url, params):
-    if len(params.items()) == 0:
+    if len(list(params.items())) == 0:
```
fredj: can be simplified to `if len(params) == 0:`

PR #3029, opened 2017-08-14
[wip] Use json from the standard library instead of simplejson

fredj:


sbrunner:
+1, should also be removed form the requirements :-)

fredj:
simplejson is still used in one place

PR #3031, opened 2017-08-14
fix function get_ogc_server_wfs_url_ids

jwkaltz:
This was discovered when trying to get TinyOWS to work in GEO-368. Changed the function "get_ogc_server_wfs_url_ids" to use the input as key (instead of using the looked up url), the goal being to look up the layer ids for these keys, not for the urls: this is how the caller filter_wfst_capabilities() expects the dictionary.

### DIFF
--- a/c2cgeoportal/lib/filter_capabilities.py
+++ b/c2cgeoportal/lib/filter_capabilities.py
@@ -57,7 +57,9 @@ log = logging.getLogger(__name__)
 @cache_region.cache_on_arguments()
 def get_protected_layers(role_id, ogc_server_ids):
     q = get_protected_layers_query(role_id, ogc_server_ids, what=LayerWMS, version=2)
-    return {r.id: r for r in q.all()}
+    results = q.all()
+    DBSession.expunge_all()
+    return {r.id: r for r in results}
 
 
 @cache_region.cache_on_arguments()
@@ -74,7 +76,9 @@ def get_private_layers(ogc_server_ids):
 @cache_region.cache_on_arguments()
 def get_writable_layers(role_id, ogc_server_ids):
     q = get_writable_layers_query(role_id, ogc_server_ids)
-    return {r.id: r for r in q.all()}
+    results = q.all()
+    DBSession.expunge_all()
+    return {r.id: r for r in results}
 
 
 @cache_region.cache_on_arguments()
--- a/c2cgeoportal/views/mapserverproxy.py
+++ b/c2cgeoportal/views/mapserverproxy.py
@@ -32,7 +32,6 @@ import logging
 
 from pyramid.view import view_config
 
-from c2cgeoportal.lib import get_url2
 from c2cgeoportal.lib.caching import get_region, NO_CACHE, PUBLIC_CACHE, PRIVATE_CACHE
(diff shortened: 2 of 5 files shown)


jwkaltz:
@sbrunner I believe this fixes the first request to get the WFS capabilities. I noticed that the following requests, via Cache, do not work - but that is a separate issue (code error is in a different place, dealing with cache), so I think we should address the cache issue in a separate branch.
I noticed that the Travis CI build fails - but I think this has nothing to do with the code changes of this branch - what do you think?

sbrunner:
This Pull request will fix you issue :-)
https://github.com/camptocamp/c2cgeoportal/pull/3032

jwkaltz:
Actually I should get this to work against a 2.2 version, not against master - 2.2 is the target of the customer site. If we get 2.2 to work and the fixes are also appropriate for master branch, we should apply these then to master, not the other way around. So I am closing this PR, I will make a new one.

sbrunner:
I just change the base to 2.2..., no needed to create an other pull request, just edit the current pull request to change the base :-)

ger-benjamin:
Is there the same issue in the method "get_ogc_server_wms_url_ids" (just above the same method for WFS) ?

jwkaltz:
OK thanks for changing the base branch to the right one.
But now, do we need something else to get CI to work on the 2.2 branch (if I understand correctly, the CI fails not because of the changes of this fix, but fails on 2.2 in general)?

sbrunner:
see: https://travis-ci.org/camptocamp/c2cgeoportal/builds/265064481#L8180-L8181

jwkaltz:
Thanks for the hint, @sbrunner  ... I am not yet used to reading Travis reports :crying_cat_face: 
Let's see if the CI is OK now.

jwkaltz:
Thanks for your comment, @ger-benjamin . Looking at that code, it is not so plausible that is does not work as expected - I think the problem rather lies in the fact that "get_ogc_server_wfs_url_ids" is used in two different manners, and the two functions that call it have different expectations about what the function should return. So actually, I need to do some more debugging...

sbrunner:
up ...

jwkaltz:
@sbrunner the CI fails with an error "node install.js" - do you have an idea why?

sbrunner:
I just restart the build and now it's OK, You just needs to squash your commits before merging :-)

jwkaltz:
@sbrunner I now also added the same fix to get_protected_layers. CI is OK. What do you think, shall we merge this into 2.2 (and 2.1 as well, I suppose?)

sbrunner:
* It you need it in 2.1 you should do a rebase on the 2.1 branch and change the pull request base.
* Before merging you should squash all your commits

To make it available on the 2.2 (and master) branch I will merge the 2.1 branch into the 2.2 branch.

jwkaltz:
I now squashed everything into one commit. We need it primarily for GEO-368, which is based on 2.2 - so 2.2 as request base is correct.

sbrunner:
Can we merge that?

jwkaltz:
Yes, in my opinion we can merge this PR into 2.2. - so is it OK if I do it?

sbrunner:
@jwkaltz then just do it :-)

Issue #3038, opened 2017-08-15
Use localhost to run checkers on upgrade step 5


ger-benjamin:
To be able to run checkers and complete upgrade step 5

PR #3039, opened 2017-08-15
Use localhost to call checkers upgrade step 5

ger-benjamin:
FIX: #3038


ger-benjamin:
@sbrunner any idea of the travis error ?

PR #3042, opened 2017-08-16
Allows empty dimensions

sbrunner:


arnaud-morvan:
For me it make sense to have NULL in database for such cases, so the test with None make sense. But I do know how is handled "" by the forms system.

PR #3044, opened 2017-08-16
Remove required on dimension.value

arnaud-morvan:
As layer dimension config value take precedence over global dimensions values,
it needed to be able set NULL values in dimensions.

Replace #3042


sbrunner:
Sometimes SQL mix `NULL` and empty string, do you test if it's effectively returns `None`?

arnaud-morvan:
I've tested that I get null on client side, everything goes fines now, except that I cannot pass the tests, and I cannot reproduce locally.

sbrunner:
@arnaud-morvan Thanks, I'm working on the nonworking Travis build :-)

sbrunner:
It's just fixed, you can rebase your branch :-)

arnaud-morvan:
Thanks @sbrunner

PR #3064, opened 2017-08-25
renaming models into commons - wip

fvanderbiest:


commons/c2cgeoportal_commons/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Pyramid application test package
```
cmangeat: not really a pyramid application, but commons ?

PR #3071, opened 2017-08-28
Fix the case of the copy to metadata

### LINKED ISSUE #3057: copy_to metadata doesn't exist/work
Work in 1.6. Doesn't work in 2.1 and upper

Here: it exists a copyTo metadata: https://github.com/camptocamp/c2cgeoportal/blob/2.1/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl#L149-L150
But in the cgxp code, we looks for a "copy_to" metadata: https://github.com/camptocamp/cgxp/blob/master/core/src/script/CGXP/plugins/Editing.js#L1056
(and also a string, not a list !)

So, the copy_to fonction doesn't work.

Be careful to don't make any conflict with: https://github.com/camptocamp/c2cgeoportal/issues/2594

sbrunner:
fix #3057

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -146,8 +146,7 @@ vars:
         - name: isExpanded
           type: boolean
         # Layer
-        - name: copyTo
-          type: list
+        - name: copy_to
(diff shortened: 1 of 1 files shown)


sbrunner:
Done :-)

ger-benjamin:
Thanks :-)

PR #3073, opened 2017-08-28
Update admin readme iwith markdown

arnaud-morvan:


arnaud-morvan:
This will be reviewed globally before mergin to master ...

Issue #3041, opened 2017-08-16
pdfreport can't work when credentials are required


ger-benjamin:
Since GMF 2.1

The parameters given to these methods: https://github.com/camptocamp/c2cgeoportal/blob/2.1/c2cgeoportal/views/pdfreport.py#L140-L141

Doesn't match with the signature of this method: https://github.com/camptocamp/c2cgeoportal/blob/2.1/c2cgeoportal/lib/filter_capabilities.py#L57 , respectively this method: https://github.com/camptocamp/c2cgeoportal/blob/2.1/c2cgeoportal/lib/filter_capabilities.py#L63

sbrunner:
=>
in init:
```
self.mapserver_settings = request.registry.settings.get("mapserverproxy", {})
if "default_ogc_server" in self.mapserver_settings:
    self.default_ogc_server = self.get_ogcserver_byname(
        self.mapserver_settings["default_ogc_server"]
)
```
and
```
if self.layername in get_private_layers(self.default_ogc_server) and \
    self.layername not in get_protected_layers(role_id, self.default_ogc_server):
```
:-)

ger-benjamin:
`self.layername in get_private_layers(self.default_ogc_server)`

What if our layer is not in the default_ogc_server ?

ger-benjamin:
Closed with #3047

ger-benjamin:
Reopen because names are compared to object...

PR #3078, opened 2017-08-30
Get_protected_layer - clear connection to the db instance

ger-benjamin:
For #3041 again

To fix pdf_report and (almost) any other use of "get_protected_layers".
Fix also possible errors with "get_writable_layers"

(already done in 2.2 with: #3031 )


sbrunner:
Don't take care on Travis issue :-)

PR #3092, opened 2017-09-08
Upgrade print

jwkaltz:
Update so that mapfish-print 3.10.2 will be used instead of 3.6.0 for new projects based on c2cgeoportal 2.1.
This will not touch existing projects, because I assume print version upgrade is not necessarily desired at the same time as a c2cgeoportal version update.

If the changes are OK, I will also create a PR for 2.2 and master, and for the corresponding demo projects.


jwkaltz:
Hmm something is wrong with this PR... it shouldn't have all these files. I will close and try again.

PR #3093, opened 2017-09-11
Redirect to mobile if the screen is no more than 1024px wide

### LINKED ISSUE #3033: Issues with using touch events as reason for mobile redirection
The[ recommended redirection to the mobile interface](https://github.com/camptocamp/c2cgeoportal/blob/2468bf4f60066c6ca6b9f6d245ca6f81252adb7a/doc/integrator/legacy_cgxp.rst.mako#L109-L110) only checks for the availability of touch events:

`if(('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {`

The use of tablets like the Microsoft Surface as desktop replacement gets more and more common. These are often used in conjunction with a docking station and full-sized screens. On such a device you would get automatically redirected to the mobile version despite having a more than sufficiently large screen resolution.

Instead of looking for touch events, we should check the screen resolution with [screen.availWidth and screen.availHeight](https://www.quirksmode.org/dom/w3c_cssom.html#screenview). I don't know what good values for this would be. Maybe a width of 1024px and above?

fredj:
See http://caniuse.com/#feat=matchmedia

fixes #3033

### DIFF
--- a/c2cgeoportal/scripts/import_ngeo_apps.py
+++ b/c2cgeoportal/scripts/import_ngeo_apps.py
@@ -199,12 +199,13 @@ def main():
                         """<head>
     % if "no_redirect" not in request.params:
             <script>
-                if (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
+                var small_screen = window.matchMedia ? window.matchMedia('(max-width: 1024px)') : false;
+                if (small_screen && ((ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch)) {
                     window.location = "${request.route_url('mobile', _query=dict(request.GET)) | n}";
                 }
             </script>
     % endif
-""",
+""",  # noqa: E501
                         data,
                     )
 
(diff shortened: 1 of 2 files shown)


c2cgeoportal/scripts/import_ngeo_apps.py
```diff
@@ -199,7 +201,8 @@ def main():
                         """<head>
     % if "no_redirect" not in request.params:
             <script>
-                if (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
+                var small_screen = window.matchMedia ? window.matchMedia('(max-width: 1024px)') : false;
```
sbrunner: This size is too high...

doc/integrator/legacy_cgxp.rst.mako
```diff
@@ -105,7 +105,9 @@ Do the following change in the ``<package>/templates/desktop.html`` file:
    - ${'%'} if not no_redirect and mobile_url is not None:
    + ${'%'} if "no_redirect" not in request.params:
      <script>
-         if(('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
+   -     if (('ontouchstart' in window) || window.DocumentTouch && document instanceof DocumentTouch) {
+   +     var small_screen = window.matchMedia ? window.matchMedia('(max-width: 1024px)') : false;
```
sbrunner: Same here

c2cgeoportal/scripts/import_ngeo_apps.py
```diff
@@ -28,6 +28,8 @@
 # either expressed or implied, of the FreeBSD Project.
 
 
+# flake8: noqa: E501
```
sbrunner: I don't agree with ignoring the line length in all the file...

PR #3094, opened 2017-09-12
In 2.2, update used mapfish-print from 3.6.0 to 3.10.3

jwkaltz:


jwkaltz:
Yes, IMO the newest version of GMF should normally use the newest release of MapFish-Print.
Also, the problem is, we cannot really make a clean manual upgrade of mapfish-print as GMF is now (because the CONST_Makefile explicitly depends on an old version of postgresql jar etc. for mapfish-print).

c2cgeoportal/scaffolds/create/print/print-apps/+package+/config.yaml.mako
```diff
@@ -15,18 +15,18 @@ templates:
...
                 dpiSuggestions: [254]
                 zoomLevels: !zoomLevels
                     scales: [100, 250, 500, 2500, 5000, 10000, 25000, 50000, 100000, 500000]
                 width: 555
                 height: 675
+            scalebar: !scalebar &scalebar
+                width: 150
+                height: 20
+                default:
+                     fontSize: 8
```
sbrunner: this change is not needed and I think that northArrow and scalebar should be behind :-)

c2cgeoportal/scaffolds/update/CONST_print_url
```diff
@@ -1 +1 @@
-https://repo1.maven.org/maven2/org/mapfish/print/print-servlet/3.6.0/print-servlet-3.6.0.war
+https://repo1.maven.org/maven2/org/mapfish/print/print-servlet/3.10.2/print-servlet-3.10.2.war
```
sbrunner: 10.3 :-)

c2cgeoportal/scaffolds/update/print/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM camptocamp/mapfish_print:3.6.0
+FROM camptocamp/mapfish_print:3.10.2
```
sbrunner: 10.3 :-)

c2cgeoportal/scaffolds/create/print/print-apps/+package+/config.yaml.mako
jwkaltz: Regarding scalebar: we changed the order here too, so that it is consistent with the order for createMap / createScalebar. I think it is easier to read the file that way.

Regarding northArrow: are you sure? We applied the same changes as for Vaud, https://github.com/camptocamp/vaud_geomapfish/commit/119438438a282f88b084ad608f22df6408873561 
It works there (as far as we know).
Should we change the order, and if so, to which order exactly?

sbrunner: I understand then to be clean for me it's better to also move the `northArrow`...

PR #3095, opened 2017-09-12
Upgrade print version in c2cgeoportal

jwkaltz:


jwkaltz:
I will do a new PR on master to update print after the docker branch has been merged, and after the new mapfish-print is released -> I will create an Issue so this does not get forgotten.

PR #3097, opened 2017-09-12
Update from print usage version 3.6.0 to 3.10.2

jwkaltz:


jwkaltz:
Actually, this is the same PR as https://github.com/camptocamp/c2cgeoportal/pull/3095.
I opened it by accident, because the branch used for 3095 is misnamed (it is actually based on master, not on 2.2)

PR #3115, opened 2017-09-25
Use pytest in place of nose

sbrunner:


tests/test_cachebuster.py
```diff
@@ -30,7 +30,11 @@
 from unittest import TestCase
 from pyramid.testing import DummyRequest
 from c2cgeoportal.pyramid_ import CACHE_PATH
-from c2cgeoportal.lib.cacheversion import CachebusterTween
+from c2cgeoportal.lib.caching import init_region
+
+init_region({"backend": "dogpile.cache.memory"})
```
pvalsecc: Are you sure you should not put that in the setup of the test?

PR #3132, opened 2017-09-28
Move checkers forward_headers key checker in CONST_config-schema.yaml

### LINKED ISSUE #3127: forward_headers is allowed in checker sub sections but readed from checker section
Project: aeroportdelyon_sig

* GitHub URL: https://github.com/camptocamp/aeroportdelyon_sig
* GeoMapFish version: 2.2

In section 'checker' in https://github.com/camptocamp/c2cgeoportal/blob/2.2/c2cgeoportal/scaffolds/update/CONST_config-schema.yaml
`forward_headers` is allowed in `checker` sub sections (for example in section `defaults`).

But in https://github.com/camptocamp/c2cgeoportal/blob/2.2/c2cgeoportal/views/checker.py#L65
it is readed from `checker` section.

So is no more possible to transfert Cookies through checkers.

arnaud-morvan:
To be consistent with views.checker.build_url.

Fix #3127

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_config-schema.yaml
+++ b/c2cgeoportal/scaffolds/update/CONST_config-schema.yaml
@@ -355,6 +355,10 @@ mapping:
                 type: map
                 required: True
                 mapping:
+                    forward_headers:
+                        type: seq
+                        sequence:
+                        -   type: str
                     regex;(.+):
                         type: map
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
I've changed the config validator, as this do not need a new realase of c2cgeoportal.

PR #3134, opened 2017-09-29
Add a tag to the docker image

sbrunner:


pvalsecc:
Shouldn't that be computed from the branch name instead of being hard coded in 4 locations?

sbrunner:
It can't be computed from the Git branch name because when I create a branch from the 2.2 branch is should steel be 2.2.
But I agree that the situation isn't so good, and for that I already create a task: 
https://jira.camptocamp.com/browse/GSGMF-108
For me we can leave like that on this branch :-)

PR #3137, opened 2017-10-02
Run Codacy outside Docker 

sbrunner:


sbrunner:
Actually it's not working and I suspect that ti needs something he can't get in Docker.
I found that difficult to test and debug ...

pvalsecc:
c2cwsigutils does it and it works...

PR #3141, opened 2017-10-04
Allows use of smtps

yjacolin:
More and more smtp service needs authentication to use a SMTP service (SSL,
auth, stattls). So we need to support authenticated smtp, smtps, etc.

### DIFF
--- a/c2cgeoportal/lib/email_.py
+++ b/c2cgeoportal/lib/email_.py
@@ -43,13 +43,13 @@ def send_email(from_addr, to_addrs, body, subject, smtp):  # pragma: no cover
     msg.attach(MIMEText(body, "plain", "utf-8"))
 
     # Connect to server
-    if smtp.get('ssl', False):
-        smtp = smtplib.SMTP_SSL(smtp['host'])
+    if smtp.get("ssl", False):
+        smtp = smtplib.SMTP_SSL(smtp["host"])
     else:
-        smtp = smtplib.SMTP(smtp['host'])
-    if smtp.get('user', False):
-        smtp.login(smtp['user'], smtp['password'])
-    if smtp.get('starttls', False):
+        smtp = smtplib.SMTP(smtp["host"])
+    if smtp.get("user", False):
+        smtp.login(smtp["user"], smtp["password"])
+    if smtp.get("starttls", False):
         smtp.starttls()
 
     smtp.sendmail(from_addr, to_addrs, msg.as_string())
--- a/c2cgeoportal/lib/email_.py
+++ b/c2cgeoportal/lib/email_.py
@@ -34,13 +34,23 @@ from email.mime.multipart import MIMEMultipart
 from email.Utils import formatdate
 
 
-def send_email(from_addr, to_addrs, body, subject, smtp_server):  # pragma: no cover
+def send_email(from_addr, to_addrs, body, subject, smtp):  # pragma: no cover
     msg = MIMEMultipart()
     msg["From"] = from_addr
     msg["To"] = ", ".join(to_addrs)
     msg["Date"] = formatdate(localtime=True)
     msg["Subject"] = subject
     msg.attach(MIMEText(body, "plain", "utf-8"))
-    smtp = smtplib.SMTP(smtp_server)
+
+    # Connect to server
+    if smtp.get('ssl', False):
+        smtp = smtplib.SMTP_SSL(smtp['host'])
+    else:
+        smtp = smtplib.SMTP(smtp['host'])
+    if smtp.get('user', False):
+        smtp.login(smtp['user'], smtp['password'])
+    if smtp.get('starttls', False):
+        smtp.starttls()
+
     smtp.sendmail(from_addr, to_addrs, msg.as_string())
     smtp.close()
--- a/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
+++ b/c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
@@ -110,7 +110,12 @@ vars:
 
             Sincerely yours
             The GeoMapfish team
-        smtp_server: smtp.example.com
+        smtp:
+          host: smtp.example.com
+          ssl: true
+          user: <username>
+          password: <password>
+          starttls: false
 
 update_paths:
 - authtkt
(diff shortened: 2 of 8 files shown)


c2cgeoportal/lib/email_.py
```diff
@@ -34,13 +34,23 @@
...
     msg = MIMEMultipart()
     msg["From"] = from_addr
     msg["To"] = ", ".join(to_addrs)
     msg["Date"] = formatdate(localtime=True)
     msg["Subject"] = subject
     msg.attach(MIMEText(body, "plain", "utf-8"))
-    smtp = smtplib.SMTP(smtp_server)
+
+    # Connect to server
+    if 'ssl' in smtp and smtp['ssl']:
```
arnaud-morvan: if smtp.get('ssl', False) would me better for me

c2cgeoportal/lib/email_.py
```diff
@@ -34,13 +34,23 @@
...
+        smtp = smtplib.SMTP_SSL(smtp['host'])
+    else:
+        smtp = smtplib.SMTP(smtp['host'])
+    if 'user' in smtp and smtp['user']:
```
arnaud-morvan: same, better to use dict.get with default value

doc/integrator/reset_password.rst
```diff
@@ -29,6 +29,11 @@ And to generate the required e-mail, in the ``vars_<package>.yaml`` file, add th
 
             Sincerely yours
             The GeoMapfish team
-        smtp_server: smtp.example.com
+        smtp:
+          host: smtp.example.com
+          ssl: true
+          user: <username>
+          password: <password>
+          starttls: false
```
arnaud-morvan: If the hostname ends with a colon (`:') followed by a number, and
        there is no port specified, that suffix will be stripped off and the
        number interpreted as the port number to use.

c2cgeoportal/lib/email_.py
yjacolin: default value should be '' or null here?

doc/integrator/reset_password.rst
yjacolin: you mean I need to document this?

arnaud-morvan:
Note that when this will be merged in GMF2, CONST_config-schema.yaml should be updated.

yjacolin:
Note to Arnaud's note ;) : https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/scaffolds/update/CONST_config-schema.yaml#L261

c2cgeoportal/lib/email_.py
sbrunner: @arnaud-morvan it's not the same, and it just should be: 
```
if 'user' in smtp:
```

sbrunner:
Is there any sense to be able to put different SMTP servers?
Shouldn't we configure it globally?

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -1,5 +1,19 @@
...
+==============
+
+1. In the `vars_{{package}}.yaml` file, change:
+
+    smtp_server: smtp.example.com
+
+to
+
+    smtp:
+      host: smtm.example.com
```
sbrunner: Use the diff syntax (+/-) a there is done in the rest of the file...

arnaud-morvan:
@sbrunner: is it not configured globally in vars ?

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
arnaud-morvan: Coould you replace **smtm**.example.com by **smtp**.example.com

c2cgeoportal/lib/email_.py
```diff
@@ -34,13 +34,23 @@
...
     msg = MIMEMultipart()
     msg["From"] = from_addr
     msg["To"] = ", ".join(to_addrs)
     msg["Date"] = formatdate(localtime=True)
     msg["Subject"] = subject
     msg.attach(MIMEText(body, "plain", "utf-8"))
-    smtp = smtplib.SMTP(smtp_server)
+
+    # Connect to server
+    if smtp.get('ssl', false):
```
fredj: Use `False` instead of `false` (also for the lines above)

fredj: And the value will be a string, so even if the value is `"False"` in the config file, the test will be true

sbrunner:
Not ad global var, but in the root of configuration, not for each services...

c2cgeoportal/lib/email_.py
sbrunner: In the config file it should be a boolean!

arnaud-morvan:
@sbrunner: can you take a look at the error with travis, this seems related with http://ppa.launchpad.net/stephane-brunner

sbrunner:
You should add 
```yaml
dist: precise
```
in the `.travis.yaml` file https://docs.travis-ci.com/user/reference/overview/ :-)

c2cgeoportal/lib/email_.py
yjacolin: @sbrunner, @fredj  do you mean I have to change something in the vars_xx.yaml file?

yjacolin:
@sbrunner
-  I agree but for GMF 1.6 it is not useful I guess as there is only one feature using it, am I right? But it makes sense for GMF 2.x

doc/integrator/reset_password.rst
```diff
@@ -29,6 +29,15 @@ And to generate the required e-mail, in the ``vars_<package>.yaml`` file, add th
 
             Sincerely yours
             The GeoMapfish team
-        smtp_server: smtp.example.com
+        smtp:
+          host: smtp.example.com:465
+          ssl: true
+          user: <username>
+          password: <password>
+          starttls: false
```
sbrunner: incoherent indentation

c2cgeoportal/scaffolds/create/vars_+package+.yaml_tmpl
```diff
@@ -110,7 +110,12 @@ vars:
 
             Sincerely yours
             The GeoMapfish team
-        smtp_server: smtp.example.com
+        smtp:
+          host: smtp.example.com
+          ssl: true
+          user: <username>
+          password: <password>
+          starttls: false
```
sbrunner: incoherent indentation

sbrunner:
Then you will change tow times in the migrations ...

yjacolin:
@sbrunner thanks for your feedback. I commited my changes to implement this.

PR #3144, opened 2017-10-04
Set travis dist to precise

arnaud-morvan:
Fix travis in branch 1.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+dist: precise
+
 python:
 - 2.6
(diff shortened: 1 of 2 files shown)


sbrunner:
You should probably add <1.0.0 here: https://github.com/camptocamp/c2cgeoportal/blob/1.6/dev-requirements.txt#L7

arnaud-morvan:
Thanks @sbrunner, I merge to fix travis on 1.6

sbrunner:
:-)

PR #3145, opened 2017-10-04
Fix basic auth group finder

arnaud-morvan:
Principals was not returned in case of basic auth.

Regarding BasicAuthAuthenticationPolicy documentation for check:
    A callback function passed a username, password and request, in that order as positional arguments. Expected to return None if the userid doesn't exist or a sequence of principal identifiers (possibly empty) if the user does exist.

So we should return the role name here, works for me in aeroportdelyon_sig.


sbrunner:
Your branch has 2453 missing commits... can you rebase it?

arnaud-morvan:
Bad branch, based on 1.6

arnaud-morvan:
@sbrunner : I pretty sure about this, but can you give a +1 ?

PR #3157, opened 2017-10-10
Fix master

sbrunner:


.travis.yml
```diff
@@ -202,36 +213,23 @@ jobs:
...
-      script: travis/doc.sh
-      skip_cleanup: true
-      on:
-        repo: camptocamp/c2cgeoportal
-        branch: master
-        condition: '"${TRAVIS_PULL_REQUEST}" = false'
+
+  - stage: Publish
+    script: []
+    if: NOT type = pull_request
```
fredj: `NOT type = pull_request AND branch = master` like at the other places?

sbrunner: We want to publish Docker on all the branches (as previously) :-)

PR #3166, opened 2017-10-12
[admin] instrument check list

cmangeat:
base test
had to reorder models
remove back ref from restriction area to role
add a primary key on role_restrictionarea

need https://github.com/camptocamp/c2cgeoform/pull/118

### DIFF
--- a/commons/c2cgeoportal_commons/models/main.py
+++ b/commons/c2cgeoportal_commons/models/main.py
@@ -183,6 +183,72 @@ theme_functionality = Table(
     schema=_schema
 )
 
+# association table role <> restriciton area
+class RestrictionAreaForRole(Base):
+    __tablename__ = 'role_restrictionarea'
+    __table_args__ = (
+        {"schema": _schema}
+    )
+
+    id = Column("id", Integer, primary_key=True)
+    role_id = Column("role_id", Integer, ForeignKey(_schema + ".role.id"))
+    restrictionarea_id = Column("restrictionarea_id", Integer, ForeignKey(_schema + ".restrictionarea.id"))
+
+# association table layer <> restriciton area
+layer_ra = Table(
+    "layer_restrictionarea", Base.metadata,
+    Column(
+        "layer_id", Integer,
+        ForeignKey(_schema + ".layer.id"), primary_key=True
+    ),
+    Column(
+        "restrictionarea_id", Integer,
+        ForeignKey(_schema + ".restrictionarea.id"), primary_key=True
+    ),
+    schema=_schema
+)
+
+class RestrictionArea(Base):
+    __tablename__ = "restrictionarea"
+    __table_args__ = {"schema": _schema}
+    __acl__ = [
+        (Allow, AUTHORIZED_ROLE, ALL_PERMISSIONS),
+    ]
+
+    id = Column(Integer, primary_key=True)
+    area = Column(Geometry("POLYGON", srid=_srid), info={'colanderalchemy': {
+        'typ': colander_ext.Geometry(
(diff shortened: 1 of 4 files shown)


sbrunner:
Why all those changes in the model?

cmangeat:
Changes have two purposes:

- reorder allow using table definitions  from another table definition, especially at colanderalchemy level, i.e. for widget association, it's for example easier to know Role when defining User -> such changes have no impact on data model (sql one).
- had to redefine "role_restrictionarea", adding one primary key instead of two in order to make checboxes widget work-> has an impact on data model, I plan to see whether it is possible to avoid it once test coverage will be better (one test more).

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -183,6 +183,72 @@ def __unicode__(self):
     schema=_schema
 )
 
+# association table role <> restriciton area
+class RAForRole(Base):
```
sbrunner: Avoid using abbreviation =>
RestrictionArea_Role or RestrictionAreaForRole but less readable...

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -183,6 +183,72 @@ def __unicode__(self):
...
+    __tablename__ = 'role_restrictionarea'
+    __table_args__ = (
+        {"schema": _schema}
+    )
+
+    id = Column("id", Integer, primary_key=True)
+    role_id = Column("role_id", Integer, ForeignKey(_schema + ".role.id"))
+    restrictionarea_id = Column("restrictionarea_id", Integer, ForeignKey(_schema + ".restrictionarea.id"))
```
sbrunner: Missing uncity chack on (role_id, restrictionarea_id)

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -183,6 +183,72 @@ def __unicode__(self):
...
+
+# association table layer <> restriciton area
+layer_ra = Table(
```
sbrunner: Why moving this, is there any change?

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -183,6 +183,72 @@ def __unicode__(self):
...
+        ForeignKey(_schema + ".layer.id"), primary_key=True
+    ),
+    Column(
+        "restrictionarea_id", Integer,
+        ForeignKey(_schema + ".restrictionarea.id"), primary_key=True
+    ),
+    schema=_schema
+)
+
+class RestrictionArea(Base):
```
sbrunner: Why moving this, is there any change?

sbrunner:
Just for info it wasn't tow primary key but a primary key on two fields...

sbrunner:
A migration script is required ...

commons/c2cgeoportal_commons/models/main.py
cmangeat: backref to role from restriction area was removed,
ref to RestricationArea added to Role:
https://github.com/camptocamp/c2cgeoportal/blob/a124ef3ff9752ab166f1bfc93313a15d7f519aa3/commons/c2cgeoportal_commons/models/main.py#L284

commons/c2cgeoportal_commons/models/main.py
cmangeat: RestrictionArea has to know layer_ra

https://github.com/camptocamp/c2cgeoportal/blob/a124ef3ff9752ab166f1bfc93313a15d7f519aa3/commons/c2cgeoportal_commons/models/main.py#L232

PR #3167, opened 2017-10-12
[admin] Use node modules jinja2

arnaud-morvan:


admin/requirements.txt
```diff
@@ -1 +1 @@
--e git://github.com/camptocamp/c2cgeoform.git#egg=c2cgeoform
+-e git://github.com/camptocamp/c2cgeoform.git@731d8af#egg=c2cgeoform
```
cmangeat: To monitor...

arnaud-morvan: Yes this should be updated when https://github.com/camptocamp/c2cgeoform/pull/113 is merged.

arnaud-morvan:
I will merge this, comments are still welcome and will be treated later.

PR #3168, opened 2017-10-13
[admin] [WIP] Add modwsgi

arnaud-morvan:


fredj:
@arnaud-morvan @sbrunner what's the status of this PR?

arnaud-morvan:
I've started this when setting up a demo for the admin interface, but no mod_wsgi python3 on demo server.
We now include admin in geoportal as a pyramid plugin.
Can be closed.

Issue #2904, opened 2017-04-19
CSRF protection is too restrictive


arnaud-morvan:
Actually the CSRF protection is : 
```
if not (
    request.path_info_peek() == "admin" and request.referer is None or
    is_valid_referer(request.referer, settings)
):
```
But this is too much restrictive as this apply to all pyramid protected views, including GET requests.

The ``admin`` view in not the only one that need to be accessible without referer.
For example for aeroportdelyon_sig, we have protected custom apps (need authentication), and we also need authentication for the standard desktop and mobile app.

For me this should be simply : 
```
if not (request.referer is None or is_valid_referer(request.referer, settings)):
```
We could be more restrictive for non GET requests and AJAX request.

arnaud-morvan:
@pvalsecc @sbrunner : How do you feel about this ?

sbrunner:
I think that we should add a white list of path where we should be less restrictive...

arnaud-morvan:
another option could be to add a view decorator where CSRF protection is needed.

sbrunner:
By default it should be everywhere, and is should be disable where and when it's needed...

arnaud-morvan:
Patrick have check the valid ones (forgot None) and you have added (admin routes and no referer). IMHO it is not necessary to have a white list, checking that referer is None or valid one should be enough.

arnaud-morvan:
@sbrunner : This have been invalidated by #2915
We cannot anymore use links pointing to GMF.
Can you explain the difference between request.referer is None and request.referer = "".

arnaud-morvan:
Note that in aeroportdelyon_sig, user is autenticated by a kerberos token, and viewer (desktop/mobile) is not accessible to unauthenticated users. So this result in an 403 for all links pointing to c2cgeoportal.

sbrunner:
For me it's fixed on 2.3

PR #3175, opened 2017-10-16
Allow any referer in GET requests

arnaud-morvan:
We want any www site to be able to point to desktop or mobile interfaces for example.

Relate with #2904

### DIFF
--- a/c2cgeoportal/pyramid_.py
+++ b/c2cgeoportal/pyramid_.py
@@ -280,6 +280,8 @@ def _match_url_start(reference, value):
 
 
 def is_valid_referer(request, settings):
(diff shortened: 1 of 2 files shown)


c2cgeoportal/pyramid_.py
```diff
@@ -280,6 +280,8 @@ def _match_url_start(ref, val):
 
 
 def is_valid_referer(request, settings):
+    if request.method == 'GET':
```
yjacolin: Should be double quote here

arnaud-morvan:
This is really old PR, closing. Open new PR if needed.

Issue #3178, opened 2017-10-17
Unicode warnings in script theme2fts


ger-benjamin:
There is some unicode warnings by executing the script theme2fts

Some themes/groupe/layers must be in unicode
Interfaces given in arguments must be in unicode$

More details on https://jira.camptocamp.com/projects/GEO/queues/custom/52/GEO-564

PR #3179, opened 2017-10-17
Fix unicode issues in theme2fts script

ger-benjamin:
Fix: #3178

(Must also be reported on 2.2, master)

### DIFF
--- a/c2cgeoportal/scripts/theme2fts.py
+++ b/c2cgeoportal/scripts/theme2fts.py
@@ -48,6 +48,7 @@ def main():
     parser.add_argument(
         "--interfaces",
         nargs='+',
+        type=unicode,
         help="the interfaces to export",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


c2cgeoportal/scripts/theme2fts.py
```diff
@@ -241,3 +242,7 @@ def _add_layer(self, layer, interface, role):
             self._add_fts(layer, interface, "add_layer", role)
 
         return fill
+
+    @staticmethod
+    def _unicode_str(string):
+        return unicode(string, 'utf-8') if isinstance(string, str) else string
```
sbrunner: gettext return sometime a unicode and sometime a sting?
we mostly use `string.decode('utf-8')` than `unicode(string, 'utf-8')`

ger-benjamin: yes, str OR unicode...
Ok for the function.

PR #3189, opened 2017-10-23
Run commons checks and tests on Travis

arnaud-morvan:


.travis.yml
```diff
@@ -28,6 +28,20 @@ install:
...
+    - sudo apt-get autoremove postgis*
+    - sudo apt-get autoremove postgresql*
+    - sudo apt-get install postgresql-9.3-postgis-2.1
+    before_script:
+    - psql -c "CREATE USER \"www-data\" WITH PASSWORD 'www-data';" -U postgres
+    - psql -c "CREATE DATABASE c2cgeoportal_tests WITH OWNER c2cgeoform;" -U postgres
+    - psql -d c2cgeoportal_tests -c "CREATE EXTENSION postgis;" -U postgres
+    script:
+    - cd commons
+    - make check test
```
sbrunner: This is a good first point but is should use the same test infra as for geoportal, is it OK for you to merge like it and I will uniformise this?

commons/c2cgeoportal_commons/__init__.py
```diff
@@ -1,6 +1,12 @@
 """c2cgeoportal_commons package."""
 
-from c2cgeoportal_commons.models import get_session_factory, get_engine, get_tm_session, generate_mappers, init_dbsessions
+from c2cgeoportal_commons.models import (  # noqa
```
sbrunner: Why noqa?

arnaud-morvan: Why do you add this here ? it is unused, so it do not pass checks with noqa

arnaud-morvan: I've supposed you add this imports here for use in c2cgeoportal, as I do not want to break c2cgeoportal, I've prefered to put noqa than to break something in c2cgeoportal.

.travis.yml
arnaud-morvan: Yes, I'd like to merge this ASAP.

arnaud-morvan: Note we are not yet ready for docker and apache.

commons/c2cgeoportal_commons/__init__.py
sbrunner: No, they should be removed :-)

commons/README.md
```diff
@@ -14,7 +14,7 @@ make install
 ```
 sudo -u postgres psql -c "CREATE USER \"www-data\" WITH PASSWORD 'www-data';"
 
-DATABASE=c2cgeoportal_tests
+DATABASE=c2cgeoportal_test
```
sbrunner: Just one test?

arnaud-morvan: To be consistent with admin part, is this really relevant ?

sbrunner: no :-)

commons/Makefile
```diff
@@ -24,7 +24,7 @@ flake8: ${BUILD_DIR}/requirements-dev.timestamp
 
 ${BUILD_DIR}/venv.timestamp:
 	#Create a Python virtual environment.
-	python3 -m venv ${VENV}
+	virtualenv -p python3 ${VENV}
```
sbrunner: Why this change?

arnaud-morvan:
Because the old version works locally but not on travis, I do not know exactly why.

PR #3191, opened 2017-10-24
Upgrade the print to version 3.10.4

### LINKED ISSUE #3140: Print problem when setting background layer transparency
Version 2.2.0

To reproduce : 
1°/ Go to https://map.cartoriviera.ch/s/OmSf
2°/ Try to print -> in the PDF, the backgroundlayer plan cadastral is not visible

sbrunner:
fix #3140


jwkaltz:
What about geoportal/c2cgeoportal_geoportal/scaffolds/create/print/Dockerfile ? Doesn't this need to be updated as well?

sbrunner:
Thaks, fixed :-)

PR #3198, opened 2017-10-25
Fix the ids setter in the report specs

sbrunner:


jwkaltz:
@sbrunner according to BS (in ticket GEO-622) this PR works if one more change is made; quote:

In line #148 it should be:
"srsName": "epsg:" + str(srs)

c2cgeoportal/views/pdfreport.py
```diff
@@ -172,12 +178,56 @@ def get_report(self):
...
+                    "srs": srs,
+                    "mapserv_url": mapserv_url,
+                    "vector_request_url": vector_request_url,
+                })
 
         return self._do_print(spec)
+
+    def walker(self, spec, name, value):
+        if isinstance(spec, dict):
+            for k, v in spec.items():
```
ger-benjamin: I like to read not compiled code, but okay...

PR #3201, opened 2017-10-25
Configure alembic in admin

arnaud-morvan:
Here is what I would like to obtain on the admin side.
Note that everything should works fine with this, if things have not been modified in c2cgeoportal_commons.


arnaud-morvan:
To demonstrate you what we can do with alembic configuration : 
* note the use of package resource in ini file : c2cgeoportal_common:/alembic/main/version
* note the multi model in one config file with shared default values
* note the bind with pyramid app config
To demonstrate you how we can use alembic environment template for pylons which perfectly match with cookie-cutter template that we use to create c2cgeoportal_commons/models (but there no includeme in models anymore).
Because I do not like the idea that we have to copy version in /opt to use them.
Because I think that a piece of code is clearer than a long discussion.

PR #3205, opened 2017-10-27
Run admin checks and tests on travis

arnaud-morvan:


admin/Makefile
```diff
@@ -33,7 +33,7 @@ serve: build
 
 ${BUILD_DIR}/venv.timestamp:
 	#Create a Python virtual environment.
-	python3 -m venv ${VENV}
+	virtualenv -p python3 ${VENV}
```
sbrunner: Why this change?

arnaud-morvan: Because the old version works locally but not on travis, I don't know why.

sbrunner: Strange I just do a test...

sbrunner: it should work ...
https://travis-ci.org/camptocamp/c2cgeoportal/jobs/293577574

PR #3151, opened 2017-10-06
[admin] delete user from grid

cmangeat:
linked to https://github.com/camptocamp/c2cgeoform/pull/117
https://github.com/camptocamp/c2cgeoform/pull/117/commits/17def80d7d822cb14ae5a89e1fc033b85622b08c

PR #3208, opened 2017-10-27
[admin] Add delete link in grid

arnaud-morvan:
Superseed #3151


sbrunner:
An other merge without CI pass ?

PR #3213, opened 2017-10-31
[doc, 2.1] add update infos from GSSZ2-22

jwkaltz:
Add infos to c2cgeoportal 2.1 docu, as obtained when updating gmf 1.6 to gmf 2.1 in SZ project (as preliminary step to upgrade to gmf 2.2.), handled in ticket GSSZ2-22.

### DIFF
--- a/doc/builddoc.rst
+++ b/doc/builddoc.rst
@@ -8,3 +8,21 @@ Build this doc
   make doc
 
 The HTML should now be available in the ``doc/_build/html`` directory.
+
+Contribute to the documentation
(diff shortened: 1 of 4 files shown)


marionb:
@jwkaltz 
This looks good. 
You should add a short not for people who are using Intranet detection. They will have to make some modifications in the __init__.py, additional to those explained in the CONST_CHANGELOG.md 
For a description of what needs to be changed see the jira issue https://jira.camptocamp.com/browse/GEO-393.

Issue #3215, opened 2017-10-31
[doc] gmf 2.2 correct build instructions and add important update step


jwkaltz:
- https://camptocamp.github.io/c2cgeoportal/2.2/builddoc.html#build-doc
- regarding upgrade to 2.2, we should tell about the need to update to 2.1 first.

PR #3216, opened 2017-10-31
[doc, 2.2] doc build instructions and add note for upgrade

jwkaltz:
For issue #3215


sbrunner:
Can you rebase it?

jwkaltz:
Actually, an update to 2.1 docu should not be pushed to 2.2, because the instructions are different, and some parts are not compatible.
But I think I rebased it successfully, now.

PR #3229, opened 2017-11-03
Add mypy checks

sbrunner:
Add first type


Makefile
```diff
@@ -166,6 +166,17 @@ flake8:
 		--copyright-min-file-size=1 \
 		--copyright-regexp="Copyright \(c\) ([0-9][0-9][0-9][0-9]-)?$(shell date +%Y), Camptocamp SA"
 
+.PHONY: mypy
+mypy:
+	MYPYPATH=/usr/local/lib/python3.6/site-packages:/c2cwsgiutils \
```
gberaudo: Should be 3.5.

sbrunner: Why?

sbrunner: I't use on Docker who are based on Python 3.6 ...

gberaudo: Ah OK, I missed you are only calling the mypy target inside docker.

PR #3231, opened 2017-11-06
Fix private WMTS layer

sbrunner:
fix  https://github.com/camptocamp/ngeo/issues/3028


rbovard:
I think it should fix https://github.com/camptocamp/ngeo/issues/3028

Issue #3163, opened 2017-10-11
[doc] describe multiple DB setup


jwkaltz:
AFAIK, we currently have no documentation about setting up a site with multiple DBs.
One customer that has this is BS, see https://jira.camptocamp.com/browse/GEO-542, but the configuration should be done differently in GMF2 than what they have been doing so far (in GMF 1).

One place to add this documentation could be https://camptocamp.github.io/c2cgeoportal/master/integrator/legacy_advanced.html (whilst renaming the page to "advances" and making subsections), unless somebody has a better idea.

jwkaltz:
What I know so far is here: https://jira.camptocamp.com/browse/GEO-542?focusedCommentId=38357&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-38357

jwkaltz:
For doc update, waiting on https://github.com/camptocamp/c2cgeoportal/issues/3164

jwkaltz:
At the same time, the navigation tree within "integrator" should be consolidated to better group "legacy" stuff, so that the new section can be placed in a good place. Currently, some tree elements are redundant.

PR #3225, opened 2017-11-03
[doc] describe how to set up and use multiple databases

jwkaltz:
for issue #3163

PR #3244, opened 2017-11-08
Doc multiple db

jwkaltz:
for issue #3163 
(same PR as previously, retrying for CI technical reasons)


jwkaltz:
@sbrunner FYI this is actually the same contents as we had in PR #3225 , which you already saw. But for PR #3225 , codacy never worked, so I started this new PR, with the identical contents.
OK for you?

PR #3250, opened 2017-11-09
[admin] hide temp_password and iseditable fields from details

cmangeat:


commons/c2cgeoportal_commons/models/static.py
```diff
@@ -93,14 +93,16 @@ class User(Base):
     _password = Column("password", Unicode, nullable=False, info={
         'colanderalchemy': {'widget': HiddenWidget()}
     })
-    temp_password = Column("temp_password", Unicode, nullable=True)
+    temp_password = Column("temp_password", Unicode, nullable=True,
+        info={'colanderalchemy': {'widget': HiddenWidget()}}
+    )
```
arnaud-morvan: Have you tried:
```
'colanderalchemy': { 'exclude' : True }
```

commons/c2cgeoportal_commons/models/static.py
```diff
@@ -93,14 +93,16 @@ class User(Base):
...
     email = Column(Unicode, nullable=False, info={
         'colanderalchemy': {
             'title': _('email')
         }
     })
     is_password_changed = Column(Boolean, default=False, info={
-        'colanderalchemy': {'widget': CheckboxWidget(readonly=True)}
+        'colanderalchemy': {'widget': CheckboxWidget(readonly=True, hidden=True)}
```
arnaud-morvan: same thing: exclude

sbrunner:
Thanks :-)

PR #3252, opened 2017-11-09
add description, functionnalities, restrictionareas column

cmangeat:
await test datas from anaother pr
used c2cgeoform version miss column name translation as probably
of no use, disable test targetting this functionnality

### DIFF
--- a/admin/c2cgeoportal_admin/views/roles.py
+++ b/admin/c2cgeoportal_admin/views/roles.py
@@ -2,7 +2,7 @@ from pyramid.view import view_defaults
 from pyramid.view import view_config
 from c2cgeoform.views.abstract_views import AbstractViews
 from c2cgeoportal_commons.models.main import Role
-from c2cgeoform.views.abstract_views import ListField as ListField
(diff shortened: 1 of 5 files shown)


admin/c2cgeoportal_admin/views/roles.py
```diff
@@ -2,14 +2,21 @@
 from pyramid.view import view_config
 from c2cgeoform.views.abstract_views import AbstractViews
 from c2cgeoportal_commons.models.main import Role
+from c2cgeoform.views.abstract_views import ListField as ListField
```
sbrunner: => from c2cgeoform.views.abstract_views import ListField

admin/c2cgeoportal_admin/views/users.py
```diff
@@ -3,13 +3,16 @@
 from c2cgeoform.views.abstract_views import AbstractViews
 from c2cgeoportal_commons.models.static import User
 from colanderalchemy import setup_schema
+from c2cgeoform.views.abstract_views import ListField as ListField
```
sbrunner: same

PR #3254, opened 2017-11-10
Fix mixup between smtp config and email content in password reset and…

ochriste:
… shortener functions.

### DIFF
--- a/c2cgeoportal/lib/email_.py
+++ b/c2cgeoportal/lib/email_.py
@@ -34,7 +34,7 @@ from email.mime.multipart import MIMEMultipart
 from email.Utils import formatdate
 
 
-def send_email(from_addr, to_addrs, body, subject, smtp):  # pragma: no cover
+def send_email(from_addr, to_addrs, body, subject, smtp_config):  # pragma: no cover
     msg = MIMEMultipart()
     msg["From"] = from_addr
     msg["To"] = ", ".join(to_addrs)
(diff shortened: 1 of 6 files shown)


ochriste:
cf https://jira.camptocamp.com/browse/GEO-631

c2cgeoportal/lib/email_.py
```diff
@@ -34,23 +34,23 @@
...
-    msg["Subject"] = subject
-    msg.attach(MIMEText(body, "plain", "utf-8"))
+    msg["Subject"] = email_config.get("subject")
+    msg.attach(MIMEText(email_config.get("body"), "plain", "utf-8"))
 
     # Connect to server
-    if smtp.get("ssl", False):
-        smtp = smtplib.SMTP_SSL(smtp["host"])
+    if smtp_config.get("ssl", False):
+        smtp = smtplib.SMTP_SSL(smtp_config.get("host"))
```
sbrunner: The host is optional?

c2cgeoportal/lib/email_.py
```diff
@@ -34,23 +34,23 @@
...
     else:
-        smtp = smtplib.SMTP(smtp["host"])
-    if smtp.get("user", False):
-        smtp.login(smtp["user"], smtp["password"])
-    if smtp.get("starttls", False):
+        smtp = smtplib.SMTP(smtp_config.get("host"))
```
sbrunner: same ?

c2cgeoportal/lib/email_.py
```diff
@@ -34,23 +34,23 @@
...
+    if smtp_config.get("user", False):
+        smtp.login(smtp_config.get("user"), smtp_config.get("password"))
```
sbrunner: same?

c2cgeoportal/lib/email_.py
```diff
@@ -34,23 +34,23 @@
...
+    if smtp_config.get("starttls", False):
         smtp.starttls()
 
-    smtp.sendmail(from_addr, to_addrs, msg.as_string())
+    smtp.sendmail(email_config.get("from_addr"), email_config.get("to_addrs"), msg.as_string())
```
sbrunner: same?

c2cgeoportal/lib/email_.py
```diff
@@ -34,23 +34,23 @@
 from email.Utils import formatdate
 
 
-def send_email(from_addr, to_addrs, body, subject, smtp):  # pragma: no cover
+def send_email(email_config, smtp_config):  # pragma: no cover
```
sbrunner: Why do you replace `from_addr, to_addrs, body, subject` by `email_config`?

ochriste: since we pass the smtp config as an object, I thought it would be more coherent to also pass the email config as an object, 
but I can revert that if you dont like it

c2cgeoportal/lib/email_.py
ochriste: it is not,
how would you like me to write that ?

PR #3257, opened 2017-11-13
Admin wms layers

cmangeat:


arnaud-morvan:
You may add titles to fields in WMSLayer model

admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -0,0 +1,56 @@
...
+        ListField('metadata_url'),
+        ListField('description'),
+        ListField('public'),
+        ListField('geo_table'),
+        ListField('exclude_properties'),
+        ListField('layer'),
+        ListField('style'),
+        ListField('time_mode'),
+        ListField('time_widget'),
+        ListField('ogc_server', renderer=lambda wmsLayer: wmsLayer.ogc_server.name),
```
arnaud-morvan: convention in python is to use _ instead of upper case letters in variable names : wms_layer

admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -0,0 +1,56 @@
...
+        ListField('restrictionareas', renderer=lambda wmsLayer:
+                  ", ".join([r.name or '' for r in wmsLayer.restrictionareas]))]
+    _id_field = 'id'
+    _model = LayerWMS
+    _base_schema = LayerWMS.__colanderalchemy__
```
arnaud-morvan: You will have to use GeoFormSchemaNode(LayerWMS) at some time.
No more need to setup_schema(None, LayerWMS) in that case.

admin/c2cgeoportal_admin/views/layers_wms.py
arnaud-morvan: note that layer_wms may be more consistent with the model.

PR #3258, opened 2017-11-14
Fix test

sbrunner:


geoportal/tests/functional/test_themes_private.py
```diff
@@ -30,14 +30,13 @@
 
 import transaction
 
-from unittest2 import TestCase
```
ger-benjamin: Are you sure about that ? 2->1

sbrunner: unittest2 is a backport of things available in python for python 2, than th's correct :-)

PR #3259, opened 2017-11-15
Some fix to be able to upgrade from 2.2

sbrunner:


geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
```diff
@@ -75,6 +75,8 @@ def main():
     if options.nondocker:
         print("- nondocker")
     print("- git_remote=" + options.git_remote)
+    if options.use_makefile:
+        print("- use_makefile")
     print("- makefile=" + options.makefile)
```
ger-benjamin: looks a little bit strange, can you verify ?

What's the difference between "makefile" and "use_makefile" ? (also for the argument parser below (line 112))

sbrunner: The makefile is the make file used by c2cupgrade to lunch the build tasks, use_makefile is true if we lunch c2cupgrade true the makefile, it's used only tho display the commends used to run the different steps.

ger-benjamin: Ok, thanks

sbrunner: any idea?

ger-benjamin: Not really :sweat_smile:

PR #3266, opened 2017-11-16
[admin] Add role column to user grid

fvanderbiest:
We also remove the redundant `test_view_index` test for User

### DIFF
--- a/commons/c2cgeoportal_commons/models/static.py
+++ b/commons/c2cgeoportal_commons/models/static.py
@@ -103,6 +103,7 @@ class User(Base):
                                  info={'colanderalchemy': {'exclude': True}})
     role_name = Column(String, info={
         'colanderalchemy': {
+            'title': _('role'),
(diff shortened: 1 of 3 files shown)


commons/c2cgeoportal_commons/models/static.py
```diff
@@ -103,6 +103,7 @@ class User(Base):
                                  info={'colanderalchemy': {'exclude': True}})
     role_name = Column(String, info={
         'colanderalchemy': {
+            'title': _('role'),
```
arnaud-morvan: Note that I put uppercase on first character in other titles

cmangeat: This is true for wms layer... But false for role and user. So, lower or uppercase ?

arnaud-morvan:
Stephane and me agree on using upper case on first char (easier to read).
You can merge this, role and user titles should be updated later.

PR #3272, opened 2017-11-17
Add layertree view

arnaud-morvan:
Note that I use dbsession identity map to load all treeitems and relations in 2 SQL requests.
Complete tree get loaded in 100ms.
For that I've added some "polymorphic_identity" on abstract classes.

For now we do not have a real context menu but a dropdown menu.

TODO : Test delete, but this was not part of this print in facts.

### DIFF
--- a/admin/c2cgeoportal_admin/templates/layertree.jinja2
+++ b/admin/c2cgeoportal_admin/templates/layertree.jinja2
@@ -1,0 +1,130 @@
+{% extends "layout.jinja2" %}
+
+{% block extra_links %}
+<link href="{{request.static_url('c2cgeoportal_admin:node_modules/jquery-treegrid/css/jquery.treegrid.css')}}" rel="stylesheet">
+{% endblock extra_links %}
+
+{% block extra_scripts %}
+<script type="text/javascript" src="{{request.static_url('c2cgeoportal_admin:node_modules/jquery-treegrid/js/jquery.treegrid.js')}}"></script>
+<script type="text/javascript" src="{{request.static_url('c2cgeoportal_admin:node_modules/jquery.cookie/jquery.cookie.js')}}"></script>
+{% endblock extra_scripts %}
+
+{% block content %}
+
+<div id="layertree">
+
+    <div id="layertree-actions" class="btn-group">
+        <button id="layertree-expand" type="button" class="btn btn-default">{{_("Expand all")}}</button>
+        <button id="layertree-collapse" type="button" class="btn btn-default">{{_("Collapse all")}}</button>
+    </div>
+
+    <table id="layertree-table" class="tree table table-hover">
+        {% macro render_node(node, parent_id=None) -%}
+        <tr class="{{ node.item_type }} treegrid-{{ node.id }}{{' treegrid-parent-{}'.format(parent_id) if parent_id else ''}}">
+            <td>
+                {% if node.item_type == 'theme' %}
+                <span class="glyphicon glyphicon-briefcase"></span>
+                {% elif node.item_type == 'group' %}
+                <span class="glyphicon glyphicon-folder-open"></span>
+                {% elif node.item_type == 'layerv1' %}
+                <span class="glyphicon glyphicon-picture" style="color:grey"></span>
+                {% elif node.item_type == 'l_wms' %}
+                <span class="glyphicon glyphicon-picture"></span>
+                {% elif node.item_type == 'l_wmts' %}
+                <span class="glyphicon glyphicon-th"></span>
+                {% endif %}
+                {{ node.name }}
+            </td>
+            <td>
+                <div class="btn-group">
+                    <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                      <span class="glyphicon glyphicon-option-horizontal"></span>
+                    </button>
+                    <ul class="dropdown-menu">
+                        <li class="action-edit">
+                            <a title="{{_("Edit")}}"
+                                href="{{request.route_url('c2cgeoform_action', table=tables[node.item_type], id=node.id)}}">
+                                <span class="glyphicon glyphicon-pencil"></span>
+                                {{_("Edit")}}
+                            </a>
+                        </li>
+                        {% if parent_id %}
+                        <li class="action-unlink">
+                            <a title="{{_("Unlink")}}"
+                                data-url="{{request.route_url('layertree_unlink', group_id=parent_id, item_id=node.id)}}"
+                                href="#"
+                                role="button">
+                                <span class="glyphicon glyphicon-log-out"></span>
+                                {{_("Unlink")}}
+                            </a>
+                        </li>
+                        {% endif %}
+                        <li class="action-delete">
+                            <a title="{{_("Delete")}}"
+                                data-url="{{request.route_url('layertree_delete', item_id=node.id)}}"
+                                href="#"
+                                role="button">
+                                <span class="glyphicon glyphicon-remove"></span>
+                                {{_("Delete")}}
+                            </a>
+                        </li>
+                    </ul>
+                </div>
+            </td>
+        </tr>
+        {% for child in node.children %}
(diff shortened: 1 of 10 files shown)


admin/c2cgeoportal_admin/templates/layertree.jinja2
```diff
@@ -0,0 +1,139 @@
...
+<script type="text/javascript" src="{{request.static_url('c2cgeoportal_admin:node_modules/jquery-treegrid/js/jquery.treegrid.js')}}"></script>
+<script type="text/javascript" src="{{request.static_url('c2cgeoportal_admin:node_modules/jquery.cookie/jquery.cookie.js')}}"></script>
+{% endblock extra_scripts %}
+
+{% block content %}
+
+<div id="layertree">
+
+    <div id="layertree-actions" class="btn-group">
+        <button id="layertree-expand" type="button" class="btn btn-default">Expand all</button>
```
cmangeat: the _(...)  is missing ?

admin/c2cgeoportal_admin/templates/layertree.jinja2
```diff
@@ -0,0 +1,139 @@
...
+        <button id="layertree-collapse" type="button" class="btn btn-default">Collapse all</button>
+    </div>
+
+    <table id="layertree-table" class="tree table table-hover">
+        {% macro render_node(node, parent_id=None) -%}
+        <tr class="{{ node.item_type }} treegrid-{{ node.id }}{{' treegrid-parent-{}'.format(parent_id) if parent_id else ''}}">
+            <td>
+                {% if node.item_type == 'theme' %}
```
cmangeat: Is it possible to use a dictionnary for switch case ?

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -277,7 +277,7 @@ def __init__(self, name: str="") -> None:
 event.listen(TreeItem, "after_delete", cache_invalidate_cb, propagate=True)
 
 
-# association table LayerGroup <> TreeItem
+# association table TreeGroup <> TreeItem
```
sbrunner: The previous commit was better because we shouldn't have a Layer in a Theme

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -326,6 +326,9 @@ class TreeGroup(TreeItem):
     __tablename__ = "treegroup"
     __table_args__ = {"schema": _schema}
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "treegroup"  # needed for _identity_class
+    }
```
sbrunner: We shouldn't need that because it's an "abstract" class...

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -431,6 +434,9 @@ class Layer(TreeItem):
     __tablename__ = "layer"
     __table_args__ = {"schema": _schema}
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "layer"  # needed for _identity_class
+    }
```
sbrunner: same ...

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -453,6 +459,9 @@ def __init__(self, name: str="", public: bool=True) -> None:
 
 class DimensionLayer(Layer):
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "dimensionlayer"  # needed for _identity_class
+    }
```
sbrunner: same ...

admin/c2cgeoportal_admin/templates/navigation.jinja2
```diff
@@ -0,0 +1,10 @@
+<ul class="nav nav-pills nav-stacked">
+  <li role="presentation" class="{{'active' if request.matched_route.name == 'layertree' else ''}}">
```
cmangeat: Rooooh, according to mockups, need two levels in menu bar !

admin/c2cgeoportal_admin/views/layertree.py
```diff
@@ -0,0 +1,73 @@
...
+    @view_config(route_name="layertree",
+                 renderer="../templates/layertree.jinja2")
+    def index(self):
+        # Preload treegroups with children relations in dbsession.identity_map.
+        # Note that session identity_map keep weak references so we need to keep strong references.
+        groups = self._dbsession.query(TreeGroup). \
+            options(subqueryload("children_relation")). \
+            all()
+
+        # Preload all treeitems
```
cmangeat: redundant coment ?

commons/c2cgeoportal_commons/models/main.py
arnaud-morvan: Yes but we can have a LayerGroup in a Theme.
It's not a question of better or not, it's a question of right or false.

commons/c2cgeoportal_commons/models/main.py
arnaud-morvan: We need that to have treeitem in _identity_class so we get layers in session.identity_map under treeitem key. It is not usefull for the database side, but for the session side. With this I get the tree in 100ms instead of 1s.

arnaud-morvan:
@sbrunner: If a class do not have an _identity_class it breaks the chain to get a correct identity key for instances.

arnaud-morvan:
And the _identity_class depends on polymorphic_identity

admin/c2cgeoportal_admin/templates/layertree.jinja2
arnaud-morvan: I've forgot this, that is not needed anymore with layertree/delete route

commons/c2cgeoportal_commons/models/main.py
sbrunner: Effectively your right :-)

sbrunner:
@arnaud-morvan can you call me tomorrow?

arnaud-morvan:
Identity class comes from here : https://github.com/zzzeek/sqlalchemy/blob/fd4289c5829d6498495ac59fe1dccb23b4975281/lib/sqlalchemy/orm/mapper.py#L1011:L1014

arnaud-morvan:
When I search a treeitem with primary key, if a WMS layer exists in identity map under the same key, the session do not send a request to the database but return the WMS layer directly, even in a group children relation for example.

admin/c2cgeoportal_admin/templates/layertree.jinja2
arnaud-morvan: I've used a dict in the view controller for the edit routes. Concerning the icons, I do not want to do the same. Maybe we do something shorter using less.

admin/c2cgeoportal_admin/templates/navigation.jinja2
arnaud-morvan: Yes, I've seen that, but let's do it in another pull request.

admin/c2cgeoportal_admin/views/layertree.py
arnaud-morvan: why redondant, the first says that I preload treegroup with children relations and explain why, the second says that I also preload all the treeitems.

commons/c2cgeoportal_commons/models/main.py
sbrunner:
```
"polymorphic_identity": "abstract"  # For performance issue in mapper (_identity_class)
```

admin/c2cgeoportal_admin/views/layertree.py
cmangeat: Something like:

groups, items, themes = preloadAndGetStrongRef()
----
deleteStrongRef(groups, items, themes)

could be an alternative to comments ?

easy with python as:

>>> def test():
...     return 1,2
... 
>>> x,y = test()
>>> x
1
>>> y
2
>>>

PR #3273, opened 2017-11-17
[admin] WMTS layers

fvanderbiest:


admin/acceptance_tests/layers_wmts_test.py
```diff
@@ -0,0 +1,65 @@
+# pylint: disable=no-self-use
+
+import pytest
+
+from . import check_grid_headers
+
+
+@pytest.fixture(scope='class')
+@pytest.mark.usefixtures("dbsession")
+def insertLayerWMTSTestData(dbsession):
```
cmangeat: question : can we change insertLayerWMTSTestData for insertTestData as we are in layers_wmts_test.py ?

admin/acceptance_tests/layers_wmts_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+    for i in range(0, 25):
+        name = 'layer_wmts_{}'.format(i)
+        layer = LayerWMTS(name=name)
+        layer.layer = name
+        layer.url = 'https://server{}.net/wmts'.format(i)
+        layer.restrictionareas = [restrictionareas[i % 5],
+                                  restrictionareas[(i + 2) % 5]]
+        dbsession.add(layer)
+    yield
```
cmangeat: It also possible to keep data as in
https://github.com/camptocamp/c2cgeoportal/pull/3272/files#diff-376f71eaa343effe987e81143ae636b1R68
For further 'expected' building so not to need to requery.

admin/c2cgeoportal_admin/routes.py
```diff
@@ -13,11 +13,12 @@ def includeme(config):
 
     config.add_route('home', '/')
 
-    from c2cgeoportal_commons.models.main import Role
+    from c2cgeoportal_commons.models.main import (Role, LayerWMS, LayerWMTS)
```
cmangeat: What are the parenthesis purpose ?

admin/acceptance_tests/layers_wmts_test.py
arnaud-morvan: layer_wmts_test_data would be better dor me, as this is a fixture.
we need layer_wmts_test_data as dbsession or test_app.
note the example in layertree where I use `layer_wmts_test_data['themes']` in the tests

admin/acceptance_tests/layers_wmts_test.py
arnaud-morvan: I understand that we should yield the instances : `yield test_data`

admin/c2cgeoportal_admin/routes.py
arnaud-morvan: parenthesis are needed only in multiline statements

PR #3279, opened 2017-11-22
Update the upgrade documentation, Some fix for the upgrade, Test and fix the l10n generation

sbrunner:


sbrunner:
I don't understand witch part of the pull request you speak about?

docker-run
```diff
@@ -68,9 +68,15 @@ def main():
...
         "--env=GIT_BRANCH={}".format(git_branch),
-        "--env=GIT_HASH={}".format(git_hash)
+        "--env=GIT_HASH={}".format(git_hash),
     ])
 
+    if "SSH_AUTH_SOCK" in os.environ:
+        docker_cmd.extend([
+            "--volume={ssh}:{ssh}".format(ssh=os.environ["SSH_AUTH_SOCK"]),
+            "--env=SSH_AUTH_SOCK",
+        ])
```
arnaud-morvan: I don't like this. I do not want my ssh private key to be passed to container.
Note that I do not use upgrade script anymore because it run git commands in my place.

sbrunner: Do we relay want to be able to start an upgrade on a branch who isn't up do date?
For me not...

I understand why you have many issue with GeoMapFish If you don't want to use it as it's build for for ideological reason ...

arnaud-morvan: The first time I've run the upgrade procedure it has remove all my custom scripts without warning (git clean -fd).

Is there some pending pull requests to merge before the upgrade ? Who does upgrade geomapfish ? Who know if we are up to date ? Who is git commands for ? The script or the dev ?

While this script run a git clean -fd in my source folder I won't use it.

If my private ssh key is transfered to the container, not sure I will use it anymore, for security reasons, not for ideology.

arnaud-morvan: Note that if ssh key is really needed to clone private repositories in client app projects, it would make sense to put it in client app project docker compose file : 

volumes:
    - $SSH_AUTH_SOCK:$SSH_AUTH_SOCK
  environment:
    - SSH_AUTH_SOCK

geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
```diff
@@ -276,14 +276,12 @@ def step0(self, step):
     def step1(self, step):
         check_call(["git", "reset", "--hard"])
         check_call(["git", "clean", "--force", "-d"])
```
arnaud-morvan: Is it possible to propose two choices here :  stash all (including untracked files) or cancel
Before running `git reset --hard` and `git clean --force -d`.

sbrunner: We already have a message for that on step0, isn't clear enough?

arnaud-morvan: I see that there is also a status check at end of step0, can we move it at start of step1.

sbrunner: I don't understand the goal of your purpose...

arnaud-morvan: To not execute that type of commands in user repo, and leave the user clear his working dir.

arnaud-morvan: > Here is the output of 'git status'. Please make sure to commit all your "
> "changes before going further. All uncommited changes will be lost.

Last time I've read this, I've stashed all my changes to tracked files, but I've lost all my untracked files.

Note that since I use git, I've never run `git clean -fd`. And when I upgrade some c2cgeoportal app, I manually add all modified files, because I sometime have to change things in const files. This is the developper job. Don't you do like this ? C2cgeoportal is only a framework, not the application itself, this is why we have a git repo in the customer app.

sbrunner: Why not, I think that we shouldn't do that in this pull request, I purpose to create a Jira card for that...

sbrunner: > This is the developper job.

The upgrade tool is made to be run by an integrator, not necessary a developer...

> C2cgeoportal is only a framework

Clearly not, it's an application builder... than more an application itself than a framework...

But it don't change the initial discussion, we need that to be able to do the final commit. if we don't do that (eventually manually) the changes (including new files) will be mixed up with the previous changes...

arnaud-morvan: I would prefer having them mixed than lost.

docker-run
sbrunner: We are not less secure than before, I don't need that in the docker compose run because I don't use it to run the upgrade process

Jenkinsfile
```diff
@@ -84,46 +84,47 @@ timeout(time: 2, unit: 'HOURS') {
...
                 sh 'rm -rf ${HOME}/workspace/testgeomapfish'
                 sh 'docker build --tag=camptocamp/testgeomapfish-external-db:latest docker/test-external-db'
                 sh 'travis/create-new-project.sh ${HOME}/workspace'
-                sh 'travis/run-on.sh ${HOME}/workspace/testgeomapfish/ ./docker-run travis/empty-make --makefile=travis.mk help'
-                sh 'travis/run-on.sh ${HOME}/workspace/testgeomapfish/ ./docker-run make --makefile=travis.mk build'
+                sh 'travis/run-on ${HOME}/workspace/testgeomapfish/ ./docker-compose-run make update-po'
+                sh 'travis/run-on ${HOME}/workspace/testgeomapfish/ git add geoportal/testgeomapfish_geoportal/locale/*/LC_MESSAGES/testgeomapfish_geoportal-*.po'
+                sh 'travis/run-on ${HOME}/workspace/testgeomapfish/ git commit -m "Upgrade the po files"'
+                sh 'travis/run-on ${HOME}/workspace/testgeomapfish/ ./docker-run travis/empty-make --makefile=travis.mk help'
+                sh 'travis/run-on ${HOME}/workspace/testgeomapfish/ ./docker-run make --makefile=travis.mk build'
```
arnaud-morvan: I do not understand anything here, sorry.
Why do you have to make a commit ?
What does the travis/empty-make do ? no comments or description inside the script.
Can you add some comments for the other devs.

Makefile
```diff
@@ -425,8 +425,8 @@ geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/+package+_geoportal/
 	test -s $@
 
 geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/+package+_geoportal/locale/en/LC_MESSAGES/+package+_geoportal-client.po:
-	mkdir --parent $(dir $@)
-	touch $@
+	$(PRERULE_CMD)
+	@echo "Nothing to be done for $@"
```
arnaud-morvan: I do not understand, seems that we need a litlle explaination on how this works

arnaud-morvan:
@sbrunner: could add some comments in your makefiles and some descriptions in your PR so we could understand your changes.

sbrunner:
> Sorry, but I understand only 10 percent of code here.

Where?

Jenkinsfile
sbrunner: Because after I test that the build didn't modify any files ant that we don't have any create files that's not in the git ignore, the po files modification should be accepted.

Makefile
sbrunner: This file don't exists because it's the source language, and he shouldn't be empty => error on msgmerge

sbrunner:
> could add some comments in your makefiles and some descriptions in your PR so we could understand your changes.

The best it the read the commit messages...

travis/docker-compose-build.yaml.mako
```diff
@@ -0,0 +1,54 @@
+---
```
arnaud-morvan: Couldn'e we use the same docker-compose as in scaffold.
I see external db is not in scaffold, is it impossible to add it in scaffold ?

.travis.yml
```diff
@@ -137,14 +137,14 @@ jobs:
       before_script:
         - docker build --tag=external-db docker/test-external-db
         - travis/create-new-project.sh
-        - travis/run-on.sh /tmp/travis/testgeomapfish/ ./docker-run travis/empty-make --makefile=travis.mk help
-        - travis/run-on.sh /tmp/travis/testgeomapfish/ ./docker-run make --makefile=travis.mk build
-        - travis/run-on.sh /tmp/travis/testgeomapfish/ docker-compose up -d
+        - travis/run-on /tmp/travis/testgeomapfish/ ./docker-run travis/empty-make --makefile=travis.mk help
```
arnaud-morvan: concerning travis/run-on, is not equal to cd ... && which could be easyier to read ?

sbrunner: Patch welcome :-)

travis/docker-compose-build.yaml.mako
arnaud-morvan: @sbrunner: note that it should be possible to use a docker-compose.override.yml file here to add only the external_db service (and do not duplicate the others services configurations).

arnaud-morvan: The idea was good, but not sure it possible because of the use of mako.

sbrunner: I will investigate it :-)

Issue #3287, opened 2017-11-27
Caching issue on editable layers (1.6)


ger-benjamin:
On GMF 1.6 (and perhaps upper)

Sometimes, it is impossible to edit layer because the response from the server contains any editable layers.
See: https://jira.camptocamp.com/projects/GEO/queues/custom/51/GEO-269

PR #3288, opened 2017-11-27
Fix caching issues with editable layers

ger-benjamin:
Fix: #3287

Tested (one week only) within https://jira.camptocamp.com/projects/GEO/queues/custom/51/GEO-269

### DIFF
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -657,7 +657,6 @@ class Entry(object):
             "success": True
         }
 
-    @cache_region.cache_on_arguments()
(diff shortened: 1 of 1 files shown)


sbrunner:
To you test id it don't introduce performance issue on theme reload?

ger-benjamin:
I've not seen any difference in term of performances.

But the tested project (, with the 'edit' role,) has only 4 themes with a reasonable quantity of group and layers inside theme.

ger-benjamin:
I've tried to add a "log.warn(888...888)" line in this function.
I got exactly the same behaviors with, and without this "cache_region.cache_on_arguments()".

So at the first connection, I got the warning.
Then never. Or I need to logout (or login again) or to do an apache2ctl graceful

Issue #3312, opened 2017-12-06
[admin] make docker-build fails to find c2cgeoform commit


fvanderbiest:
```
Step 8 : RUN pip install --disable-pip-version-check --no-cache-dir     -r /commons/requirements.txt     -r requirements.txt
 ---> Running in 9bf7fdda49aa
Obtaining c2cgeoform from git+git://github.com/camptocamp/c2cgeoform.git@c54a04d73d66ec94420a147e9987c49642768166#egg=c2cgeoform (from -r /commons/requirements.txt (line 1))
  Cloning git://github.com/camptocamp/c2cgeoform.git (to c54a04d73d66ec94420a147e9987c49642768166) to ./src/c2cgeoform
  Could not find a tag or branch 'c54a04d73d66ec94420a147e9987c49642768166', assuming commit.
Requirement already satisfied: c2cwsgiutils in /c2cwsgiutils (from -r requirements.txt (line 1))
```

PR #3313, opened 2017-12-06
using c2cgeoform master

fvanderbiest:
This patch addresses #3312


arnaud-morvan:
Not sure this is the good strategy as c2cgeoform version is not pinned.
I would prefer having a precise version here.
The problem we have now is that this commit do not exists anymore since the related PR has been merged and branch deleted.

fvanderbiest:
Indeed.
Using a newer hash.

arnaud-morvan:
Closed because c2cgeoform have been already upgraded

PR #3337, opened 2017-12-15
Admin duplicate

cmangeat:
linked with https://github.com/camptocamp/c2cgeoform/pull/146


admin/acceptance_tests/__init__.py
```diff
@@ -65,3 +65,9 @@ def check_checkboxes(self, form, name, expected):
             assert exp['label'] == list(label.stripped_strings)[0]
             assert exp['value'] == checkbox['value']
             assert exp['checked'] == field.checked
+
+    def getFirstFieldNamed(self, form, name):
```
arnaud-morvan: the standard in python is to use underscores in function names

admin/acceptance_tests/__init__.py
```diff
@@ -65,3 +65,9 @@ def check_checkboxes(self, form, name, expected):
...
+        return form.fields.get(name)[0]
+
+    def setFirstFieldNamed(self, form, name, value):
+        form.fields.get(name)[0].value__set(value)
```
arnaud-morvan: Note that I do not find these functions really useful

cmangeat: They ought to be stub for further improvement...
For example getFirstFieldNamed(self, form, name) could become checkFirstFieldNamed((self, form, name, expected) .
I had to replace a few "form[name] = value" by "form.fields.get(name)[0] = value" then by "form.fields.get(name)[0].value__set(value)", so I thought I could be a good thing to be able to do the replacement in one place.

admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -50,19 +50,25 @@ def index(self):
...
 
-    @view_config(route_name='c2cgeoform_action',
+    @view_config(route_name='c2cgeoform_item',
                  request_method='DELETE')
     def delete(self):
         return super().delete()
+
+    @view_config(route_name='c2cgeoform_item_action',
+                 request_method='GET',
+                 renderer='../templates/edit.jinja2')
```
arnaud-morvan: There is no filter on duplicate action, maybe we should add a match_param filter here ?

cmangeat: True, it works because at this time there is one and only one action

PR #3340, opened 2017-12-19
Improve the layer cache 

sbrunner:
Test case:
```
# unautenticated theme request
curl 'https://preprod.cartoriviera.ch/sbrunner/wsgi/themes?interface=mobile&version=2&background=background' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache'
# autenticated theme request
curl 'https://preprod.cartoriviera.ch/sbrunner/wsgi/themes?interface=mobile&version=2&background=background&role=15' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' With the related cookie
```
Before (server time):
* first request: 21-23 s.
* second request: 25-26 s.

After (server time):
* first request: 10 s.
* second request: 4.5 s.

### DIFF
--- a/c2cgeoportal/models.py
+++ b/c2cgeoportal/models.py
@@ -474,6 +474,9 @@ class TreeGroup(TreeItem):
     __tablename__ = "treegroup"
     __table_args__ = {"schema": _schema}
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "treegroup"  # needed for _identity_class (performance issue)
+    }
 
     id = Column(
         Integer, ForeignKey(_schema + ".treeitem.id"), primary_key=True
@@ -593,6 +596,9 @@ class Layer(TreeItem):
     __tablename__ = "layer"
     __table_args__ = {"schema": _schema}
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "layer"  # needed for _identity_class (performance issue)
+    }
 
     id = Column(
         Integer, ForeignKey(_schema + ".treeitem.id"), primary_key=True
@@ -608,6 +614,9 @@ class Layer(TreeItem):
 
 class DimensionLayer(Layer):
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "dimensionlayer"  # needed for _identity_class (performance issue)
+    }
 
 
 class LayerV1(Layer):  # Deprecated in v2
--- a/c2cgeoportal/views/entry.py
+++ b/c2cgeoportal/views/entry.py
@@ -47,6 +47,7 @@ from pyramid.httpexceptions import HTTPFound, HTTPBadRequest, HTTPForbidden, HTT
 from pyramid.security import remember, forget
 from pyramid.response import Response
 from sqlalchemy import func
+from sqlalchemy.orm import with_polymorphic
 from sqlalchemy.orm.exc import NoResultFound
 from owslib.wms import WebMapService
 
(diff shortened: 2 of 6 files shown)


c2cgeoportal/views/entry.py
```diff
@@ -900,6 +898,13 @@ def _wms_layers(self, role_id, ogc_server):
 
         return wms, list(wms.contents), wms_errors
 
+    @cache_region.cache_on_arguments()
+    def _load_tree_itmes(self, cache_version):
```
arnaud-morvan: _load_tree_items

c2cgeoportal/views/entry.py
```diff
@@ -909,6 +914,7 @@ def _themes(
         This function returns theme information for the role identified
         by ``role_id``.
         """
+        self._load_tree_itmes(get_cache_version())
```
arnaud-morvan: _load_tree_items

c2cgeoportal/views/entry.py
```diff
@@ -900,6 +898,13 @@ def _wms_layers(self, role_id, ogc_server):
 
         return wms, list(wms.contents), wms_errors
 
+    @cache_region.cache_on_arguments()
+    def _load_tree_itmes(self, cache_version):
+        self._themes_cache = DBSession.query(Theme).options(joinedload(Theme.functionalities)).all()
+        self._layerswms_cache = DBSession.query(LayerWMS).options(joinedload(LayerWMS.dimensions)).all()
+        self._layerswmts_cache = DBSession.query(LayerWMTS).options(joinedload(LayerWMTS.dimensions)).all()
+        self._treeitem_cache = DBSession.query(with_polymorphic(TreeItem, [LayerGroup, LayerV1])).all()
+
```
arnaud-morvan: Pretty sure you could load other things here like children_relation, ui_metadata, it should be possible to render themes in less that 1s without cache. You should take a look at sqlalchemy logs in debug mode. Concerning dimensions you should use a subqueryload (do not return all LayerWMS fields for each dimension value).

arnaud-morvan: You could have at least only one request per table

c2cgeoportal/views/entry.py
sbrunner: :-)

c2cgeoportal/views/entry.py
sbrunner: I can do some tests on `children_relation`, 
I will add `ui_metadata`

> it should be possible to render themes in less that 1s without cache

with WMS capabilities requests that take more than 8s. :-/

I will do a quick test on `subqueryload`.

Thanks for your feedback @arnaud-morvan :-)

arnaud-morvan: joinedload should be used when you have only one remote record (ogc_server for example) in case your relation is a collection (uselist = True), it should be better to use subqueryload

arnaud-morvan: this will populate all children relations in one database request: 
```
groups = self._dbsession.query(TreeGroup). \
            options(subqueryload('children_relation')). \
all()
```

c2cgeoportal/models.py
```diff
@@ -474,6 +474,9 @@ class TreeGroup(TreeItem):
     __tablename__ = "treegroup"
     __table_args__ = {"schema": _schema}
     __acl__ = [DENY_ALL]
+    __mapper_args__ = {
+        "polymorphic_identity": "treegroup"  # needed for _identity_class (preformence issue)
```
fredj: `performance` instead of `preformence` (same above)

c2cgeoportal/views/entry.py
```diff
@@ -753,7 +757,7 @@ def _get_ogc_servers(self, group, depth=1):
             return ogc_servers
 
         # recurse on children
-        if isinstance(group, LayerGroup) and group.children > 0:
+        if isinstance(group, LayerGroup) and len(group.children) > 0:
```
fredj: good catch!

PR #3346, opened 2017-12-20
Admin ordered children

arnaud-morvan:


commons/c2cgeoportal_commons/models/main.py
```diff
@@ -294,22 +296,32 @@ class LayergroupTreeitem(Base):
     ]
 
     # required by formalchemy
-    id = Column(Integer, primary_key=True)
-    description = Column(Unicode)
-    treegroup_id = Column(Integer, ForeignKey(_schema + '.treegroup.id'))
+    id = Column(Integer, primary_key=True, info={
+        'colanderalchemy': { 'widget': HiddenWidget() } }
+    )
+    description = Column(Unicode, info={ 'colanderalchemy': { 'exclude': True }})
```
sbrunner: Why excluded?

admin/package.json
```diff
@@ -13,6 +13,7 @@
     "jquery-steps": "1.1.0",
     "jquery-treegrid": "0.3.0",
     "jquery.cookie": "1.4.1",
+    "magicsuggest-alpine": "1.0.0",
```
sbrunner: Is it a good choice?
in ngeo we use corejs-typeahead who has about 12000 monthly download and this one has only 16 monthly download :-/ 
or do he has an other goal?

commons/c2cgeoportal_commons/models/main.py
arnaud-morvan: This description column in group / item relation seems to be an error.
Why do we want a description in a link between two nodes ?
How do you want to present this to the user ?
Note that the treeitem already have a description.
Am I missing something ?

admin/package.json
arnaud-morvan: Can be discussed, I do not know corejs-typeahead very well.

arnaud-morvan: Last time I have used typeahead, interface was moving in each release.

commons/c2cgeoportal_commons/models/main.py
sbrunner: I don't catch that's in the relation, a customer need this for special operations, then we can't remove it, but effectively it's clear that she is hidden in the new interface :-)

admin/package.json
sbrunner: Not personally but at camptocamp we have a good knowledge on this library.
The API didn't change the last 3 years...

arnaud-morvan: I would like to merge, for now, it is working with magicsuggest, this could be changed later.

PR #3355, opened 2018-01-03
Ignore .build and node_modules in flake8

arnaud-morvan:
In admin, this allow us to run ./docker-run make flake8 locally without removing our build files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,10 @@ tests:
 .PHONY: flake8
 flake8:
 	# E712 is not compatible with SQLAlchemy
-	find $(VALIDATE_PY_FOLDERS) -name \*.py | xargs flake8 \
+	find $(VALIDATE_PY_FOLDERS) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -202,7 +202,10 @@ tests:
 .PHONY: flake8
 flake8:
 	# E712 is not compatible with SQLAlchemy
-	find $(VALIDATE_PY_FOLDERS) -name \*.py | xargs flake8 \
+	find $(VALIDATE_PY_FOLDERS) \
+		-not \( -path "*/.build" -prune \) \
+		-not \( -path "*/node_modules" -prune \) \
```
sbrunner: from where does those folders comes from?

arnaud-morvan: From non docker build in the admin folder.

sbrunner: OK :-/

PR #3357, opened 2018-01-04
Use srid from settings in geometry schemas

arnaud-morvan:


arnaud-morvan:
@cmangeat : Test added, if you want to have a look before merge

PR #3359, opened 2018-01-05
Remove mypy from docker/build/requirements.txt

arnaud-morvan:
To avoid conflicts, as it is already installed in c2cwsgiutils image in a more recent version.

### DIFF
--- a/docker/build/requirements.txt
+++ b/docker/build/requirements.txt
@@ -6,7 +6,6 @@ bottle==0.12.13  # i18n
 c2c.cssmin==0.7.dev6  # CGXP build
 c2c.template==1.5.1  # geoportal
 c2cgeoform==2.0.dev20180103  # commons
-codacy-coverage==1.3.9  # Codacy send report
 codespell==1.11.0  # Lint
 ColanderAlchemy==0.3.3  # commons
 coverage==4.4.2  # Build coverage XML for Codacy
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
@sbrunner @pvalsecc : For me it do not make sense to have pinned version of python packages in c2cwsgiutils. This create conflicts in app using it. Python packages needed by c2cwsgiutils should be intalled by the application using it.

arnaud-morvan:
Note that as we use a c2cwsgiutils docker image, not the python package, this have break c2cgeoportal master silently.

pvalsecc:
> For me it does not make sense to have pinned version of python packages in c2cwsgiutils. This create conflicts in app using it. Python packages needed by c2cwsgiutils should be intalled by the application using it.

Not sure about that since c2cwsgiutils is using those libraries. Plus, this accelerates the build since the libs are already in the image.

sbrunner:
I will work on unpin the dependencies in the setup.py of c2cwsgiutils...

sbrunner:
see: https://github.com/camptocamp/c2cwsgiutils/pull/71

pvalsecc:
I'm OK with the un-pinning, but not with the removal of the two libs. This would require a major release.

arnaud-morvan:
Installing python packages in c2cwsgiutils images may accelerate the build of applications in some cases. But this may also cause conflicts or increase the image size with different package version in different layers.

arnaud-morvan:
I will clause this PR, as it seems that we should unpin versions in c2cwsgiutils.

pvalsecc:
I'm a mess this morning. I thought this PR was for c2cwsgiutils. So I don't car if you remove those two packages.

PR #3360, opened 2018-01-05
[admin] Only show groups in theme children selector

arnaud-morvan:


arnaud-morvan:
@cmangeat : can you have a look

PR #3363, opened 2018-01-08
Wait for db to be healthy before starting build container

arnaud-morvan:
Hope this will fix repetitive failed connections to the database with jenkins.
This take place in the composition, so it should not affect commands runned with `./docker-run`

### DIFF
--- a/docker-compose.yaml.mako
+++ b/docker-compose.yaml.mako
@@ -1,30 +1,42 @@
----
+version: '2.1'
 
-db:
-  image: camptocamp/geomapfish-test-db:latest
-  environment:
-    - POSTGRES_USER=www-data
-    - POSTGRES_PASSWORD=www-data
-    - POSTGRES_DB=geomapfish_tests
+volumes:
+  ${build_volume_name}:
 
-mapserver:
-  image: camptocamp/geomapfish-test-mapserver:latest
-  links:
-    - db
+services:
+  db:
+    image: camptocamp/geomapfish-test-db:latest
+    environment:
+      - POSTGRES_USER=www-data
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
another possibility could be to add a wait-for-db script in composition build entry point

arnaud-morvan:
@sbrunner : what do think about this solution to wait for the db, it requires docker-compose v2.1 and it seems that georchestra team will also use this solution.

sbrunner:
Looks good, but you probably should do the same in
https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-build.yaml.mako, and in https://github.com/camptocamp/c2cgeoportal/blob/master/travis/docker-compose-build.yaml.mako

arnaud-morvan:
@sbrunner: Regarding the jenkins logs, it seems that the admin package is not installed. I've searched in docker files and make files, but I do not find the c2cgeoportal_admin package should be installed in /build/venv (run of setup.py).
Not sure this is related with this PR, but I've rebased on master, relaunched the build, nothing changes.

arnaud-morvan:
Seems that docker-compose V3 do not support depends_on with condition anymore.
https://docs.docker.com/compose/compose-file/#depends_on

So this strategy does not seems sustainable.

arnaud-morvan:
Ok, so I propose to use the c2c corporate flexible entry point, move the actual python run in docker-entrypoint.d without call to command, and add the waitdb script in admin-build image.
Is the admin-build image used somewhere out of a compostiion ?

arnaud-morvan:
As it seems not sustainable, preference goes to #3364

PR #3364, opened 2018-01-08
Wait for db in compose build cmd

arnaud-morvan:
Check another technique than #3363

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,10 +77,13 @@ timeout(time: 2, unit: 'HOURS') {
                     geoportal/tests/functional/alembic.ini \
                     docker/test-mapserver/mapserver.map prepare-tests'''
                 sh 'docker-compose rm --stop --force'
-                sh './docker-compose-run sleep 10'
-                sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=main upgrade head'
-                sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=static upgrade head'
-                sh './docker-compose-run make tests'
+                try {
+                    sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=main upgrade head'
+                    sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=static upgrade head'
+                    sh './docker-compose-run make tests'
+                } finally {
+                    sh 'docker-compose down'
+                }
                 sh './docker-run travis/codacy.sh'
             }
             stage('Test Docker app') {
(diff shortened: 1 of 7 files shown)


docker/build/bin/wait-for-db
```diff
@@ -0,0 +1,8 @@
+RETRIES=5
```
sbrunner: Please add the header line ...

docker-compose.yaml.mako
```diff
@@ -22,9 +22,12 @@ build:
     - USER_ID
     - GROUP_ID
     - CI
+    - PGUSER=www-data
+    - PGPASSWORD=www-data
+    - PGDATABASE=geomapfish_tests
   stdin_open: true
   tty: true
-  command: ${'$'}{RUN}
+  command: ["wait-for-db", "${'$'}{RUN}"]
```
sbrunner: Why adding the quotes? For me it's more readable without the quotes ...

arnaud-morvan:
@sbrunner : I do not understand why you need to json encode the docker cmd:

https://github.com/camptocamp/c2cgeoportal/blob/579b1d1aaaf0eb3fa7222d82a75afbf1078071aa/docker-compose-run#L13
https://github.com/camptocamp/c2cgeoportal/blob/579b1d1aaaf0eb3fa7222d82a75afbf1078071aa/docker/build/run#L17

So it is impossible to do exec "$@" like in a classic docker entry points:
```
build_1      | /usr/bin/wait-for-db: line 10: exec: %5B%22true%22%5D: not found
```

arnaud-morvan:
Note that I do not understand why we need an entry point in python (docker/build/run), seems that everything would be possible and really simpler with a bash script.

And why you need to create a python file with subprocess.check_call to run your command.

I think we should use something like https://www.camptocamp.com/actualite/utiliser-scripts-dentrypoint-docker-flexible/

sbrunner:
> I do not understand why you need to JSON encode the docker cmd

To be sure that the argument are passed correctly, Personally I prefer writing python script because I don't have good experience with bash script and arguments forwarding ...

sbrunner:
To have the right umask and the right user ...

arnaud-morvan:
Ok, so I propose to use the c2c corporate flexible entry point, move the actual python run in docker-entrypoint.d without call to command, and add the waitdb script in admin-build image.
Is the admin-build image used somewhere out of a compostiion ?

docker-compose.yaml.mako
arnaud-morvan: Because this is the prefered form.

docker-compose-run
```diff
@@ -1,18 +1,7 @@
...
-
-try:
-    subprocess.check_call(["docker-compose", "up", "--exit-code-from=build"])
-except subprocess.CalledProcessError as e:
-    exit(e.returncode)
+docker-compose run \
+    -e USER_NAME=$USER \
+    -e USER_ID=`id -u $USER` \
+    -e GROUP_ID=`id -g $USER` \
+    build $@
```
sbrunner: Why converting that in bash?
are you sure the the id command works on Windows?
Are you sure the the $@ handle correctly the arguments?

arnaud-morvan: > Why converting that in bash?
because i found it really simpler.

> are you sure the the id command works on Windows?
yes, I missed this.

> Are you sure the the $@ handle correctly the arguments?
yes, I'm sure in bash, but bash won't work on windows.

Everything here should be in the makefile.

sbrunner: ??? We don't want  a secound makefile ...

geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-run
```diff
@@ -10,11 +10,15 @@ import urllib.parse
...
 
 try:
     subprocess.check_call([
-        "docker-compose", "--file", "docker-compose-build.yaml", "up", "--exit-code-from=build"
+        "docker-compose",
+        "--file",
+        "docker-compose-build.yaml",
+        "run",
+        "build",
+        urllib.parse.quote(json.dumps(sys.argv[1:]))
```
sbrunner: This can't just be `sys.argv[1:]`?

geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-run
```diff
@@ -10,11 +10,15 @@ import urllib.parse
...
 os.environ["USER_ID"] = str(os.getuid())
 os.environ["GROUP_ID"] = str(os.getgid())
-os.environ["RUN"] = urllib.parse.quote(json.dumps(sys.argv[1:]))
 
 try:
     subprocess.check_call([
-        "docker-compose", "--file", "docker-compose-build.yaml", "up", "--exit-code-from=build"
+        "docker-compose",
+        "--file",
+        "docker-compose-build.yaml",
```
sbrunner: ` "--file=docker-compose-build.yaml",`
is more readable ...

Jenkinsfile
```diff
@@ -81,6 +81,7 @@ timeout(time: 2, unit: 'HOURS') {
                 sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=main upgrade head'
                 sh './docker-compose-run alembic --config=geoportal/tests/functional/alembic.ini --name=static upgrade head'
                 sh './docker-compose-run make tests'
+                sh 'docker-compose down'
```
sbrunner: Why do you add this?

arnaud-morvan:
@sbrunner: wait_for_db get all green statuses !! do you want to have a last look on these changes before merge ?

arnaud-morvan:
Note that I plan to suppress json encoding later

PR #3368, opened 2018-01-10
Admin metadata simple type

arnaud-morvan:


admin/acceptance_tests/metadatas_test.py
```diff
@@ -0,0 +1,209 @@
...
+        resp = self._post_metadata(
+            test_app, url, base_mapping, name, value, 200
+        )
+        assert error_msg == \
+            resp.html.select_one(
+                '.item-{} .help-block'.format(self.__metadata_ui_type(test_app, name))
+            ).getText().strip()
+        return resp
+
+    def _base_metadata_params(self, metadatas_test_data):
```
cmangeat: no need for self ? pure static ?

admin/acceptance_tests/metadatas_test.py
```diff
@@ -0,0 +1,209 @@
...
+        resp = self.get(test_app, '{}/{}'.format(prefix, item.id))
+        self._check_metadatas(test_app,
+                              resp.html.select_one('.item-metadatas'),
+                              item.metadatas)
+        resp.form.submit('submit', status=302)
+
+    def test_layer_wms_metadatas(self, metadatas_test_data, test_app):
+        self._test_edit_treeitem('layers_wms', metadatas_test_data['layer_wms'], test_app)
+
+    def test_layer_wmts_metadatas(self, metadatas_test_data, test_app):
```
cmangeat: following item.metadatas are empty ?

admin/c2cgeoportal_admin/schemas/metadata.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _add_value_node(self, type_name, colander_type, **kw):
+        self.add_before(
+            'description',
+            colander.SchemaNode(
+                colander_type,
+                name=type_name,
+                missing=colander.null,
+                **kw))
+        self.available_types.append(type_name)
```
cmangeat: available_type could be built elsewhere, I dont konw, may be a dict{'float', colander.Float()...}), and directly be accessed from metadata_types(...).

admin/c2cgeoportal_admin/schemas/metadata.py
```diff
@@ -0,0 +1,102 @@
...
+        '''
+        import dateutil.parser
+        ui_type = self._ui_type(obj.name)
+        if ui_type == 'date':
+            value = dateutil.parser.parse(value).date()
+        if ui_type == 'datetime':
+            value = dateutil.parser.parse(value)
+        if ui_type == 'time':
+            value = dateutil.parser.parse(value).time()
+        '''
```
cmangeat: A dict could be used ? dateutil.parser.parse(value) isolated ? an utility class used ?

admin/c2cgeoportal_admin/schemas/metadata.py
```diff
@@ -0,0 +1,102 @@
...
+    def objectify(self, dict_, context=None):
+        # depending on the type get the value from the right widget
+        dict_['value'] = dict_[self._ui_type(dict_['name'])]
+        return super().objectify(dict_, context)
+
+    def dictify(self, obj):
+        dict_ = super().dictify(obj)
+        value = obj.value or colander.null
+        '''
+        import dateutil.parser
```
cmangeat: import in header

admin/c2cgeoportal_admin/views/layer_groups.py
```diff
@@ -5,13 +5,15 @@
...
 from c2cgeoportal_admin.schemas.treegroup import children_schema_node
+from c2cgeoportal_admin.schemas.metadata import metadatas_schema_node
 from c2cgeoportal_admin.views.treeitems import TreeItemViews
 
 _list_field = partial(ListField, LayerGroup)
 
 
 base_schema = GeoFormSchemaNode(LayerGroup)
 base_schema.add(children_schema_node())
+base_schema.add(metadatas_schema_node.clone())
```
cmangeat: maybe no need for a clone

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -1163,24 +1163,33 @@ def __unicode__(self) -> str:  # pragma: no cover
 
 class Metadata(Base):
     __tablename__ = 'metadata'
-    __table_args__ = {'schema': _schema}
+    __table_args__ = (
+        UniqueConstraint('item_id', 'name'),
```
cmangeat: Constraint do no seems to be taken into account ????

admin/acceptance_tests/metadatas_test.py
arnaud-morvan: Yes, for the moment, this only checks the sequence is rendered.
We should filter available names depending on treeitem types later.

admin/c2cgeoportal_admin/views/layer_groups.py
arnaud-morvan: I prefer to keep clone() in case we alter it later, it will not alter schema in other views.

commons/c2cgeoportal_commons/models/main.py
arnaud-morvan: Yes, we should add a alembic script

admin/c2cgeoportal_admin/schemas/metadata.py
arnaud-morvan: to be moved when we uncomment these lines

admin/c2cgeoportal_admin/schemas/metadata.py
arnaud-morvan: maybe, when we will uncomment these lines, not really useful for the moment.

PR #3373, opened 2018-01-12
[admin], edit group from tree has to target layer_groups

cmangeat:


admin/acceptance_tests/layertree_test.py
```diff
@@ -103,13 +103,14 @@ def test_edit_button(self, test_app, layertree_test_data):
 
         for table, item_id in (
             ('themes', layertree_test_data['themes'][0].id),
-            ('groups', layertree_test_data['groups'][0].id),
+            ('layer_groups', layertree_test_data['groups'][0].id),
             # ('layers_v1', layertree_test_data['layers_wms'][0].id),
```
arnaud-morvan: this could be activated now, if we have corresponding test data.

PR #3379, opened 2018-01-15
[admin] Add temporary settings file for dev purpose

arnaud-morvan:
We need a settings file for admin map settings and admin metadata settings, so I've created a first PR only creating the settings file.


sbrunner:
Looks good to be merged :-)

PR #3389, opened 2018-01-19
[admin] Do not propose group ancestors to avoid cycles

arnaud-morvan:


admin/c2cgeoportal_admin/schemas/treegroup.py
```diff
@@ -38,6 +58,7 @@ def children_validator(node, cstruct):
     for dict_ in cstruct:
         if not dict_['treeitem_id'] in [item.id for item, group in node.treeitems]:
             raise colander.Invalid(
+                node,
                 'Value {} does not exist in table {}'.
                 format(dict_['treeitem_id'], TreeItem.__tablename__))
```
cmangeat: For end-user, for developer also, it should help having a less error prone message, e.g., "either value does not exits, either it exist but referencing it will cause a cycle amongst group", so to avoid looking for the non-existence of an existing value.

admin/c2cgeoportal_admin/schemas/treegroup.py
```diff
@@ -17,17 +18,36 @@ def objectify(self, dict_, context=None):
...
 
+    # Do not propose ancestors to avoid cycles
+    item_id = kw['request'].matchdict['id']
+    if item_id != 'new':
+        search_ancestors = dbsession.query(LayergroupTreeitem.treegroup_id). \
+            filter(LayergroupTreeitem.treeitem_id == item_id). \
+            cte(name='ancestors', recursive=True)
+        search_alias = aliased(search_ancestors, name='search_ancestors')
+        relation_alias = aliased(LayergroupTreeitem, name='relation')
+        search_ancestors = search_alias.union_all(
```
cmangeat: Reusing the same variable for storing data of different types should be avoided ?

arnaud-morvan: Not sure that union_all change the type, it should return an altered clone of the inital "selectable" object.

PR #3401, opened 2018-01-25
Admin unicity constraints2

cmangeat:
relates to https://github.com/camptocamp/c2cgeoform/pull/156

### DIFF
--- a/admin/acceptance_tests/__init__.py
+++ b/admin/acceptance_tests/__init__.py
@@ -122,3 +122,12 @@ class AbstractViewsTests():
                 assert exp['value'] == option['value']
             if 'selected' in exp:
                 assert exp['selected'] == ('selected' in option.attrs)
+
+    @staticmethod
+    def check_one_submission_problem(expected_msg, resp):
+        assert 'There was a problem with your submission' == \
+            resp.html.select_one('div[class="error-msg-lbl"]').text
+        assert 'Errors have been highlighted below' == \
+            resp.html.select_one('div[class="error-msg-detail"]').text
+        assert expected_msg == \
+            resp.html.select_one("[class~'has-error']").select_one("[class~'help-block']").getText()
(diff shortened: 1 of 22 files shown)


cmangeat:
@julsbreakdown , @arnaud-morvan , it is green !

admin/acceptance_tests/__init__.py
```diff
@@ -122,3 +122,12 @@ def _check_select(self, select, expected):
...
                 assert exp['selected'] == ('selected' in option.attrs)
+
+    @staticmethod
+    def check_one_submission_problem(expected_msg, resp):
+        assert 'There was a problem with your submission' == \
+            resp.html.select_one('div[class="error-msg-lbl"]').text
+        assert 'Errors have been highlighted below' == \
+            resp.html.select_one('div[class="error-msg-detail"]').text
+        assert expected_msg == \
+            resp.html.select_one("[class~'has-error']").select_one("[class~'help-block']").getText()
```
arnaud-morvan: This is good, but for consistency, I would prefer keeping same prototype as others checks for the moment : not static and protected, ex: 
``def _check_submission_problem(elem, expected_msg):``
or
``def _check_submission_problem(resp, item, expected_msg):``

Note that other tests could be factorized, maybe later.

admin/c2cgeoportal_admin/views/functionalities.py
```diff
@@ -0,0 +1,28 @@
...
+        'id',
+        'label',
+        order_by='label',
+        edit_url=lambda request, value: request.route_url(
+            'c2cgeoform_item',
+            table='functionalities', id=value
+        )
+    ),
+    validator=manytomany_validator
+)
```
arnaud-morvan: should be placed in schemas folder

admin/c2cgeoportal_admin/views/roles.py
```diff
@@ -10,58 +10,35 @@
...
+        'name',
+        order_by='name',
+        edit_url=lambda request, value: request.route_url(
+            'c2cgeoform_item',
+            table='roles',
+            id=value
+        )
+    ),
+    validator=manytomany_validator
+)
```
arnaud-morvan: should be placed in schema folder

PR #3407, opened 2018-01-30
Tilegeneration uses --capabilities flag now

yjacolin:


fredj:
`--capabilities` is an alias to `--generate-wmts-capabilities` ? 
https://github.com/camptocamp/tilecloud-chain/blob/master/tilecloud_chain/controller.py#L43-L45

doc/integrator/tilegeneration.rst
```diff
@@ -147,7 +147,7 @@ The main thing to do is to:
 
   .. prompt:: bash
 
-     ./docker-run generate_controller --generate_wmts_capabilities
+     ./docker-run generate_controller --capabilities
```
sbrunner: Use long name: `--generate-wmts-capabilities`...

fredj:
fixed in #3539

PR #3409, opened 2018-01-30
update mfp version

jwkaltz:
update mfp version; needs an update to Makefile, too, because mfp starting 3.11 no longer contains a print-apps directory.

Note that mfp 3.11 and older require a Java 8 runtime on the system. A Debian 8 backport for Java 8 is available, and installed on c2c's test and demo servers.

### DIFF
--- a/c2cgeoportal/scaffolds/create/CONST_print_url
+++ b/c2cgeoportal/scaffolds/create/CONST_print_url
@@ -1,0 +1,1 @@
+https://repo1.maven.org/maven2/org/mapfish/print/print-servlet/3.10.7/print-servlet-3.10.7.war
--- a/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -1,6 +1,14 @@
 This file includes migration steps for each release of c2cgeoportal.
 
 
+Version 2.2.3
+=============
+
+1. To update MapFish Print, you should get the corresponding file:
+
+    cp CONST_create_template/CONST_print_url .
+
+
 Version 2.2.0
 =============
 
--- a/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -114,7 +114,7 @@ ifndef TEST_PACKAGES
 TEST_PACKAGES = common
 ifeq ($(DOCKER), FALSE)
 TEST_PACKAGES += main
-ifneq (, $(filter $(PRINT_VERSION), 2 3))
+ifeq ($(PRINT_VERSION), 3)
 TEST_PACKAGES += print
 endif
 ifeq ($(MAPSERVER), TRUE)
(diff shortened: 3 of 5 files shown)


jwkaltz:
@sbrunner you mean, we don't want the version 3.12 in gmf 2.2? If so, why not?

sbrunner:
We don't do some upgrade in stable branch and this one is a huge upgrade because it needs a new Java version, that needs a new Debian version!!!

jwkaltz:
Changed as discussed.
Note: mfp 3.12 works on Debian 8, so no update of Debian needed. Just need to activate the backport of Java 8 (this has already been done on our test and demo servers, and tested).

c2cgeoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -1151,7 +1139,9 @@ endif
 else
 	cp $(PRINT_BASE_DIR)/$(PRINT_BASE_WAR) $(PRINT_TMP)/$(PRINT_WAR)
 ifeq ($(PRINT_VERSION), 3)
-	zip -d $(PRINT_TMP)/$(PRINT_WAR) print-apps/
+        # Versions mfp up to 3.10 have a print-apps directory; later versions do not, 
+        # so ignore any errors resulting from zip
+	zip -d $(PRINT_TMP)/$(PRINT_WAR) print-apps/ | true
```
sbrunner: double pipe ...

c2cgeoportal/scaffolds/update/print/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM camptocamp/mapfish_print:3.10.6
+FROM camptocamp/mapfish_print:3.12.0
```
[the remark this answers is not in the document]
jwkaltz: OK. 
But another question: should I move this file from scaffolds/update to scaffolds/create as well
(like we are doing for the CONST_print_url file)?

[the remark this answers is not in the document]
sbrunner: This file isn't used in standard project then we still keep as it is...

c2cgeoportal/scaffolds/create/CONST_print_url
```diff
@@ -1 +1 @@
-https://repo1.maven.org/maven2/org/mapfish/print/print-servlet/3.10.6/print-servlet-3.10.6.war
+https://repo1.maven.org/maven2/org/mapfish/print/print-servlet/3.10.7/print-servlet-3.10.7.war
```
sbrunner: Shouldn't we add a changelog entry to say that we should get this file from the CONST_crate_template?

sbrunner:
Squash your commits before merge :-)

c2cgeoportal/scaffolds/create/CONST_print_url
jwkaltz: I now added one - does it make sense like that?

c2cgeoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -1,6 +1,15 @@
 This file includes migration steps for each release of c2cgeoportal.
 
 
+Version 2.2.3
+=============
+
+1. If you wish to update the version of mapfish-print being used, get the 
+   corresponding file:
```
sbrunner: To update Mapfish print you should get ...

sbrunner:
Thanks, don't forget to squash your commits before merge :-)

PR #3411, opened 2018-02-01
Admin add new layertree

julsbreakdown:
see https://github.com/camptocamp/c2cgeoform/pull/159

### DIFF
--- a/admin/acceptance_tests/layertree_test.py
+++ b/admin/acceptance_tests/layertree_test.py
@@ -71,6 +71,7 @@ def layertree_test_data(dbsession):
         'layers_v1': layers_v1,
         'layers_wms': layers_wms,
         'layers_wmts': layers_wmts,
+        'ogc_servers': [ogc_server]
     })
 
     dbsession.rollback()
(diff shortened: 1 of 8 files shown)


admin/acceptance_tests/layertree_test.py
```diff
@@ -151,15 +152,69 @@ def test_unlink(self, test_app, layertree_test_data, dbsession):
...
+        layer_wms = layertree_test_data['layers_wms'][0]
+        layer_wmts = layertree_test_data['layers_wmts'][0]
+        # no new on layers
+        for item_id in (layer_wmts.id, layer_wms.id):
+            assert 0 == len(
+                resp.html.select(
+                    'tr.treegrid-{} li.action-new a'.format('_{}_{}_{}'.
+                                                            format(theme.id, group.id, item_id))))
+        for item_id, path, new_item, title_name, required_fields in (
+            (
```
cmangeat: item_id should be named parent_id ?

admin/c2cgeoportal_admin/templates/layertree.jinja2
```diff
@@ -31,6 +31,34 @@
...
+                           <a title="{{_("New Layer Group")}}"
+                               href="{{request.route_url('c2cgeoform_item', table='layer_groups', id='new')}}?parent_id={{node.id}}"
+                               role="button">
+                               <span class="glyphicon glyphicon-plus"></span>
+                               {{_("New Layer Group")}}
+                           </a>
+                       </li>
+                       {% endif %}
+                       {% if parent_id and node.item_type == 'group'%}
+                       <li class="action-new">
```
cmangeat: I don't understand why to test for parent_id.

admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -84,7 +98,15 @@ def view(self):
...
-        return super().save()
+        response = super().save()
+        parent_id = self._appstruct['parent_id'] \
+            if hasattr(self, '_appstruct') and self._appstruct is not None \
+            and self._appstruct.get('parent_id') else None
+        if parent_id:
+            parent = self._request.dbsession.query(TreeGroup).get(parent_id)
+            rel = LayergroupTreeitem(parent, self._obj, 100)
+            self._request.dbsession.add(rel)
+        return response
```
cmangeat: as layerWmsView, WmtsView, treeGroupView are TreeItemViews, this customized save could be declared at TreeItemView level.

admin/c2cgeoportal_admin/views/layer_groups.py
```diff
@@ -16,6 +19,15 @@
...
 
+# Used for the creation of a new layer group from the layertree
+base_schema.add(colander.SchemaNode(
+    colander.Integer(),
+    name='parent_id',
+    missing=colander.drop,
+    validator=colander.deferred(partial(base_deferred_parent_id_validator, model=Theme)),
+    widget=deform.widget.HiddenWidget()
+))
+
```
cmangeat: Despite of model, this could be factorized amongst layerWmsView, WmtsView, and treeGroupView ? Moved to schema directory ?

admin/c2cgeoportal_admin/views/layer_groups.py
```diff
@@ -54,7 +66,15 @@ def view(self):
                  request_method='POST',
                  renderer='../templates/edit.jinja2')
     def save(self):
-        return super().save()
+        response = super().save()
+        parent_id = self._appstruct['parent_id'] \
+            if hasattr(self, '_appstruct') and self._appstruct is not None \
+            and self._appstruct.get('parent_id') else None
```
cmangeat: I am not a big fan of such indirection... 
We could have state 
has_to_be_registred_in_parent = hasattr(self, '_appstruct') and self._appstruct is not None and self._appstruct.get('parent_id')

Anyway, I am surprised that we have to look for it in appStruct and that it was not objectfied, probably because we are setting an int where model look for a list... Is it ?

admin/acceptance_tests/layertree_test.py
arnaud-morvan: +1 and new_item could be changed to route_table or something similar but more precise.

admin/acceptance_tests/layertree_test.py
```diff
@@ -151,15 +152,69 @@ def test_unlink(self, test_app, layertree_test_data, dbsession):
...
-@pytest.mark.usefixtures('selenium', 'selenium_app', 'layertree_test_data')
-class TestLayerTreeSelenium():
-
-    _prefix = '/layertree'
-
-    def test_unlink(self, dbsession, selenium, selenium_app, layertree_test_data):
+    @skip_if_ci
+    @pytest.mark.selenium
+    @pytest.mark.usefixtures('selenium', 'selenium_app')
+    def test_unlink_selenium(self, dbsession, selenium, selenium_app, layertree_test_data):
```
arnaud-morvan: Why this, note that self._prefix is used some lines above.

admin/c2cgeoportal_admin/views/layer_groups.py
```diff
@@ -1,12 +1,15 @@
...
 from functools import partial
 from pyramid.view import view_defaults
 from pyramid.view import view_config
 from c2cgeoform.schema import GeoFormSchemaNode
 from c2cgeoform.views.abstract_views import ListField
-from c2cgeoportal_commons.models.main import LayerGroup
 from c2cgeoportal_admin.schemas.treegroup import children_schema_node
 from c2cgeoportal_admin.schemas.metadata import metadatas_schema_node
 from c2cgeoportal_admin.views.treeitems import TreeItemViews
+from c2cgeoportal_admin.schemas.treegroup import base_deferred_parent_id_validator
```
arnaud-morvan: why do not import it on the same line as other imports from c2cgeoportal_admin.schemas.treegroup

admin/c2cgeoportal_admin/views/layer_groups.py
arnaud-morvan: using something like: 
```
def parent_id_node(model):
    return colander.SchemaNode(
        colander.Integer(),
        ...
        validator=colander.deferred(...., model=model)),
```

admin/c2cgeoportal_admin/views/layer_groups.py
arnaud-morvan: Note that this is strictly equivalent to:
```
parent_id = getattr(self, '_appstruct', {}).get('parent_id', None)
```
My question would be : Does this correctly handle validation errors.
If there is a validation error, something like ```self._obj and inspect(self._obj).persistent``` should be False.

Is it possible to add a test with a validation error, that the form correctly render with validation error message.

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -10,12 +12,14 @@
...
-from c2cgeoportal_commons.models.main import LayerWMTS, LayerWMS, TreeItem
 from c2cgeoportal_admin.schemas.dimensions import dimensions_schema_node
+from c2cgeoportal_commons.models.main import \
+    LayerWMTS, LayerWMS, TreeGroup, LayergroupTreeitem, LayerGroup, TreeItem
+from c2cgeoportal_admin.views.dimension_layers import DimensionLayerViews
 from c2cgeoportal_admin.schemas.metadata import metadatas_schema_node
 from c2cgeoportal_admin.schemas.interfaces import interfaces_schema_node
 from c2cgeoportal_admin.schemas.restrictionareas import restrictionareas_schema_node
-from c2cgeoportal_admin.views.dimension_layers import DimensionLayerViews
+from c2cgeoportal_admin.schemas.treegroup import base_deferred_parent_id_validator
```
arnaud-morvan: Why you do move imports that are in alphabetic order ?

admin/c2cgeoportal_admin/views/layer_groups.py
cmangeat: It does correctly handle validation error as if there is a validation error, cstruct is empty, that is the magic.

arnaud-morvan: OK so adding a comment could be a good idea

admin/c2cgeoportal_admin/views/layers_wmts.py
julsbreakdown: Probably when  I resolved the merge conflicts

admin/c2cgeoportal_admin/views/layer_groups.py
julsbreakdown: ok

admin/c2cgeoportal_admin/views/layer_groups.py
julsbreakdown: So here i updated with christophe's suggerstion and add a comment

admin/c2cgeoportal_admin/views/layer_groups.py
julsbreakdown: So i factorize this by creating a new function in the schema?

admin/c2cgeoportal_admin/views/layers_wms.py
julsbreakdown: Ok

admin/acceptance_tests/layertree_test.py
julsbreakdown: 👍

admin/c2cgeoportal_admin/views/layer_groups.py
cmangeat: In schemas directory ? c2cgeoportal/admin/c2cgeoportal_admin/schemas/ ?

arnaud-morvan: This is related to TreeItem subclasses so I would put it in admin/c2cgeoportal_admin/schemas/treeitem.py, seems that related validator shoud be in same file.

julsbreakdown: this is what i meant. I will do that when I get back to work on monday

PR #3412, opened 2018-02-02
admin, conversion wms <-> wmts

cmangeat:
admin, delete wms, insert wmts, update treeitem when wms -> wmts
admin, delete wmts, insert wms, update treeitem when wmts -> wms
set url et ogcserver id  using wms-defaults and wmts-defaults


cmangeat:
@arnaud-morvan @julsbreakdown , it is green, please have a look !

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -53,7 +56,15 @@ def grid(self):
...
                  renderer='../templates/edit.jinja2')
     def view(self):
-        return super().edit()
+        _view = super().edit()
+        if not self._is_new():
+            convert_url = self._request.route_url(
+                'layers_wms_from_wmts',
+                table='layers_wms',
+                wmts_layer_id=self._request.matchdict.get('id'),
+                action='from_wmts')
```
arnaud-morvan: why do we need an action here ? seems not useful.

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -53,7 +56,15 @@ def grid(self):
                  request_method='GET',
                  renderer='../templates/edit.jinja2')
     def view(self):
-        return super().edit()
+        _view = super().edit()
```
arnaud-morvan: this is usally named response, note that _ (protected) make do not really make sense inside a method.

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -74,32 +85,35 @@ def duplicate(self):
...
-
-        self._populate_widgets(form.schema)
-
-        rendered = form.render(dict_, request=self._request)
-        return {
-            'form': rendered,
-            'deform_dependencies': form.get_widget_resources(),
-            'actions': self._get_back_to_overview_actions()
-        }
+        _dbsession = self._request.dbsession
```
arnaud-morvan: _ (protected) do not really make sense inside a method.

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -74,32 +85,35 @@ def duplicate(self):
...
+            u = update(TreeItem.__table__)
+            u = u.where(TreeItem.__table__.c.id == pk)
+            u = u.values({'type': 'l_wmts'})
+            _dbsession.execute(u)
+            _dbsession.expunge(src)
+
+        _dbsession.flush()
+        mark_changed(_dbsession)
+
+        return HTTPFound(self._request.route_url('c2cgeoform_item', action='edit', id=pk))
```
arnaud-morvan: There is no confirmation before potential data loss !

arnaud-morvan: We do not keep existing "Create new WMTS from WMS" action

admin/c2cgeoportal_admin/templates/widgets/dimensions.pt
```diff
@@ -0,0 +1,5 @@
+<?python import json ?>
```
arnaud-morvan: this is not used

PR #3416, opened 2018-02-05
Admin layout

arnaud-morvan:


admin/c2cgeoportal_admin/templates/widgets/layer_fields.pt
```diff
@@ -0,0 +1,66 @@
...
+         tal:condition="getattr(child.schema, 'column', 1) == 1 and child.schema.name not in ('interfaces', 'restrictionareas', 'dimensions', 'metadatas')"
+         tal:replace="structure child.render_template(field.widget.item_template)">
+      </div>
+    </div>
+  </div>
+
+  <div class="col-md-4 col-sm-6">
+    <div tal:repeat="child field.children" tal:omit-tag="">
+      <div
+         tal:condition="getattr(child.schema, 'column', 1) == 2 and child.schema.name not in ('interfaces', 'restrictionareas', 'dimensions', 'metadatas')"
```
cmangeat: mixed strategies for layout, sometimes use column from model, sometimes schema name.
also, column may let think about sqlalchemy, not about chameleon layout.
later, in layer v1, 'group' is used to carry layout information.

admin/c2cgeoportal_admin/templates/widgets/theme_fields.pt
```diff
@@ -0,0 +1,61 @@
...
+  <div class="col-md-6">
+      <div tal:repeat="child field.children" tal:omit-tag="">
+        <div
+           tal:condition="child.schema.name in ('metadatas')"
+           tal:replace="structure child.render_template(field.widget.item_template)">
+        </div>
+      </div>
+  </div>
+
+</div>
```
julsbreakdown: No new line

PR #3417, opened 2018-02-06
fix debug mode

marionb:
match closure library path with module name in package json to fix debug mode


marionb:
This fixes an error that appears in the base.js when in debug mode 

 `ImportError: No module named ERROR when running the expression ''process.stdout.write(require("closure-util").getLibraryPath())'' `

and which keeps the application from work in this mode.

PR #3419, opened 2018-02-07
JSON encoding not needed anymore

sbrunner:


sbrunner:
Thanks for the review :-)

PR #3424, opened 2018-02-07
Forward the request tracking ID to the other service

jwkaltz:
This will allow to follow the logs belonging to a single request coming from
the user.

This is a copy-paste of master's commit 69b4cd372bf621473aab998c15d0bd56645a674d; cherry-pick not possible because of difference in paths.


sbrunner:
Who use this in 2.2?

PR #3430, opened 2018-02-09
Upgrade c2c.template

sbrunner:
Update the vars files
Better ngeo dependency
Configurable container dependencies


arnaud-morvan:
I do not understand why you've moved initialize_db to admin/acceptance_tests, this should be the prefered option to create a new database schema from scratch. It should be accessible in commons.

sbrunner:
Because it's used only for the tests, it shouldn't be used on a relay environment...

commons/acceptance_tests/initializedb.py
```diff
@@ -0,0 +1,112 @@
...
+    dbsession.merge(User(
+        username='admin',
+        email='admin@camptocamp.com',
+        role=role_admin
+    ))
+    dbsession.merge(User(
+        username='user',
+        email='user@camptocamp.com',
+        role=role_user
+    ))
```
arnaud-morvan: Is this a duplication with admin/acceptance_tests/initialize_db ?

arnaud-morvan:
This is not the standard way of doing. This is why I have a 500 lines script generated by alembic.

arnaud-morvan:
For me it make sens to have the init db script in same package as the model, instead of multiples copies in tests.

sbrunner:
Witch 500 lines?

arnaud-morvan:
when you generate your alembic script using `alembic revision --autogenerate`

sbrunner:
Then we should create a testing stuff package ...

sbrunner:
I it generate 500 lines script It shouldn't be used ...

arnaud-morvan:
The way I see things here is : 
- generate the database schema using initializedb to get a clean model exactly corresponding to your model.
- alter you model
- use alembic to generate a script to upgrade your database schema
This way, new database schemas exactly match with sqlalchemy model and generated script are clean.

PR #3432, opened 2018-02-12
Don't install ngeo in the admin interface

sbrunner:


geoportal/c2cgeoportal_geoportal/scripts/import_ngeo_apps.py
```diff
@@ -199,17 +134,12 @@ def main():
...
+                    data,
+                )
 
-# temporary disable ...
-#        if args.js:
-            # Full text search
-#            data = _sub(r"datasetTitle: 'Internal',", r"datasetTitle: '{{project}}',", data)
+        data = re.sub("app", "{{package}}", data)
+        data = re.sub(re.escape("{{package}}lication"), "application", data)
+        data = re.sub(re.escape("{{package}}le"), "apple", data)
```
ger-benjamin: Except that, it looks always weird ?

geoportal/c2cgeoportal_geoportal/scripts/npm_packages.py
```diff
@@ -0,0 +1,101 @@
...
+    parser.add_argument('--ngeo', action="store_true", help="Add ngeo package")
+    parser.add_argument('excludes', nargs='*', metavar='EXCLUDE', help="The npm package to exclude")
+    parser.add_argument('src', metavar='SRC', help="The ngeo source file")
+    parser.add_argument('dst', metavar='DST', help="The destination file")
+
+    args = parser.parse_args()
+
+    with open(args.src) as src:
+        data = src.read()
+        packages = loads(data).get("devDependencies", loads(data).get("dependencies", []))
```
arnaud-morvan: This seems strange to me, I would suppose a dict update here.

geoportal/c2cgeoportal_geoportal/scripts/import_ngeo_apps.py
sbrunner: In one case we have `apple`, that's replaced by `{{package}}le`, this is to be back to apple ...

geoportal/c2cgeoportal_geoportal/scripts/npm_packages.py
sbrunner: In fact we don't have booth, but to don't be strange I will correct it :-)

sbrunner: Just fixed :-), thanks

PR #3434, opened 2018-02-13
[admin] Use bootstrap-table in layertree

arnaud-morvan:
Use bootstrap-table from source because of issue fixed but not yet released: https://github.com/wenzhixin/bootstrap-table/issues/3294
Note that jquery-treegrid plugin is no more yet released.

TODO: update tests

### DIFF
--- a/admin/c2cgeoportal_admin/static/theme.css
+++ b/admin/c2cgeoportal_admin/static/theme.css
@@ -3,10 +3,10 @@
   padding-right: 8px;
 }
 
-#layertree-table {
-  margin-top: 5px;
-  margin-bottom: 5px;
-  width: auto;
+.bootstrap-table th.actions,
+.bootstrap-table td.actions
+{
+  width: 30px;
 }
 
 .icon-theme:before {
--- a/admin/c2cgeoportal_admin/templates/layertree.jinja2
+++ b/admin/c2cgeoportal_admin/templates/layertree.jinja2
@@ -1,18 +1,23 @@
 {% extends "layout.jinja2" %}
 
 {% block extra_links %}
+<link href="{{request.static_url('c2cgeoportal_admin:node_modules/bootstrap-table/dist/bootstrap-table.min.css')}}" rel="stylesheet">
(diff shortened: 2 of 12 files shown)


arnaud-morvan:
Here is the new layer tree : localhost_6543_layertree galaxy s5

admin/package.json
```diff
@@ -8,9 +8,9 @@
   "author": "",
   "dependencies": {
     "bootstrap": "3.3.7",
-    "bootstrap-table": "1.11.2",
+    "opencollective": "*",
```
sbrunner: I don't understand what's this fore..., pleas specify the version :-)

arnaud-morvan: Good catch, I will pin the version. We need this to install bootstrap-table from source, it displays a message about donation on stdout during npm install.

sbrunner: OK, thanks :-)

admin/c2cgeoportal_admin/static/theme.css
```diff
@@ -3,14 +3,14 @@
   padding-right: 8px;
 }
 
-#layertree-actions {
-  margin-top: 5px;
-  margin-bottom: 5px;
+
+.bootstrap-table {
+  /* color: #42325b; */
```
cmangeat: dead code, wip

admin/c2cgeoportal_admin/templates/layertree.jinja2
```diff
@@ -1,117 +1,176 @@
...
 
     <script type="text/javascript">
       $(function() {
-        $('.tree').treegrid({
-          'initialState': 'collapsed',
-          'saveState': true,
-          'saveStateName': 'layertree'
+        var $grid = $('#layertree-table');
+        $grid.bootstrapTable({
+          height: $(window).height() - $('nav').height() - 30,
```
cmangeat: 30 should be variabilized ?

arnaud-morvan:
The need was to have same visual aspect as other grids (not for functional needs but visual consistency).
Note that this bring base factorization for action sharing between layertree and grids/forms.

PR #3435, opened 2018-02-14
 Fix application path for Closure, Make the Docker compose configurable, Make it working behind a proxy

sbrunner:
Depends on #3433


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
```diff
@@ -4,45 +4,54 @@ version: '2'
 services:
   db:
     image: ${docker_base}-testdb:${docker_tag}
+% if 'environment' in docker_services.get('db', {}):
```
cmangeat: Does such a cooking implies that a full build has to be triggered every time we change db user password ?

geoportal/c2cgeoportal_geoportal/lib/__init__.py
```diff
@@ -396,3 +397,14 @@ def add(self, config, name, spec, **extra):
             extra["pregenerator"] = C2CPregenerator(subdomain=True, version=False)
         return super(MultiDomainStaticURLInfo, self) \
             .add(config, name, spec, **extra)
+
+
+_formatter = Formatter()
+
+
+def confd_env(val):
```
arnaud-morvan: I not sure to understand, is it possible to add a docstring, an example ? 
=> Mako filter that allow to use environnement variables in mako templates. 
${"{host}:{port}" | confd_env} => {{getenv "host"}}:{{getenv "port"}}

geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
```diff
@@ -4,45 +4,54 @@ version: '2'
...
-      VISIBLE_ENTRY_POINT: /
-      TINYOWS_URL: http://tinyows/
-      MAPSERVER_URL: http://mapserver/
-      PRINT_URL: http://print:8080/print/
+% for key, value in docker_services['geoportal']['environment'].items():
+      ${key}: ${value}
+% endfor
+% endif
+    ports:
+      - ${docker_services['geoportal']['port']}:80
```
arnaud-morvan: Note that it is possible to use a .env file with docker-compose, this is the docker compose way of using variables in docker-compose : https://docs.docker.com/compose/environment-variables/

docker/build/run
```diff
@@ -39,6 +39,10 @@ try:
 except subprocess.CalledProcessError:
     pass
 
+if 'DOCKER_HOST_' in os.environ:
+    with open('/etc/hosts', 'a') as f:
+        f.write('\n{} docker-host\n'.format(os.environ['DOCKER_HOST_']))
```
arnaud-morvan: Why do need this ? I do not found where it is used.

geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
sbrunner: No It's used only in the docker compose, but if we run `./docker-run make build`, a big part of the build is done...

geoportal/c2cgeoportal_geoportal/lib/__init__.py
sbrunner: ${ host | confd_env}:${ port | confd_env} => {{getenv "host"}}:{{getenv "port"}}
I will add a comment :-)

geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
sbrunner: Not sure thats releay useful because we want to be able to override only some variable..., I will open a separate pull request for that :-)

docker/build/run
sbrunner: Not directly but is can be useful to have it :-)

sbrunner:
I will split this pull request in smaller ones :-)

PR #3439, opened 2018-02-15
[admin], validators for json and url metadatas

cmangeat:


admin/acceptance_tests/metadatas_test.py
```diff
@@ -184,6 +184,58 @@ def test_valid_int_metadata(self, test_app, metadatas_test_data):
...
             302)
 
+    def test_invalid_url_metadata(self, test_app, metadatas_test_data):
+        self._post_invalid_metadata(
+            test_app,
+            '/layers_wms/new',
+            self._base_metadata_params(metadatas_test_data),
+            '_url',
+            'gnagnagna',
+            'Must be a URL')
```
arnaud-morvan: an URL

admin/acceptance_tests/metadatas_test.py
```diff
@@ -184,6 +184,58 @@ def test_valid_int_metadata(self, test_app, metadatas_test_data):
...
+            '''{"colors": [{
+                "color": "black",
+                "category": "hue",
+                "type": "primary",
+                "code": {
+                    "rgba": [255,255,255,1,
+                    "hex": "#000"
+                }
+            }]}''',
+            'a parser would say "Expecting \',\' delimiter: line 7 column 26 (char 213)"')
```
arnaud-morvan: Not sure we need a translated introduction, but for the case the parser "says".
Another possibility for introduction: "Parser report: ..."

admin/acceptance_tests/metadatas_test.py
cmangeat: So says colander.url validator ! Do we have to translate it ?

arnaud-morvan: Seems already done in colander : https://github.com/Pylons/colander/blob/master/colander/locale/fr/LC_MESSAGES/colander.po#L59
If not done we should add colander translations in our translation fallback system in related story.

PR #3440, opened 2018-02-15
[admin], check for map widget in role

cmangeat:


admin/acceptance_tests/role_test.py
```diff
@@ -223,3 +226,33 @@ def test_grid_dberror(self, dbsession):
...
+        index_page.check_pagination_info('Showing 1 to 23 of 23 rows', 10)
+        index_page.select_page_size(10)
+        index_page.check_pagination_info('Showing 1 to 10 of 23 rows', 10)
+
+        # delete
+        role = roles_test_data['roles'][3]
+        index_page.click_delete(role.id)
+        index_page.check_pagination_info('Showing 1 to 10 of 22 rows', 10)
+        with pytest.raises(NoSuchElementException):
+            selenium.find_element_by_xpath("//a[@data-url='{}/roles/{}')]"
```
arnaud-morvan: now that delete button is in a dropdown, it should not be accessible before click on dropdown, so this is always true.

cmangeat: I do let a comment in place, could be linked with "One click edit from grid (layer tree ?)" issue.
Initially, the purpose was to ensure that the data / the page are reloaded.

PR #3441, opened 2018-02-15
[admin], make landing on layertree

cmangeat:


admin/c2cgeoportal_admin/views/home.py
```diff
@@ -1,6 +1,7 @@
+from pyramid.httpexceptions import HTTPFound
 from pyramid.view import view_config
 
 
-@view_config(route_name='home', renderer='../templates/home.jinja2')
```
arnaud-morvan: Seems that we could remove ../templates/home.jinja2
Won't it be simpler to set the layertree route = '/' intead of a redirection.

cmangeat: can't add twice the same route with different path, maybe it is possible to declare a path using regex witch match both '/' and 'layertree/', but I dindn't manage to do it and I am not sure it i is simplier. The avantage of redirection is that browser url bar carry the right url, i.e., '..../layertree', and so does referer.location.

PR #3447, opened 2018-02-16
[admin], use and test dblclick on row from c2cgeoform

cmangeat:
depends on https://github.com/camptocamp/c2cgeoform/pull/162

### DIFF
--- a/admin/Makefile
+++ b/admin/Makefile
@@ -20,7 +20,7 @@ cleanall:
 .PHONY: test
 test: build ${BUILD_DIR}/requirements-dev.timestamp
 	#Run your project's tests.
-	${VENV}/bin/pytest --driver Firefox
+	${VENV}/bin/pytest --driver Chrome
 
 .PHONY: check
 check: flake8
--- a/admin/acceptance_tests/layers_wms_test.py
+++ b/admin/acceptance_tests/layers_wms_test.py
@@ -1,5 +1,8 @@
 # pylint: disable=no-self-use
 
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.common.by import By
 import re
 import pytest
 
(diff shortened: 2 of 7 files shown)


admin/Makefile
```diff
@@ -20,7 +20,7 @@ cleanall:
 .PHONY: test
 test: build ${BUILD_DIR}/requirements-dev.timestamp
 	#Run your project's tests.
-	${VENV}/bin/pytest --driver Firefox
+	${VENV}/bin/pytest --driver Chrome
```
arnaud-morvan: could you add a line about installing the new driver in readme

admin/acceptance_tests/layers_wms_test.py
```diff
@@ -373,18 +376,33 @@ def test_submit_new_no_layer_name(self, test_app):
                     for x in resp.html.select("[class~'has-error']")])
 
 
+# following test implies use of chrome selenium web driver instead of gecko (Firefox)
```
arnaud-morvan: could you put this line exactly where the problem is

admin/acceptance_tests/layers_wms_test.py
```diff
@@ -380,11 +383,25 @@ class TestLayerWMSSelenium():
...
+                'return (window.jQuery != undefined && jQuery.active == 0)'))
+
+        el = index_page.find_element(
+            By.XPATH,
+            '//td[contains(text(),"{}")]'.format(layer.geo_table),
+            timeout=5)
+        action_chains = ActionChains(selenium)
+        # following event implies use of chrome selenium web driver instead of gecko (Firefox)
+        action_chains.move_to_element(el).double_click().perform()
+        assert selenium.current_url.endswith('/layers_wms/{}'.format(layer.id))
```
arnaud-morvan: Sorry to be late, but is it possible to factorize this in IndexPage please.

cmangeat: You are not late.

PR #3448, opened 2018-02-19
Remove intranet user detection instructions

jwkaltz:
(Same goal as https://github.com/camptocamp/c2cgeoportal/pull/3444, now applied separately to master because doc structures are not the same)

Remove intranet user detection instructions, because this is not a standard feature. Instead, this customization possibility is now described in Confluence (https://confluence.camptocamp.com/confluence/display/GeoSupport/GMF+Intranet+Users)


jwkaltz:
OK thanks

PR #3454, opened 2018-02-20
[requires.io] dependency update on master branch

sdebayle:


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_config-schema.yaml
```diff
@@ -11,6 +11,17 @@ mapping:
     required: True
     sequence:
       - type: str
+  interpreted:
+    type: map
+    required: True
+    mapping:
+      regex;.*:
```
fredj: `:` instead of `;` ?

sbrunner: no it's effectively a `;` as the other regex :-)

PR #3466, opened 2018-02-22
wms and wmts layers copy fields from default when new, discarding lay…

cmangeat:
…er and name

relate to https://github.com/camptocamp/c2cgeoform/pull/164

### DIFF
--- a/admin/acceptance_tests/layers_wms_test.py
+++ b/admin/acceptance_tests/layers_wms_test.py
@@ -100,6 +100,32 @@ class TestLayerWMSViews(AbstractViewsTests):
         # check search on interfaces
         self.check_search(test_app, 'mobile', total=9)
 
+    def test_new_no_default(self, test_app, layer_wms_test_data, dbsession):
+        default_wms = layer_wms_test_data['default']['wms']
+        default_wms.name = 'so_can_I_not_be found'
+        dbsession.flush()
+
+        form = self.get_item(test_app, 'new').form
+
+        assert '' == self.get_first_field_named(form, 'id').value
+        assert '' == self.get_first_field_named(form, 'name').value
+        assert '' == self.get_first_field_named(form, 'layer').value
+        assert '' == self.get_first_field_named(form, 'ogc_server_id').value
+        assert 'disabled' == self.get_first_field_named(form, 'time_mode').value
+        assert 'slider' == self.get_first_field_named(form, 'time_widget').value
+
+    def test_new_default(self, test_app, layer_wms_test_data):
+        default_wms = layer_wms_test_data['default']['wms']
+
+        form = self.get_item(test_app, 'new').form
+
+        assert '' == self.get_first_field_named(form, 'id').value
+        assert '' == self.get_first_field_named(form, 'name').value
+        assert '' == self.get_first_field_named(form, 'layer').value
+        assert str(default_wms.ogc_server.id) == self.get_first_field_named(form, 'ogc_server_id').value
+        assert default_wms.time_mode == self.get_first_field_named(form, 'time_mode').value
+        assert default_wms.time_widget == self.get_first_field_named(form, 'time_widget').value
+
     def test_base_edit(self, test_app, layer_wms_test_data):
         layer = layer_wms_test_data['layers'][10]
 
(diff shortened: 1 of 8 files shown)


admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -78,6 +78,12 @@ def _item_actions(self, item, grid=False):
                  request_method='GET',
                  renderer='../templates/edit.jinja2')
     def view(self):
+        if self._is_new():
+            dbsession = self._request.dbsession
+            default_wms = dbsession.query(LayerWMS). \
+                filter(LayerWMS.name == 'wms-defaults').one_or_none()
+            if default_wms:
```
julsbreakdown: just being prickly:
if dbsession.query(LayerWMS). \
                filter(LayerWMS.name == 'wms-defaults').one_or_none():
would avoid an extra line =')

arnaud-morvan: @julsbreakdown : `default_wms` is used another time on the next line ;)

@cmangeat : I would like a helper directly in model, to avoid repetition, for example :
default_wms = LayerWMS.default(dbsession)
or something similar

admin/acceptance_tests/layers_wms_test.py
```diff
@@ -393,15 +417,13 @@ def test_index(self, selenium, selenium_app, layer_wms_test_data):
         index_page.select_page_size(10)
         index_page.check_pagination_info('Showing 1 to 10 of 26 rows', 10)
 
-        WebDriverWait(selenium, 10).until(
-            lambda driver: driver.execute_script(
-                'return (window.jQuery != undefined && jQuery.active == 0)'))
+        index_page.wait_jquery_to_be_active()
```
arnaud-morvan: Should not be needed as pagination info is already uptodate

admin/c2cgeoportal_admin/views/layers_wms.py
```diff
@@ -81,6 +81,11 @@ def _item_actions(self, item):
                  request_method='GET',
                  renderer='../templates/edit.jinja2')
     def view(self):
+        if self._is_new():
+            dbsession = self._request.dbsession
+            default_wms = LayerWMS.get_default(dbsession)
+            if default_wms:
+                return self.copy(default_wms, excludes=['name', 'layer'])
```
arnaud-morvan: I would prefer to override `self._new_object()`

admin/c2cgeoportal_admin/views/layers_wmts.py
```diff
@@ -75,6 +75,11 @@ def _item_actions(self, item):
                  request_method='GET',
                  renderer='../templates/edit.jinja2')
     def view(self):
+        if self._is_new():
+            dbsession = self._request.dbsession
+            default_wmts = LayerWMTS.get_default(dbsession)
+            if default_wmts:
+                return self.copy(default_wmts, excludes=['name', 'layer'])
```
arnaud-morvan: Same, override self._new_object()

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -963,6 +963,10 @@ def __init__(
         self.time_mode = time_mode
         self.time_widget = time_widget
 
+    @staticmethod
+    def get_default(dbsession: Session) -> DimensionLayer:
```
arnaud-morvan: Why DimensionLayer and not LayerWMS ?

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -1025,6 +1029,10 @@ def __init__(self, name: str='', public: bool=True, image_type: str='image/png')
         DimensionLayer.__init__(self, name=name, public=public)
         self.image_type = image_type
 
+    @staticmethod
+    def get_default(dbsession: Session) -> DimensionLayer:
```
arnaud-morvan: Why DimensionLayer and not LayerWMS ?

cmangeat: Because we are declaring LayerWMS...

commons/c2cgeoportal_commons/models/main.py
cmangeat: Because we are declaring LayerWMTS...

PR #3470, opened 2018-02-24
Subscribers

sbrunner:


arnaud-morvan:
Next step should be to use config.add_translation_dirs in includeme in each package instead of custom localizer. https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_translation_dirs

sbrunner:
Thanks @arnaud-morvan

PR #3472, opened 2018-02-24
No timestamp on docker build

sbrunner:


Makefile
```diff
@@ -97,8 +97,8 @@ build: c2c-egg \
 	$(L10N_PO_FILES) \
 	$(APPS_FILES) \
 	admin/tests.ini \
-	/build/adminbuild-docker.timestamp \
-	/build/build-docker.timestamp \
+	adminbuild-docker \
+	build-docker \
```
arnaud-morvan: I would prefer PHONY task names beginning with a verb like : `docker-build-build` and `docker-build-testdb` (because this run a `docker build` command (here the verb is docker build), so it is easier to understand, like `compile-catalog` for example)

PR #3445, opened 2018-02-16
[WIP] [admin] Add themes ordering by drag&drop

arnaud-morvan:

PR #3477, opened 2018-02-26
admin, two column for ogc_server when edit

cmangeat:


commons/c2cgeoportal_commons/models/main.py
```diff
@@ -765,7 +765,8 @@ class OGCServer(Base):
     ]
     __colanderalchemy_config__ = {
         'title': _('OGC Server'),
-        'plural': _('OGC Servers')
+        'plural': _('OGC Servers'),
+        'widget': FormWidget(fields_template='ogcserver_fields')
```
arnaud-morvan: Note that it is preferable to override the widget in views or schema because colanderalchemy take precedence. I've moved other ones in https://github.com/camptocamp/c2cgeoportal/pull/3445, feel free to merge and I will add this one in my PR.

PR #3478, opened 2018-02-26
update version of tilecloud and tilecloud-chain

jwkaltz:


jwkaltz:
CI fails:
`Collecting c2cwsgiutils (from tilecloud-chain==1.4.1->-r dev-requirements.txt (line 87))
  Could not find a version that satisfies the requirement c2cwsgiutils (from tilecloud-chain==1.4.1->-r dev-requirements.txt (line 87)) (from versions: )`

This is because tilecloud-chain in its new version now has this as requirement: c2cwsgiutils>=1,<2
But in reality that should be satisfied by https://pypi.python.org/pypi/c2cwsgiutils/1.6.1

So I do not know why it is not found.

fredj:
maybe because c2cwsgiutils 1.6.1 is python3 only?

jwkaltz:
@fredj  OK thanks I did not realize that, in fact, I did not realize that the newer tilecloud-chain is not meant for usage with gmf 2.2 in the first place.

PR #3479, opened 2018-02-26
admin, add validator for color

cmangeat:


admin/config.yaml
```diff
@@ -79,6 +83,9 @@ vars:
         type: json
       - name: _unknown
         type: unknown
+      - name: _color
+        type: color
```
sbrunner: Why not instance of throes tow configuration block using a config like this:
```yaml
- name: _color
  type: regexp
  regexp: ^#([A-Fa-f0-9]{{6}}|[A-Fa-f0-9]{{3}})$
  error_message: Expecting hex format for color, e.g. #007DCD
```

admin/config.yaml
```diff
@@ -12,6 +12,10 @@ vars:
     map_zoom: 10
     map_fit_max_zoom: 14
 
+    color_validator:
+        regex: "^#([A-Fa-f0-9]{{6}}|[A-Fa-f0-9]{{3}})$"
+        msg: "Expecting hex format for color, e.g. #007DCD"
+
```
arnaud-morvan: color_validator should be added CONST_vars.yaml in update scaffold : https://github.com/camptocamp/c2cgeoportal/blob/c853cd14be7f3451a989df8f8b691ceb1bd68af6/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl

So you will need to add it in config schema to pass the tests : https://github.com/camptocamp/c2cgeoportal/blob/c853cd14be7f3451a989df8f8b691ceb1bd68af6/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_config-schema.yaml#L154

Maybe in another PR.

arnaud-morvan:
Superseeded by #3486

PR #3481, opened 2018-02-26
Require boto3 instead of boto

fredj:
The two packages are already in the requirements of tilecloud and tilecloud-chain.


fredj:
Because these libraries are not used in c2cgeoportal but in tilecloud. They are already in the requirements.txt of tilecloud.

sbrunner:
But we needs to pin them, and the pining in the tilecloud libraries it only used for the tests ...

PR #3486, opened 2018-02-28
[admin] Add metadata regex type

arnaud-morvan:
Superseed #3479

### DIFF
--- a/admin/config.yaml
+++ b/admin/config.yaml
@@ -79,6 +79,11 @@ vars:
         type: json
       - name: _unknown
         type: unknown
+      - name: _color
+        type: regex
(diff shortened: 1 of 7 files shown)


admin/config.yaml
```diff
@@ -79,6 +79,11 @@ vars:
         type: json
       - name: _unknown
         type: unknown
+      - name: _color
+        type: regex
+        regex: "^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$"
+        msg: "Expecting hex format for color, e.g. #007DCD"
```
sbrunner: msg, what does it mean?
In this case I understand, but for me it's too shorten...

PR #3491, opened 2018-03-01
Use Redis for cache, fix errors, log 

sbrunner:


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -197,10 +197,24 @@ def _get_metadatas(self, item, errors):
...
+
+        wms = None
+        try:
+            wms = WebMapService(None, xml=content)
+        except Exception as e:  # pragma: no cover
+            error = "WARNING! an error occurred while trying to read the mapfile and recover the themes." \
+                "\nURL: {}\n{}".format(url, content)
+            errors.add(error)
+            log.error(error)
+            log.exception(e)
```
pvalsecc: You should never do that. Just just add `exc_info=True` to the previous line.

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -244,39 +257,32 @@ def _wms_getcap_cached(self, ogc_server, _):
...
         try:
             resp, content = http.request(url, method="GET", headers=headers)
-        except Exception:  # pragma: no cover
-            errors.add("Unable to GetCapabilities from url {0!s}".format(url))
-            return None, errors
+        except Exception as e:  # pragma: no cover
+            error = "Unable to GetCapabilities from url {}".format(url)
+            errors.add(error)
+            log.error(error)
+            log.exception(e)
```
pvalsecc: Same here

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: Thanks, fixed :-)

PR #3495, opened 2018-03-02
fix missprint and use Value error instead of JSONDecoderError

cmangeat:


admin/c2cgeoportal_admin/schemas/metadata.py
```diff
@@ -1,5 +1,4 @@
 import json
-from json.decoder import JSONDecodeError
```
arnaud-morvan: Note that it is possible to do : 
try:
    import JSONDecodeError
except:
    JSONDecodeError = ValueError

sbrunner: @arnaud-morvan why this?

PR #3498, opened 2018-03-02
Improve session hanling in admin tests

arnaud-morvan:
This avoid any problem with session instances expiration in admin tests.

### DIFF
--- a/admin/acceptance_tests/conftest.py
+++ b/admin/acceptance_tests/conftest.py
@@ -21,15 +21,16 @@ def dbsession(settings):
     init_db(engine, force=True)
     session_factory = get_session_factory(engine)
     session = get_tm_session(session_factory, transaction.manager)
-    return session
+    yield session
 
 
 @pytest.fixture(scope='function')
 @pytest.mark.usefixtures("dbsession")
 def transact(dbsession):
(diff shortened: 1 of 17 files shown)


admin/acceptance_tests/conftest.py
```diff
@@ -21,15 +21,16 @@ def dbsession(settings):
     init_db(engine, force=True)
     session_factory = get_session_factory(engine)
     session = get_tm_session(session_factory, transaction.manager)
-    return session
+    yield session
```
sbrunner: Just for info, does it change something?

arnaud-morvan: No, but I've used a profiler here, and I've live it like this

sbrunner: OK, thanks :-)

arnaud-morvan:
warnings have been fixed in c2cgeoform, needs to upgrade.
@sbrunner : New issue in jenkings seems related to ngeo. Would appreciate a pinned version of ngeo.

sbrunner:
This one will aspires event if we pin ngeo because OpenLayers don't pin one of his important dependency ..., We are working on it to fix it...

PR #3499, opened 2018-03-02
admin, restriction area

cmangeat:


admin/acceptance_tests/restriction_areas_test.py
```diff
@@ -0,0 +1,136 @@
...
+import re
+from geoalchemy2.shape import from_shape
+from shapely.geometry import box, Polygon, shape
+
+from . import AbstractViewsTests
+
+
+@pytest.fixture(scope='class')
+@pytest.mark.usefixtures('dbsession')
+def restriction_area_test_data(dbsession):
```
arnaud-morvan: You can already set scope to 'function' and add 'transact' in usefixture.
Note that you need to add trasact also as parameter and del transact for linter.
In that case you should remove internal transaction.
I would prefer having a flush before return.

admin/c2cgeoportal_admin/schemas/roles.py
```diff
@@ -6,19 +6,27 @@
...
-        edit_url=lambda request, value: request.route_url(
-            'c2cgeoform_item',
-            table='roles',
-            id=value
-        )
-    ),
-    validator=manytomany_validator
-)
+
+def build(name):
```
arnaud-morvan: Is it possible to give an explicit name, ex : `def roles_schema_node(name):`

admin/c2cgeoportal_admin/schemas/roles.py
```diff
@@ -6,19 +6,27 @@
...
+                table='roles',
+                id=value
+            )
+        ),
+        validator=manytomany_validator,
+        missing=colander.drop
+    )
+
+
+roles_schema_node_for_theme = build('restricted_roles')
```
arnaud-morvan: I would prefer to call this directly in parent schema contruction in views modules.

cmangeat: We were used to clone nodes in view... Is there a difference between scratch building and cloning ?

PR #3500, opened 2018-03-06
admin, wait the page to be effectively refreshed

cmangeat:


admin/c2cgeoportal_admin/views/layertree.py
```diff
@@ -34,19 +32,7 @@ def index(self):
...
-        groups = self._dbsession.query(TreeGroup). \
-            options(subqueryload('children_relation')). \
-            all()
-
-        items = self._dbsession.query(TreeItem). \
-            all()
-
-        themes = self._dbsession.query(Theme). \
-            order_by(Theme.ordering). \
-            all()
```
arnaud-morvan: This is for performance, not functional, it reduce the tree loading by 10 or more

PR #3501, opened 2018-03-06
Remove pyramid_retry

sbrunner:
Do we really need that?


arnaud-morvan:
This comes from cookie cutter template, it was initially placed in the includeme method from the model, I suppose for the case a request to the database fails: https://github.com/Pylons/pyramid-cookiecutter-alchemy/blob/latest/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D/models/__init__.py#L66

sbrunner:
Then I will remove it...

PR #3503, opened 2018-03-07
Move the testing stuff in a dedicated package

sbrunner:


sbrunner:
Thanks @arnaud-morvan

PR #3515, opened 2018-03-08
admin, order fields in restriction area form

cmangeat:


arnaud-morvan:
To simply change order, you do not need to create a custom template.
Custom templates are used to create columns.

PR #3516, opened 2018-03-08
Synchonize model's not null with alembic

arnaud-morvan:
Complete https://github.com/camptocamp/c2cgeoportal/blob/98879de6877ca123ae3571519dafa310590866f8/commons/c2cgeoportal_commons/alembic/main/e004f76e951a_add_missing_not_null.py
Fix GSGMF-309


arnaud-morvan:
@sbrunner : Should be good to synchronize alembic and model using alembic --autogenerate (adding missing clause each side) and use it after that to create migrations.

PR #3518, opened 2018-03-09
Last cleanup before adding admin interface to geoportal

sbrunner:


admin/c2cgeoportal_admin/__init__.py
```diff
@@ -66,4 +66,6 @@ def includeme(config: Configurator):
     config.include('c2cgeoform')
     config.include('c2cgeoportal_commons')
     config.include('c2cgeoportal_admin.routes')
+    # use pyramid_tm to hook the transaction lifecycle to the request
+    config.include('pyramid_tm')
```
arnaud-morvan: 'pyramid_tm' included twice in case of c2cgeoportal_admin.main

admin/acceptance_tests/left_menu_test.py
```diff
@@ -29,11 +29,6 @@ class TestLeftMenu(AbstractViewsTests):
 
     _prefix = '/roles'
 
-    def test_404(self, test_app):
```
cmangeat: even if the page does not exist, were were used to have nav bar displayed... It still work so with demo.

arnaud-morvan: Yes, but we do not want admin 404 with navbar displayed for non admin users.

arnaud-morvan: Would be cool to have the admin 404 with navbar in case we are in /admin, but this should be handled in general not_found_view of the scaffold app.

arnaud-morvan: It may be possible to register a /admin/* route (to be verified) but will not handle the case a NotFound exception is raised when a record does not exist.

arnaud-morvan: could be cool to provide a helper to render the admin 404 in global not_found_view.

admin/c2cgeoportal_admin/__init__.py
sbrunner: Thanks, fixed

admin/acceptance_tests/left_menu_test.py
sbrunner: For me we should target this point later, I relay don't want that interfere with the main application...

PR #3484, opened 2018-02-28
Admin and geoportal in the same final environment 

sbrunner:

PR #3519, opened 2018-03-09
Admin and geoportal in the same final environment

sbrunner:
Replace #3484


sbrunner:
Thanks @arnaud-morvan, I just add a check on `layertree/nodes?order=asc` :-)

sbrunner:
Thanks for the review :-)

PR #3521, opened 2018-03-09
Transmit only some headers in the ressource proxy

sbrunner:


c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -263,10 +263,20 @@ vars:
...
           #exempletargetname: http://www.camptocamp.com/?param1=%s&param2=%s
 
+        # List of allowed headers
+        allowed_headers:
+        - Cache-Control
+        - Content-Encoding
+        - Content-Language
+        - Content-Type
+        - Date
+        - etag
```
fredj: I don't know if the case is important but it should be `ETag`

sbrunner: +1, I will change that :-)

PR #3527, opened 2018-03-13
Better performances on 2.2

sbrunner:
Before an authenticated query on the theme after an unauthenticated query on the theme takes more than 1 minutes on BS-ref
Now it takes less than 10s.


sbrunner:
Thanks :-)

ger-benjamin:
Nice !

PR #3543, opened 2018-03-16
admin, search for layer item name in a translation factory

cmangeat:


admin/c2cgeoportal_admin/__init__.py
```diff
@@ -15,6 +15,7 @@
 c2cgeoform.default_search_paths = search_paths
 
 _ = TranslationStringFactory('admin')
+client_translation_string_factory = TranslationStringFactory('admin')
```
arnaud-morvan: Should be something like "{}-client".format(config.root_package.__name__)

admin/c2cgeoportal_admin/views/layertree.py
```diff
@@ -54,6 +53,7 @@ def render_node(nodes, item, parent_id=None, path=''):
                 'id': item.id,
                 'item_type': item.item_type,
                 'name': item.name,
+                'translated_name': client_translation_string_factory(item.name),
```
arnaud-morvan: We discussed that we will call request.localizer here to avoid including client domain in _

admin/c2cgeoportal_admin/__init__.py
sbrunner: `'{}-client'.format(Config.get('package')`

admin/c2cgeoportal_admin/views/layertree.py
sbrunner: @arnaud-morvan I don't understand, for me this is correct...

arnaud-morvan: This return a TranslationString object, not a translated text. The tranlations needs to be done by request.localizer. Calling localizer here with the good domain will avoid to add client domain in auto_translate.

sbrunner: +1 @arnaud-morvan => something like:
```
request.localizer.translate(client_translation_string_factory(item.name))
```

admin/acceptance_tests/layertree_test.py
```diff
@@ -136,6 +138,19 @@ def test_nodes(self, test_app, layertree_test_data):
             assert 'http://localhost/layertree/unlink/{}/{}'.format(group_id, item_id) == action['url']
             test_app.delete(action['url'], status=200)
 
+    @patch(
+        'c2cgeoportal_admin.views.layertree.TranslationStringFactory',
+        new=lambda x: (str(x) + '_{}_').format)
```
arnaud-morvan: '{}_translated'

PR #3546, opened 2018-03-16
[admin] support translation in geoportal env

arnaud-morvan:


arnaud-morvan:
@sbrunner : I'm not completely aware with transifex tasks, but it seems already implemented, the domain exists in Transifex with some not relevant strings for the moment, I suppose this will be updated by CI after merge in master.

commons/c2cgeoportal_commons/models/__init__.py
```diff
@@ -34,6 +34,12 @@
 import sqlalchemy.ext.declarative
 import sqlalchemy.ext.declarative.api
 
+try:
+    from pyramid.i18n import TranslationStringFactory
+    _ = TranslationStringFactory('c2cgeoportal_admin')
+except ImportError as e:
+    _ = lambda s: s
```
arnaud-morvan: @sbrunner: I blocked here, lingua message extractor does not like python3 typing, mypi does not like python3 wihout typing and flake does not like lambda ...

arnaud-morvan:
```
def null_translation(s: str) -> str:
        return s
    _ = null_translation
```
should pass the message extractor ...

Makefile
```diff
@@ -398,10 +403,10 @@ geoportal/c2cgeoportal_geoportal/locale/c2cgeoportal_geoportal.pot: \
 	mkdir --parent $(dir $@)
 	pot-create --keyword _ --config $< --output $@ $(SRC_FILES)
 
-admin/c2cgeoportal_admin/locale/c2cgeoportal_admin.pot: lingua.cfg
+admin/c2cgeoportal_admin/locale/c2cgeoportal_admin.pot: admin/lingua.cfg $(ADMIN_SRC_FILES)
 	$(PRERULE_CMD)
 	mkdir --parent $(dir $@)
-	pot-create --keyword _ --config $< --output $@ $(SRC_FILES)
+	pot-create --config $< --output $@ $(ADMIN_SRC_FILES)
```
sbrunner: Can't we share the same config file?

admin/c2cgeoportal_admin/templates/navigation_navbar.jinja2
```diff
@@ -28,8 +28,9 @@
         <li id="language-dropdown" class="dropdown">
           <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">{{request.locale_name}} <span class="caret"></span></a>
           <ul class="dropdown-menu">
-            <li><a href="{{request.route_url('locale', _query=(('language', 'en'),))}}">en</a></li>
-            <li><a href="{{request.route_url('locale', _query=(('language', 'fr'),))}}">fr</a></li>
+            {% for locale in request.registry.settings['available_locale_names']|sort() %}
```
sbrunner: for me this syntax looks strange, and why do we sort it, for me we should keep the order...

admin/config.yaml
```diff
@@ -4,6 +4,8 @@ vars:
   schema_static: main_static
   srid: 21781
 
+  available_locales_names: en fr de it
```
sbrunner: in array => `[en, fr, de, it]`

commons/c2cgeoportal_commons/models/__init__.py
```diff
@@ -34,6 +34,14 @@
 import sqlalchemy.ext.declarative
 import sqlalchemy.ext.declarative.api
 
+try:
+    from pyramid.i18n import TranslationStringFactory
+    _ = TranslationStringFactory('c2cgeoportal_admin')
+except ImportError as e:
+    def null_translation(s: str) -> str:
+        return s
+    _ = null_translation
```
sbrunner: Can't just be:
```
def _(s: str) -> str:
    """ For localized string search ```
    return s
```
Because `TranslationStringFactory` don't return a string, than the translations should be in the views or renderer ...

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/+package+_geoportal/subscribers.py_tmpl
```diff
@@ -15,8 +17,11 @@ def add_localizer(event):
     localizer = get_localizer(request)
 
     def auto_translate(string):
-        result = localizer.translate(tsf1(string))
-        return localizer.translate(tsf2(string)) if result == string else result
+        for tsf in [tsf1, tsf2, tsf3, tsf4]:
+            result = localizer.translate(tsf(string))
+            if result != string:
+                break
+        return result
```
sbrunner: Can't we use deferent `auto_translate` for geoportal and admin?

With for e.g. `if request.path_info.starts_with('/admin/')`?

admin/c2cgeoportal_admin/templates/navigation_navbar.jinja2
arnaud-morvan: Concerning the syntax, it is a jinja2 file, here I use a filter. For me, it is common to sort entries in dropdown lists. This is just : user friendly.

commons/c2cgeoportal_commons/models/__init__.py
arnaud-morvan: Message extractor fails on _(s: str), note that it search for something like _('ma chaine').

Makefile
arnaud-morvan: We could, but we may have different options, geoportal do not use jinja2 for example.

admin/c2cgeoportal_admin/templates/navigation_navbar.jinja2
sbrunner: For me it's the integrator who should be able to do what he want...

commons/c2cgeoportal_commons/models/__init__.py
sbrunner: OK, I understand, but my first message was unclear because the mine part is that for me we don't needs the try part the content of the expect part should be enough, not?

Makefile
sbrunner: Yes but we can add it for booth?
In general I prefer to have on file it it stay simple :-)

commons/c2cgeoportal_commons/models/__init__.py
arnaud-morvan: I've taken it from main.py and static.py. We do not need the try if we do not use the model outside of pyramid. Ok, to remove it.

sbrunner: +1 for me :-)

arnaud-morvan:
Thanks for review, I treated all comments.
Concerning tests, this could be done later once we get translations from transifex.

PR #3547, opened 2018-03-16
Fix HOME when using make as entry point

arnaud-morvan:
Here I fix HOME when running make directly ex : make docker-build.
HOME_DIR seems always equals to HOME so I suppose it could be removed completely.
With this I can now create a amorvan.mk file with custom tasks for special dev purposes.
Ex: 
```
.PHONY: testgeomapfish
testgeomapfish:
	# ./docker-run make build-docker
	export SRID=21781 EXTENT=489246.36,78873.44,837119.76,296543.14
	./docker-run --image=camptocamp/geomapfish-build --share ${HOME}/workspace pcreate --scaffold=c2cgeoportal_create \
		--ignore-conflicting-name --package-name testgeomapfish ${HOME}/workspace/testgeomapfish --overwrite
	./docker-run --image=camptocamp/geomapfish-build --share ${HOME}/workspace pcreate --scaffold=c2cgeoportal_update \
		--ignore-conflicting-name --package-name testgeomapfish ${HOME}/workspace/testgeomapfish --overwrite
```

### DIFF
--- a/docker-run
+++ b/docker-run
@@ -96,7 +96,6 @@ def main():
     else:
         # For Jenkins slave in Docker
         docker_adrs = netifaces.gateways()[netifaces.AF_INET][0][0]
-    docker_cmd.append("--env=HOME_DIR={}".format(os.environ["HOME"]))
     docker_cmd.append("--env=DOCKER_HOST_={}".format(docker_adrs))
     docker_cmd.append("--env=BUILD_VOLUME_NAME={}".format(build_volume_name))
     docker_cmd.append("--env=PROJECT_DIRECTORY={}".format(os.getcwd()))
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ VALIDATE_PY_TEST_FOLDERS = geoportal/tests
 SPHINX_FILES = $(shell find doc -name "*.rst" -print)
 SPHINX_MAKO_FILES = $(shell find doc -name "*.rst.mako" -print)
 
-HOME=$(HOME_DIR)
+HOME_DIR=$(HOME)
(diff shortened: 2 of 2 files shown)


docker-run
```diff
@@ -96,7 +96,6 @@ def main():
     else:
         # For Jenkins slave in Docker
         docker_adrs = netifaces.gateways()[netifaces.AF_INET][0][0]
-    docker_cmd.append("--env=HOME_DIR={}".format(os.environ["HOME"]))
```
sbrunner: HOME at line 86 should be changed to `HOME_DIR`...

arnaud-morvan: I do not understand why we need a HOME_DIR variable, all times equals to HOME.

arnaud-morvan: This cause a lot of confusion as we set HOME=HOME_DIR and HOME_DIR=HOME so where is the chicken and where is the egg ?

sbrunner: Yes but at line 86 https://github.com/camptocamp/c2cgeoportal/blob/17507da584407ab6e5d08fa90c72ab1d5aa3b8de/docker-run#L86 there still have a deference on ``HOME``, that should be changed to ``HOME_DIR``, to have ``--env=HOME_DIR`` in the command line to transmit the ``HOME_DIR`` ...

sbrunner: Sorry I'm wrong I switch HOME and DOME_DIN in my head...

sbrunner:
I think that's a better solution:
https://github.com/camptocamp/c2cgeoportal/compare/home-dir
That should pass the CI
The difference between HOME and HOME_DIR will be only between docker-run and run, HOME is (in run) the home directory of root, and HOME_DIR of the user.

PR #3556, opened 2018-03-20
Avoid error when the path can have only one element

sbrunner:


sbrunner:
In the previous version if the path has only one element it just fail with an exception, it's what I want to avoid :-)

PR #3558, opened 2018-03-20
Admin demo

cmangeat:


admin/c2cgeoportal_admin/__init__.py
```diff
@@ -25,7 +25,7 @@ def main(_, **settings):
     settings.update(configuration.get_config())
 
     config = Configurator(settings=settings)
-    config.include('c2cwsgiutils.pyramid.includeme')
+    config.include(pyramid_tm.includeme)
```
sbrunner: Why not `config.include('pyramid_tm.includeme')` as the others?

PR #3563, opened 2018-03-20
Fix usage of HOME_DIR

sbrunner:
Alternate of #3547


sbrunner:
Effectively, thanks :-)

PR #3564, opened 2018-03-20
Removes the directories if existing

sbrunner:


sbrunner:
More explicit, the code more readable with long arguments :-)

PR #3566, opened 2018-03-20
The usr path always exists, but shouldn't always be used

sbrunner:


sbrunner:
Thanks for the review :-)

PR #3572, opened 2018-03-21
add _base_query to interface

julsbreakdown:


admin/c2cgeoportal_admin/views/interfaces.py
```diff
@@ -30,6 +33,11 @@ class InterfacesViews(AbstractViews):
     _model = Interface
     _base_schema = base_schema
 
+    def _base_query(self):
+        return self._request.dbsession.query(Interface). \
+            options(subqueryload('layers')). \
```
arnaud-morvan: For me it make no senses to diplay the list of layers in interface (maybe more that 1000 layers in one interface) this may unnecessary increase load on database.

arnaud-morvan: If you want to display them, consider limiting your subqueryload to names

julsbreakdown:
As @arnaud-morvan says, displaying all Layers & Themes  in the interfaces does not make real sense.
Closing

PR #3573, opened 2018-03-21
Add missing PRERULE_CMD and touch

sbrunner:


sbrunner:
It's copy past from the CONST_Makefile :-)

PR #3586, opened 2018-03-22
Add users to role test

julsbreakdown:
'relation "main.user" does not exist' on role update

Fixes 9268a1dffac0 alembic migration


julsbreakdown:
#3596

PR #3588, opened 2018-03-22
Move email function to common, simplify

sbrunner:
Prepare to use for welcome message with:
```
        send_email_config(
            self.request.registry.settings, "welcome_email", user.email,
            user=user.username, password=password
        )
```

### DIFF
--- a/commons/c2cgeoportal_commons/lib/email_.py
+++ b/commons/c2cgeoportal_commons/lib/email_.py
@@ -28,28 +28,58 @@
 # either expressed or implied, of the FreeBSD Project.
 
 
+import logging
 import smtplib
+from typing import Dict, Any, List, cast
+from socket import gaierror
 from email.mime.text import MIMEText
 from email.mime.multipart import MIMEMultipart
 from email.utils import formatdate
+from pyramid.httpexceptions import HTTPInternalServerError
 
 
-def send_email(from_addr, to_addrs, body, subject, smtp_config):  # pragma: no cover
+LOG = logging.getLogger(__name__)
+
+
+def send_email_config(
+    settings: Dict[str, Any],
+    email_config_name: str,
+    email: str, **kwargs: Dict[str, Any]
+) -> None:  # pragma: no cover
+    smtp_config = settings['smtp']
+    email_config = cast(Dict[str, str], settings[email_config_name])
+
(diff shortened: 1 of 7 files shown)


commons/c2cgeoportal_commons/lib/email_.py
```diff
@@ -28,28 +28,58 @@
...
+    try:
+        send_email(
+            email_config['email_from'], [email],
+            email_config['email_body'].format(kwargs),
+            email_config['email_subject'],
+            smtp_config,
+        )
+    except gaierror:
+        LOG.error('Unable to send the email.', exc_info=True)
+        raise HTTPBadRequest('See server logs for details')
```
arnaud-morvan: This seems to hide the original error, nothing bad in http request, maybe we should leave the error to the caller, or raise something in 500 range.
We may leave error handling to the caller.

commons/c2cgeoportal_commons/lib/email_.py
```diff
@@ -28,28 +28,58 @@
...
 
 
-def send_email(from_addr, to_addrs, body, subject, smtp_config):  # pragma: no cover
+LOG = logging.getLogger(__name__)
+
+
+def send_email_config(
+    settings: Dict[str, Any],
+    email_config_name: str,
+    email: str, **kwargs: Dict[str, Any]
```
arnaud-morvan: dest email could be List[str]

sbrunner: You see a reason of that?

commons/c2cgeoportal_commons/lib/email_.py
sbrunner: I will change to HTTPInternalServerError.
This don't hide the error because `exc_info=True` will do it.
The goal of this function is to be simple to use, it we want to do more complex things we can use `sent_email`

commons/c2cgeoportal_commons/lib/email_.py
arnaud-morvan: Not for the moment. Do you a reason to restrict it ?

sbrunner: Just to be simpler to use :-)

PR #3590, opened 2018-03-22
Make translation postfix overwritable

kalbermattenm:
Because an user might want more then one unique translation file (*.po)

No need to backport to 2.2.


sbrunner:
Thanks :-)

PR #3599, opened 2018-03-26
Be able to build es6 version with Webpack

sbrunner:


sbrunner:
No, the build works only in Docker :-)

PR #3602, opened 2018-03-26
admin, send mail when creating user

cmangeat:
can be merged so, call to common mail utilities are blocking calls, but has to work 'nominal', i.e. mail facilities have to work...

### DIFF
--- a/admin/acceptance_tests/user_test.py
+++ b/admin/acceptance_tests/user_test.py
@@ -6,6 +6,8 @@ from pyramid.testing import DummyRequest
 
 from . import skip_if_ci, AbstractViewsTests
 from .selenium.page import IndexPage
+from unittest.mock import patch, MagicMock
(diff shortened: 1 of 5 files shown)


admin/acceptance_tests/user_test.py
```diff
@@ -38,6 +40,11 @@ def users_test_data(dbsession, transact):
     }
 
 
+EXPECTED_WELCOME_MAIL = \
+    'Hello {},\n\nYou have a new account on GeoMapFish demo: https://geomapfish-demo.camptocamp.com/2.3\n' + \
+    'Your user name is: {}\nYour password is: {}\n\nSincerely yours\nThe GeoMapfish team\n'
```
julsbreakdown: Isn't 'Sincerely yours' a bit much/personal
=> Best regards

PR #3605, opened 2018-03-27
Fix node_modules route

cmangeat:


admin/c2cgeoportal_admin/routes.py
```diff
@@ -3,7 +3,10 @@
 
 
 def includeme(config):
-    config.add_static_view('node_modules', 'c2cgeoportal_admin:node_modules')
+    config.add_static_view('node_modules_for_insider', 'c2cgeoportal_admin:node_modules')
+    config.add_static_view(
+        'node_modules_for_outsider',
+        '{}:node_modules'.format(config.root_package.__name__))
```
arnaud-morvan: Are you sure it make sense to add two static views with same name but different path.

arnaud-morvan:
I do not understand what you've done here, is it to the admin module to register routes for outsiders ?

cmangeat:
https://github.com/camptocamp/c2cgeoform/blob/master/c2cgeoform/__init__.py#L57 -> the outsider
https://github.com/camptocamp/c2cgeoportal/blob/master/admin/c2cgeoportal_admin/templates/layertree.jinja2#L4 -> one insider ?

arnaud-morvan:
ok, thanks

PR #3606, opened 2018-03-27
FTS: Use similarity by default

ger-benjamin:
For GEO-908

In 2.3, we want to use "similarity" by default.

As "similarity" is now used by default, should I add a "CREATE EXTENSION pg_trgm;" in an alembic script ?

### DIFF
--- a/doc/integrator/fulltext_search.rst
+++ b/doc/integrator/fulltext_search.rst
@@ -196,22 +196,26 @@ following variables:
 Ranking system
 --------------
 
-By default, the full-text search uses `tsvector` and `ts_rank_cd` to rank your search results (See:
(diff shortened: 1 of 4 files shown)


ger-benjamin:
@sbrunner two questions:
1. As "similarity" is now used by default, should I add a "CREATE EXTENSION pg_trgm;" in an alembic script ?
2. What is the problem with jenkins ? The link give me a 404....

sbrunner:
1. No it's not the role of alembic to do that...

2. I didn't see that in the current build ...

ger-benjamin:
@sbrunner it is okay like that ? (news changes since your review)

PR #3607, opened 2018-03-27
remove dead code

julsbreakdown:
No longer used


julsbreakdown:
Yes I delete the wrong part...

cmangeat:
Not sure of it, but there is still an main here:
https://github.com/camptocamp/c2cgeoportal/blob/dd7caec58e4e89738941a1edf1e7116848538aaf/commons/c2cgeoportal_commons/testing/initializedb.py#L31

PR #3618, opened 2018-03-28
declare admin permission to be required for all views

cmangeat:


arnaud-morvan:
+1 for use of context manager as we do not know what can happen in config.scan as it imports all package modules.

PR #3624, opened 2018-03-29
Geo694 expire login

ger-benjamin:
For: https://jira.camptocamp.com/browse/GEO-694
For GMF 2.3.0

Tasks:

A) In the DB / model
* [x] Add a column last_login (datetime not null)
* [x] Add a column expire_on (datetime)
* [x] Add a column deactivated (boolean)

B) In the admin
* [x] We must be able to see but not edit the last connection datetime
* [x] We must be able to deactivate a user (deactivated = true )
* [x] We must be able to activate a user (deactivated = false)
* [x] We must be able to set manually the expire_on.

C) At login (in c2cgeoportal)
* [x] Check if a expiration_date exists, use it to determine if the user is allowed to be connected.
* [x] Check if the user is not deactivated, to allow the user to be connected.
* [x] Update the "last_login" column with the current date if the login is a success

X)
* [x] In ngeo adapt "Unable to login" message
* [x] Verify that the client give an utc datetime
* [x] Add tests
* [x] Add a jira ticket to inform that the user is only unable to log in if the user is expired/deactivated. So it will be able to keep it's connection as long as he haves a valid cookie.
* [x] ~~Add documentation~~ nothing to do

Out of c2c) In a tool script (not commited in c2cgeoportal)
* [x] A supplementary script that deactivate account if now() - last_login > xx days
(done, but to test -> what about last_login == Null ? )


commons/c2cgeoportal_commons/alembic/static/1857owc78a07_add_last_login_and_expiration_date.py
```diff
@@ -0,0 +1,76 @@
...
+revision = '1857owc78a07'
+down_revision = '5472fbc19f39'
+
+
+def upgrade():
+    staticschema = config['schema_static']
+
+    # Instructions
+    op.add_column('user', Column('last_login', DateTime(timezone=True)), schema=staticschema)
+    op.add_column('user', Column('expire_on', DateTime(timezone=True)), schema=staticschema)
```
sbrunner: Are sou sure that we needs the timezone?

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -1478,11 +1480,11 @@ def login(self):
                 return set_common_headers(
                     self.request, "login", NO_CACHE,
                     response=Response(json.dumps(self._user(
-                        self.request.get_user(user)
+                        self.request.get_user(username)
                     )), headers=headers),
                 )
         else:
-            log.info("bad credentials for login '{0!s}'.".format(login))
+            log.info("Expired login or bad credentials for login '{0!s}'.".format(login))
```
sbrunner: Isn't a duplicated message?

commons/c2cgeoportal_commons/models/static.py
```diff
@@ -171,6 +196,12 @@ def validate_password(self, passwd: str) -> bool:
             return True
         return False
 
+    def expired(self) -> bool:
+        return self.expire_on is not None and self.expire_on < datetime.utcnow()
+
+    def set_last_login(self) -> None:
```
sbrunner: update_last_login ...

commons/c2cgeoportal_commons/alembic/static/1857owc78a07_add_last_login_and_expiration_date.py
ger-benjamin: @arnaud-morvan it is useful ? We don't need the information of timezone, no ?

arnaud-morvan: I never put a datetime in a database without timezone as I 'm not sure everybody is in the same timezone.

arnaud-morvan: but I'm not sure we need the time.

arnaud-morvan: the datetime is stored in UTC, then the time zone is ... **UTC**.
when you store a datetime with time zone, the datetime is also stored in UTC but display in local timezone.
So when you store your datetime in UTC like a datetime with time zone why don't you give the timezone ?

arnaud-morvan: Note that creating the field with timezone avoid passing a datetime in bad timezone.

sbrunner: > So when you store your datetime in UTC like a datetime with time zone why don't you give the timezone ?

I don't want to store unneeded data..., and I want to be coherent in all the application...

ger-benjamin: So, I removes it, thanks (that effectively complicate things in some cases)

arnaud-morvan: I'm sorry to insist, but this not useless data, it is important to qualify a datetime with it's timezone. This avoid doing mistakes about timezones.

arnaud-morvan: Mon voisin me dit qu'une datetime sans timezone ca ne vaut rien

arnaud-morvan: No, this is not the effective behavior with timezone enabled.
With time zone enabled, all dates are stored in UTC and returned in local timezone. This ensure compare and sort work as expected and avoid confusion about timezones.
This ensure that all datetimes are stored in UTC transparently.
So you're telling me that you do not want to set timezone to get the behavior you have with it.

ger-benjamin: I can add that by "that effectively complicate things in some cases" i said before, I mean that it breaks tests (and probably the app) because we compare datetime with explicit timezone (the db, with timezone=True) to datetime without explicit timezone (datetime.utcnow())

If we want to use the timezone in the db, we probably need to cast all the method we use to compare time to the time that comes from db. That's probably not hard, but I'm out of time, and that was not required for this feature. And if one day we need timezone in the db, we can still do the change.

sbrunner: What you say it's not what's written in Table 8-9. here: https://www.postgresql.org/docs/9.1/static/datatype-datetime.html

arnaud-morvan: => To address these difficulties, we recommend using date/time types that contain both date and time when using time zones. 
...
All timezone-aware dates and times are stored internally in UTC. They are converted to local time in the zone specified by the timezone configuration parameter before being displayed to the client.

arnaud-morvan: If we do not set timezone in datetime fields, they will not be converted from local timezone to UTC between input and database, and not converted from UTC to local timezone for display.
We need the timezone for correct behavior between database and UI.

sbrunner: And that's for 2.4...

sbrunner: I just open a Jira ticket for that

arnaud-morvan: Have you tested the UI with timezone enabled ? I've done, it works perfectly with time zone enabled. Without datetimes will not be saved as UTC in the database.

ger-benjamin:
Done. See new tickets here: 
 - GSGMF-431 (cookie)
 - GSGMF-422 (timezone)

PR #3630, opened 2018-04-03
Add user.last_login, expire_on and deactivated

arnaud-morvan:
Note that this is based on @ger-benjamin branch, could be squashed before merge to master.

Not sure it make sense to set last_login = now during migration. This could be left to administrator.

Note that in mosts systems we use last_connection instead of last_login, as login is not required in all cases (remember using cookie, kerberos authentication, and maybe others).

### DIFF
--- a/commons/c2cgeoportal_commons/alembic/static/1857owc78a07_add_last_login_and_expiration_date.py
+++ b/commons/c2cgeoportal_commons/alembic/static/1857owc78a07_add_last_login_and_expiration_date.py
@@ -27,7 +27,7 @@
 # either expressed or implied, of the FreeBSD Project.
 
 
-"""                                                                              
+"""
 Add column last_login and expiration_date (both datetime) on
 table main_static."user".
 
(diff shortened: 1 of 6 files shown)


commons/c2cgeoportal_commons/alembic/static/1857owc78a07_add_last_login_and_expiration_date.py
```diff
@@ -36,27 +36,46 @@
...
+    metadata = MetaData()
+    user = Table(
+        'user',
+        metadata,
+        Column('last_login', DateTime(timezone=True)),
+        Column('deactivated', Boolean),
+        schema=staticschema
+    )
+    op.execute(
+        user.update().values({
```
ger-benjamin: According with the spec I have:
 - last login can be null. So it's useless to fulfill it.
 - Also, please, remove the not used "deactivated" field.

(Or ask wolfgang /yves to agree new specs)

ger-benjamin:
Or have you no time for that ? Perhaps should I re-take this task ?

ger-benjamin:
I merge this branch into mine. Then I'll finish the work.

PR #3633, opened 2018-04-05
Fix GeoMapfishThemeExtractor. Relate d2266025645e48d5a9040fe77a442605…

arnaud-morvan:
…f0c700f3


arnaud-morvan:
@sbrunner: good catch, thanks

PR #3634, opened 2018-04-05
Correctly set static schema name as dynamic varaible in ALTER TABLE c…

ochriste:
…ommand.


sbrunner:
What's the status of this pull request?

PR #3635, opened 2018-04-06
more verbose lingua extractor

jwkaltz:
makes lingua extractor more visible, useful when a config entry is wrong


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -454,6 +453,7 @@ def _layer_attributes(self, url, layer):
             if hostname == "localhost":  # pragma: no cover
                 h["Host"] = self.package["host"]
             try:
+                print(u"Get WMS GetCapabilities for URL {} with headers {}".format(wms_getcap_url, h))
```
sbrunner: This is a relay Ugly print ..., The header dictionary should be manually formated ...

Issue #3637, opened 2018-04-09
Fix dommant in https://github.com/camptocamp/c2cgeoportal/pull/3635


sbrunner:

jwkaltz:
Solved in https://github.com/camptocamp/c2cgeoportal/pull/3639

PR #3639, opened 2018-04-09
more beautiful header print

jwkaltz:
For https://github.com/camptocamp/c2cgeoportal/issues/3637

### DIFF
--- a/c2cgeoportal/lib/lingua_extractor.py
+++ b/c2cgeoportal/lib/lingua_extractor.py
@@ -453,7 +453,9 @@ class GeoMapfishThemeExtractor(Extractor):  # pragma: no cover
             if hostname == "localhost":  # pragma: no cover
                 h["Host"] = self.package["host"]
             try:
-                print(u"Get WMS GetCapabilities for URL {} with headers {}".format(wms_getcap_url, h))
+                print(u"Get WMS GetCapabilities for URL {}".format(wms_getcap_url))
+                for key, value in h.items():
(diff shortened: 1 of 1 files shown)


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -453,7 +453,9 @@ def _layer_attributes(self, url, layer):
             if hostname == "localhost":  # pragma: no cover
                 h["Host"] = self.package["host"]
             try:
-                print(u"Get WMS GetCapabilities for URL {} with headers {}".format(wms_getcap_url, h))
+                print(u"Get WMS GetCapabilities for URL {}".format(wms_getcap_url))
+                for key in h:
+                    print(u"--- GetCapabilities using header {} with value {}".format(key, h[key]))
```
sbrunner: Look good but in Python we prefer this syntax:
```
                for key, value in h.items():
                    print(u"--- GetCapabilities using header {} with value {}".format(key, value))
```
:-)

jwkaltz:
CI fails for a reason unrelated to this PR. Will investigate that in a separate PR. 
Merging.

jwkaltz:
Hmm actually the new CI error
"c2cgeoportal/lib/lingua_extractor.py:429:5: C901 'GeoMapfishThemeExtractor._layer_attributes' is too complex (31)"

results from exactly this PR - the new for loop for nicer printing now made the whole function too complex according to flake8...

PR #3643, opened 2018-04-09
admin, set acl for admin views: 'role_admin' can all

cmangeat:
Wonder where to declare acl in order them to be taken into account when scaffold update, in scaffold create resources or in c2cgeoportal ?

As authentification policy is set in constructor, authorization policy default to acl, cf. lib/python3.5/site-packages/pyramid/config/__init__.py line 427.


arnaud-morvan:
@cmangeat : As you changed things in scaffold create, you inform integrators about it in https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl

PR #3659, opened 2018-04-10
Store the raster file in an LRU cache

fredj:


geoportal/c2cgeoportal_geoportal/views/raster.py
```diff
@@ -70,6 +76,23 @@ def raster(self):
         set_common_headers(self.request, "raster", NO_CACHE)
         return result
 
+    def _get_raster(self, path):
+        dataset = _datasets.get(path)
+        if dataset is None:
+            log.info('cache miss for dataset')
```
sbrunner: log.debug :-)

fredj:
The `WeakValueDictionary` doesn't work in this case: no reference is kept so the dict entry is immediately removed

PR #3678, opened 2018-04-12
Use the released version of bootstrap-table

sbrunner:


arnaud-morvan:
Looks good, needs to be tested in browser.
I suppose opencollective could be removed.

sbrunner:
Tested here: https://geomapfish-demo.camptocamp.com/sbrunner23/
Looks to work well :-)

arnaud-morvan:
Also note that we may use the minified version now in admin/c2cgeoportal_admin/templates/index.jinja2

arnaud-morvan:
I was using the non minified version because the repository minified version seems to be updated update on release tag only.

PR #3699, opened 2018-04-19
Import pwd where it's used to avoid arror on windows

### LINKED ISSUE #3698: Upgrading to 2.3 - pwd module
In Windows, `pwd` module does not exist... Is there a way I can circumvent this ?

sbrunner:
Should fix  #3698

### DIFF
--- a/docker-run
+++ b/docker-run
@@ -5,7 +5,6 @@ import argparse
 import subprocess
 import netifaces
 import configparser
-import pwd
 import sys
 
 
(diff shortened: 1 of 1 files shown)


kalbermattenm:
This line https://github.com/camptocamp/c2cgeoportal/blob/4f7c65034adae03a9a3a25cc9b3335a2e531a36c/docker-run#L66 has nothing to do with the module, right ?

sbrunner:
No, it's a replacement name in the string template.

Issue #3701, opened 2018-04-20
Short link failing with authenticated user


brossardm:
# Project:

* GitHub URL: https://github.com/camptocamp/demo_geomapfish/tree/2.2
* Host: https://geomapfish-demo.camptocamp.net/2.2/
* GeoMapFish version: 2.2.2

# Running command: 
* ask for a short link
  * the window appears with the shortlink
* login in the interface as admin
* ask for a short link

# Result:

The window appears with no shortlink and and timeout appear with:  Erreur, impossible d'obtenir l'URL réduite. 

![selection_013](https://user-images.githubusercontent.com/11328768/39045433-36dd68bc-4493-11e8-969d-1b3f7f9cfc1e.png)

pfirpfel:
I think, I found the cause:
If you remove the email_body property inside the shortener property of your vars file, the shortener works again. There seems to be an issue with this multi-line syntax. The password reset mechanism seems to be affected as well.

sbrunner:
Done

Issue #3406, opened 2018-01-29
Permalink Bug in 2.2.2


svagap:
Each time a permalink is created, a mail is sent to the logged in User. This makes no sense.

The mail should only be send to the address typed in the email field of the form and only if the "send" button is used. The wrong mail is also sent while using the cgxp interface to create a permalink / shortlink. (of course, as it is the same function serverside)

ybolognini:
Done

PR #3708, opened 2018-04-24
Don't send an email to the creator of the short URL

sbrunner:
Fix https://github.com/camptocamp/c2cgeoportal/issues/3701, and https://github.com/camptocamp/c2cgeoportal/issues/3406

### DIFF
--- a/c2cgeoportal/views/shortener.py
+++ b/c2cgeoportal/views/shortener.py
@@ -138,8 +138,6 @@ class Shortener:
         else:
             s_url = self.request.route_url("shortener_get", ref=ref)
 
-        email = email or user_email
-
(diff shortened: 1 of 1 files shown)


c2cgeoportal/views/shortener.py
```diff
@@ -138,7 +138,7 @@ def create(self):
         else:
             s_url = self.request.route_url("shortener_get", ref=ref)
 
-        email = email or user_email
+        email = email
```
fredj: The line can be simply removed

sbrunner: :-P

sbrunner: fixed :-)

PR #3723, opened 2018-05-07
Fix spelling

fredj:


Makefile
```diff
@@ -3,7 +3,7 @@ MAKO_FILES = docker-compose.yaml.mako $(shell find .tx doc docker geoportal/test
 VARS_FILE ?= vars.yaml
 VARS_FILES += vars.yaml
 
-DEVELOPPEMENT ?= FALSE
```
fredj: is this variable still used? `git grep DEVELOPPEMENT` return no results

sbrunner: The correct version is steel used :-)

fredj:
> How do you find all those typo?

https://github.com/lucasdemarchi/codespell

PR #3731, opened 2018-05-09
Fix the module name

sbrunner:


sbrunner:
It will be Appdesktop to simplify, than it's correct :-)

PR #3737, opened 2018-05-09
Add dev environment

sbrunner:


doc/developer/debugging.rst
```diff
@@ -5,14 +5,31 @@ Debugging
...
 
-General
+Webpack
 -------
 
-First you should add ``?debug`` in the application URL to have all the
-JavaScript and stylesheets in separated non-minified files.
+To have all the JavaScript and stylesheets in separated non-minified files.
+
+Add in your makefile ``<user>.mk`` (Each devlopper should have a different port, e.g.: 8081):
```
jwkaltz: devlopper -> developer

doc/developer/debugging.rst
```diff
@@ -5,14 +5,31 @@ Debugging
...
+
+Browser
+-------
 
 Using a browser-integrated debugging tool usually available with the ``F12`` key.
 
+Pyramid
+-------
+
 If the ``pyramid_debugtoolbar`` is enabled the error is directly returned
```
jwkaltz: "is enabled" -> "is enabled,"
"is directly returned" -> "is directly shown"

doc/developer/debugging.rst
```diff
@@ -5,14 +5,31 @@ Debugging
 
 The goal of this document is to give some troubleshooting tips.
 
-General
+Webpack
 -------
 
-First you should add ``?debug`` in the application URL to have all the
-JavaScript and stylesheets in separated non-minified files.
+To have all the JavaScript and stylesheets in separated non-minified files.
```
jwkaltz: I do not understand the connection between the header "Webpack", this sentence ("To have all the JavaScript...") and the following instructions.
Is there something missing?

sbrunner: It's Webpack who minify the js and the css, and we use webpack-dev-derver to don't have this minification...

jwkaltz: So we could say something like:
"For debugging purposes, it is better to have all the JavaScript and stylesheets in separated, non-minified files. To achieve this, you need to disable Webpack's minification; you can achieve this as follows.

What do you think?

sbrunner: I just updated with a variant of your proposal :-)

PR #3754, opened 2018-05-15
More parallal

sbrunner:


Jenkinsfile
```diff
@@ -32,14 +34,13 @@ dockerBuild {
                 clean()
             }
             stage('Build') {
-                checkout scm
```
pvalsecc: Not sure it's a good idea to remove that

PR #3772, opened 2018-05-28
Setup a global front

pvalsecc:


commons/c2cgeoportal_commons/alembic/env.py
```diff
@@ -38,6 +39,11 @@
 
 
 def get_config():
+    os.environ.update({
+        'VISIBLE_WEB_PROTOCOL': 'https',
+        'VISIBLE_WEB_PORT': '443',
+        'VISIBLE_ENTRY_POINT': '/'
```
sbrunner: missing VISIBLE_WEB_HOST ...

pvalsecc: Added

PR #3774, opened 2018-05-28
Cache control

sbrunner:


admin/README.md
```diff
@@ -1,5 +1,28 @@
...
+
+The following views are also provided by `c2cwsgiutils`:
+ * http://localhost:8888/versions.json
+ * http://localhost:8888/stats.json
+ * http://localhost:8888/health_check?max_level=3
+
+
+## Run without docker
+
+
```
gberaudo: Is running without docker unsupported? If no then I think we should keep the lines in gitignore and in the find command.

PR #3778, opened 2018-05-29
Change datetime fields to datetime with timezone

arnaud-morvan:
Note that in SQLAlchemy model, fields was already defined with timezone.


commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
```diff
@@ -0,0 +1,66 @@
...
+depends_on = None
+
+
+def upgrade():
+    staticschema = config['schema_static']
+
+    op.execute('''
+SET TIME ZONE 'UTC';
+ALTER TABLE {staticschema}.user ALTER COLUMN last_login TYPE timestamp with time zone;
+SET TIME ZONE LOCAL;
```
sbrunner: Why it's not the same as previous? it looks like an hack ...

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
```diff
@@ -0,0 +1,66 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    staticschema = config['schema_static']
+
+    op.execute('''
+SET TIME ZONE 'UTC';
+ALTER TABLE {staticschema}.user ALTER COLUMN last_login TYPE timestamp with time zone;
```
sbrunner: I test it but I get the wrong time (shifted of 2 hours), with old and new logins ...

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
arnaud-morvan: Because timezone handling was wrong before.

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
arnaud-morvan: Yes, it is the purpose of this pull request.

arnaud-morvan: Note the +2 at the end of the ISO format in admin UI.
Before datetimes were stored and displayed in UTC.
Now timezones are stored in UTC and displayed in UTC+2.

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
arnaud-morvan: last_login was obtained from datetime.utcnow (stored in UTC)
expire_on was an user input => stored in local time

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
sbrunner: In the admin interface I get `2018-05-30 12:57:22.058844` (table) or `2018-05-30 12:57:22` (form) But I login at `2018-05-30 14:57:22 GMT+2` ...

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
arnaud-morvan: I do not understand what compare does here ?
comparison take two datetimes with time zone, not required to be the same.
What I'm doing here is applying the right time zone depending the one the datetime comes from.

commons/c2cgeoportal_commons/alembic/static/53d671b17b20_add_timezone_on_datetime_fields.py
arnaud-morvan: With my PR you should get 2018-05-30 14:57:22 GMT+2 in admin UI.
Note that pg_admin will show you the date in local time as postgresql already return it in local time.

sbrunner: I do a mistake in my test, and it's effectively working :-)

arnaud-morvan: By creating fields as timestamp with timezone, and creating python datetime objects with a timezone, as datetimes are timezone aware we do not have to worried about timezone afterward and everything should work as expected.

arnaud-morvan:
```
>>> d_utc = datetime.now(pytz.utc)
>>> d_local = d_utc.astimezone(pytz.timezone('Europe/Paris'))
>>> print(d_utc - d_local)
0:00:00
```
When datetimes are time zone aware, all should works as expected !
No ambiguity in comparisons, no ambiguity on what we have in database, no ambiguity in database request and results.
It like coordinates with a projection system.

sbrunner: but: 
```
>>> d1 = datetime.now()
>>> d2 = datetime.utcnow()
>>> print(d1-d2)
2:00:00
```
This mean that it change the comportment...

sbrunner:
You should do a rebase for the CI (see: #3781)

sbrunner:
Yes, this column is consider to be in UTC and in the admin use it as is.

arnaud-morvan:
Seems that everybody do not agree on the expected behavior, because I was having time this week to display and input datetimes in local time. Pretty sure that this is the behavior expected by most of users.
Note that time in expire_on is not really mandatory from my point of view.

PR #3794, opened 2018-06-06
Fix signal send for Gunicorn instead of for Apache

sbrunner:


doc/developer/debugging.rst
```diff
@@ -85,7 +85,7 @@ Edit a file in a running apache WSGI container
 
    docker-compose exec geoportal bash
    vi ...
-   kill -s USR1 1  # graceful
+   kill -s HUP `ps aux|grep gunicorn|head --lines=1|awk '{print $2}'`  # graceful
```
pvalsecc: Then, the title is wrong...

Why don't you suggest to put gunicorn in devel mode (the --reload option) instead?

doc/developer/debugging.rst
```diff
@@ -78,14 +78,20 @@ Actually we display the running rule and why she is running (dependence update).
 Docker
 ------
 
-Edit a file in a running apache WSGI container
+Run gunicorn to reload on modified python file
```
jwkaltz: ... to reload on modifications of python files

doc/developer/debugging.rst
```diff
@@ -78,14 +78,20 @@ Actually we display the running rule and why she is running (dependence update).
...
 ..............................................
 
+Add the following environement variable to the geoportal container:
```
jwkaltz: environment

doc/developer/debugging.rst
```diff
@@ -78,14 +78,20 @@ Actually we display the running rule and why she is running (dependence update).
...
+
+``GUNICORN_PARAMS="-b :80 --worker-class gthread --threads 1 --workers 1 --reload"``
+
+Gaceful the running geoportl container
```
jwkaltz: Do a graceful restart of the running geoportal container

sbrunner: Thanks, fixed :-)

PR #3795, opened 2018-06-06
Return a better message

sbrunner:


geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -493,8 +493,11 @@ def call_hook(settings, name, *args, **kwargs):
 
 
 def notfound(request):
-    del request
-    return HTTPNotFound()
+    return {
```
fredj: you can use: `HTTPNotFound(request.path_info)`
see https://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html#pyramid.httpexceptions.HTTPNotFound

sbrunner: No because I will get the same message as with c2cwsgiutils, here: https://github.com/camptocamp/c2cwsgiutils/blob/master/c2cwsgiutils/errors.py#L80

fredj: ok, I get it now !

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -493,8 +493,11 @@ def call_hook(settings, name, *args, **kwargs):
 
 
 def notfound(request):
-    del request
-    return HTTPNotFound()
```
fredj: `HTTPNotFound`can be removed from the python import

sbrunner: Effectively, thanks :-)

PR #3797, opened 2018-06-06
Fix tow variable with the same name

sbrunner:


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -235,7 +235,7 @@ def _collect_app_config(cls, config, filename):
                             value.encode("ascii", errors="replace") if isinstance(value, unicode) else value)
                         enums.append(Message(None, msgid, None, [], u"", u"", (filename, location)))
 
-        metadata = []
+        metadatas = []
```
jwkaltz: I understand it is just a variable name... but metadatas is not a word (there is just metadata, no metadatas). How about metadata_list?

sbrunner: Why not, done :-)

PR #3808, opened 2018-06-12
Fix the list

sbrunner:


sbrunner:
Thanks @jwkaltz for the review :-)

PR #3817, opened 2018-06-13
Parse the WMS capabilities less times.

sbrunner:
One time per process and OGC server,
Not one time per layers.

fix: https://jira.camptocamp.com/browse/GEO-1246


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-build.yaml.mako
```diff
@@ -32,7 +32,7 @@ ${service_defaults('mapserver', 80)}\
       - run
     links:
       - mapserver
-${service_defaults('geoportal', 80)}\
+${service_defaults('geoportal-run', 80)}\
```
pvalsecc: -build would be more explicit

PR #3821, opened 2018-06-14
Add some documentation to be able to have concurent instance on the s…

sbrunner:
…ame server

### DIFF
--- a/doc/developer/debugging.rst
+++ b/doc/developer/debugging.rst
@@ -78,6 +78,66 @@ Actually we display the running rule and why she is running (dependence update).
 Docker
 ------
 
+Multiple dev on one server
+..........................
+
+When you want to run multiple instances on the same server you should:
+
+- Use the global front
+- Use a different docker tag for each instance
+- Use a different project name for each instance
+
+Global front
+............
+
+The global front will offer a unique entry point on port 80 that provide the 'main' project on `/` and the
+others on `/<project_name>/`.
+
+Activate it in the vars:
+
+.. code:: yaml
+
+   vars:
+     docker_global_front: true
+
+Build the project:
+
+.. prompt:: bash
+
+   ./docker-run make build
+
+Run the global front:
+
+.. prompt:: bash
+
+   (cd global-front; docker-compose --project-name=global up --build)
+
+
+And we should defined different instance name for the build:
+
+.. prompt:: bash
+
+   INSTANCE=<name> ./docker-run make build
+
+
+Use a different docker tag
+..........................
+
+Just define an environment variable in the build:
+
+.. prompt:: bash
+
+   DOCKER_TAG=<tag> ./docker-run make build
+
+Use a different project name
+............................
+
+Define the project name when you run the Docker composition:
+
+.. prompt:: bash
+
+   docker-compose --project-name=<name> ...
+
 Run gunicorn to reload on modifications of Python files
 .......................................................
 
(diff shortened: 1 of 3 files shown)


doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
 Docker
 ------
 
+Multiple dev on one server
+..........................
+
+When you want ro run multiple instance on the same server you should:
```
jwkaltz: instances

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+
+- Use the global front
```
jwkaltz: What is that?

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+- Use different docker tag
```
jwkaltz: Use a different docker tag for each instance

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+- Use different project name
```
jwkaltz: Use a different project name for each instance

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+
+Global front
+............
+
+Activate it int the vars:
```
jwkaltz: int -> in

doc/developer/debugging.rst
```diff
@@ -107,6 +164,25 @@ Add in the ``docker-compose.yaml`` file, in the ``geoportal`` service the follow
...
+To expose a service out of the docker-compose you can add a port for the service in the vars, e.g.:
+
+.. code:: yaml
+
+   vars:
+     docker_services:
+       <service>:
+         port: 8086
+
+Be careful one port can be open only one time on a server. To avoid that you can do a curl directly from a
```
jwkaltz: I don't understand how doing a curl helps to avoid the "one port can be open only one time on a server". What are we avoiding?

doc/developer/debugging.rst
```diff
@@ -107,6 +164,25 @@ Add in the ``docker-compose.yaml`` file, in the ``geoportal`` service the follow
          - <c2cgeoportal_git_root>/geoportal/c2cgeoportal_geoportal:/opt/c2cgeoportal_geoportal/c2cgeoportal_geoportal
          - <c2cgeoportal_git_root>/geoportal/c2cgeoportal_admin:/opt/c2cgeoportal_geoportal/c2cgeoportal_admin
 
+Expose a service
+................
+
+To expose a service out of the docker-compose you can add a port for the service in the vars, e.g.:
```
jwkaltz: In my understanding: we are exposing a service out of the docker container, not out of the docker-compose?
Or are we saying here:
To expose a service from the docker container via docker-compose, you can add...

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+Just define a variable environment in the build:
+
+.. prompt:: bash
+
+   DOCKER_TAG=<tag> ./docker-run make build
+
+Use different project name
+..........................
+
+Define the project name when you ren docker-compose:
```
jwkaltz: ren -> run

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+
+.. prompt:: bash
+
+   INSTANCE=<name> ./docker-run make build
+
+
+Use different docker tag
+........................
+
+Just define a variable environment in the build:
```
jwkaltz: a variable environment -> an environment variable

doc/developer/debugging.rst
```diff
@@ -78,6 +78,63 @@ Actually we display the running rule and why she is running (dependence update).
...
+   ./docker-run make build
+
+Run the global front:
+
+.. prompt:: bash
+
+   (cd global-front; docker-compose --project-name=global up --build)
+
+
+And the instance should have different instance name in the build:
```
jwkaltz: I would say: "To define a different instance name for the build:" ?

doc/developer/debugging.rst
sbrunner: See later...

doc/developer/debugging.rst
sbrunner: When you do a curl in a container you can directly access to all the port of all the other container of the composition, than you didn't needs to publish them on the main host....

doc/developer/debugging.rst
sbrunner: No inside the composition we can access to non exposed port ...

doc/developer/debugging.rst
sbrunner: => Within the Docker composition you can access a port of a container, you can achieve this via curl.
This way, you do not need to publish this port on the main host.
:-)

PR #3824, opened 2018-06-15
Make the invalidate call broadcasted to every processes

pvalsecc:


geoportal/c2cgeoportal_geoportal/lib/check_collector.py
```diff
@@ -56,6 +56,7 @@ def check(request):
                 )
                 r = requests.get(params={"max_level": str(host.get("max_level", max_level))}, **url_headers)
                 r.raise_for_status()
+                return r.json()
```
sbrunner: What's for?

pvalsecc: Include the result of the health check in the results of this one

PR #3826, opened 2018-06-18
Use print version 3.14 in Docker mode

sbrunner:
To have the daily updates

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
@@ -11,7 +11,7 @@ services:
 ${service_defaults('config')}\
 
   print:
-    image: camptocamp/mapfish_print:3.12.1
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
```diff
@@ -11,7 +11,7 @@ services:
 ${service_defaults('config')}\
 
   print:
-    image: camptocamp/mapfish_print:3.12.1
+    image: camptocamp/mapfish_print:3.14
```
jwkaltz: Should be 3.14.1, no?

pvalsecc: Not if you want to take bug fixes automatically and have daily rebuilds.

sbrunner: no because we want the daily updates

PR #3828, opened 2018-06-18
Limit print memory usage

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -76,6 +76,9 @@ vars:
         C2C_BASE_PATH: /c2c_tiles
         C2C_REDIS_URL: redis://redis:6379
         C2C_BROADCAST_PREFIX: broadcast_tilecloudchain_
+    print:
+      environment:
+        JAVA_OPTS: -Xmx1024m
```
pvalsecc: Not sure this works... in the other projects, I use CATALINA_OPTS instead.

sbrunner: In the catalina.sh I read:
```
#   CATALINA_OPTS   (Optional) Java runtime options used when the "start",
#                   "run" or "debug" command is executed.
#                   Include here and not in JAVA_OPTS all options, that should
#                   only be used by Tomcat itself, not by the stop process,
#                   the version command etc.
#                   Examples are heap size, GC logging, JMX ports etc.
...
#   JAVA_OPTS       (Optional) Java runtime options used when any command
#                   is executed.
#                   Include here and not in CATALINA_OPTS all options, that
#                   should be used by Tomcat and also by the stop process,
#                   the version command etc.
#                   Most options should go into CATALINA_OPTS.
```
Than there is not a big deference...

sbrunner: Updated :-)

PR #3837, opened 2018-06-19
Make the raster files (profile) work over http(s)

pvalsecc:
Now, we can have something like that in the config:

```yaml
  raster:
    srtm:
      file: /vsicurl/http://url-of-some-http-service/SRTM21781/srtm.shp
      round: 1
```

That http service could be S3.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/raster.py
+++ b/geoportal/c2cgeoportal_geoportal/views/raster.py
@@ -33,7 +33,7 @@ import os
 from decimal import Decimal
 from repoze.lru import LRUCache
 
-import fiona
+from fiona.collection import Collection
 import rasterio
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/views/raster.py
```diff
@@ -82,7 +82,7 @@ def _get_raster_value(self, layer, lon, lat):
         path = layer["file"]
         if layer.get("type", "shp_index") == "shp_index":
 
-            with fiona.open(path) as collection:
+            with Collection(path) as collection:
```
sbrunner: Too simple :-)

pvalsecc: Sorry ;-)

PR #3838, opened 2018-06-20
Fix dev mode

sbrunner:
And also fix admin to admin/ redirect

Tested on Docker and non docker environment

### DIFF
--- a/doc/developer/debugging.rst
+++ b/doc/developer/debugging.rst
@@ -20,9 +20,9 @@ Add in your makefile ``<user>.mk`` (Each developer should have a different port,
 
 .. prompt:: bash
 
-   make serve-<interface>
+   FINALISE=TRUE make serve-<interface>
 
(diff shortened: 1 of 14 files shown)


geoportal/c2cgeoportal_geoportal/views/utils.py
```diff
@@ -0,0 +1,57 @@
...
+from pyramid.view import view_config
+from pyramid.httpexceptions import HTTPFound
+
+from c2cgeoportal_geoportal.lib.caching import set_common_headers, NO_CACHE
+
+log = logging.getLogger(__name__)
+
+
+@view_config(context=Exception, renderer="json", http_cache=0)
+def other_error(exception, request):  # pragma: no cover
```
pvalsecc: Isn't that done in c2cwsgiutils already?

pvalsecc: Then this should be fixed in c2cwsgiutils...

PR #3840, opened 2018-06-20
Qgis dimensions

arnaud-morvan:
Improve QGIS compatibility : 
* Add field property to dimensions.
* Add version_wms to ogc_server.

### DIFF
--- a/c2cgeoportal/models.py
+++ b/c2cgeoportal/models.py
@@ -1021,10 +1021,12 @@ class Dimension(Base):
             cascade="save-update,merge,delete,delete-orphan",
         ),
     )
+    field = Column(Unicode, label=_(u"Field"))
 
-    def __init__(self, name="", value="", layer=None):
+    def __init__(self, name="", value="", layer=None, field=None):
         self.name = name
         self.value = value
+        self.field = field
         if layer is not None:
             self.layer = layer
 
--- a/c2cgeoportal/models.py
+++ b/c2cgeoportal/models.py
@@ -710,6 +710,7 @@ class OGCServer(Base):
     name = Column(Unicode, label=_(u"Name"), nullable=False, unique=True)
     description = Column(Unicode, label=_(u"Description"))
     url = Column(Unicode, label=_(u"Base URL"), nullable=False)
+    version_wms = Column(Unicode, label=_(u"WMS version"))
     url_wfs = Column(Unicode, label=_(u"WFS URL"))
     type = Column(Enum(
         OGCSERVER_TYPE_MAPSERVER,
--- a/c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
+++ b/c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2017, Camptocamp SA
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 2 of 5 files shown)


c2cgeoportal/models.py
```diff
@@ -1021,10 +1022,12 @@ class Dimension(Base):
             cascade="save-update,merge,delete,delete-orphan",
         ),
     )
+    field = Column(Unicode, label=_(u"Field"))
```
sbrunner: Why not put it just around the description?

c2cgeoportal/views/entry.py
```diff
@@ -73,15 +73,17 @@
 
 class DimensionInformation:
 
-    URL_PART_RE = re.compile("[a-zA-Z0-9_\-\+~\.]+$")
+    URL_PART_RE = re.compile("[a-zA-Z0-9_\-\+~\.]*$")
```
sbrunner: Why this change?

c2cgeoportal/models.py
```diff
@@ -710,6 +710,7 @@ class OGCServer(Base):
     name = Column(Unicode, label=_(u"Name"), nullable=False, unique=True)
     description = Column(Unicode, label=_(u"Description"))
     url = Column(Unicode, label=_(u"Base URL"), nullable=False)
+    version_wms = Column(Unicode, label=_(u"WMS version"))
```
sbrunner: Default to 1.1.1?

c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
```diff
@@ -0,0 +1,57 @@
...
+revision = '72eaad5daf34'
+down_revision = '809650bd04c3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    schema = context.get_context().config.get_main_option('schema')
+
+    op.add_column('ogc_server', Column('version_wms', Unicode), schema=schema)
```
sbrunner: Default to 1.1.1?

c2cgeoportal/views/entry.py
```diff
@@ -269,16 +280,20 @@ def _wms_getcap_cached(self, ogc_server, role_id):
...
-            wms = WebMapService(None, xml=content)
-        except:  # pragma: no cover
+            wms = WebMapService(None, xml=content, version=version)
+        except Exception as e:  # pragma: no cover
             error = _(
                 "WARNING! an error occurred while trying to "
                 "read the mapfile and recover the themes."
             )
-            error = u"{0!s}\nURL: {1!s}\n{2!s}".format(error, url, content.encode("utf-8"))
+            error = u"{0!s}\nURL: {1!s}".format(error, url)
```
sbrunner: Why removing the content?

c2cgeoportal/views/entry.py
arnaud-morvan: This is a backport of merged commit in 2.3 (support NULL dimensions), because I need it to work with ADL.

sbrunner: OK :-)

c2cgeoportal/models.py
arnaud-morvan: Because I like field to appear after layer in admin UI

c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
arnaud-morvan: This is done in Entry class `version_wms or '1.1.1'`

c2cgeoportal/views/entry.py
arnaud-morvan: I removed content because it's hard to find error as content is very long.
Note that you have the url, easy to get content if needed.

c2cgeoportal/models.py
sbrunner: The field order can be change like this:
https://github.com/arnaud-morvan/c2cgeoportal/blob/f4f07948167539be05f9291838a13ffbfbeee05e/c2cgeoportal/forms.py#L704-L709

c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
sbrunner: For me the default value should be visible in the admin interface...

c2cgeoportal/views/entry.py
sbrunner: But many times the error is in the content, or related to the content...

arnaud-morvan: I note that it may be useful if status is 200 but response is a service exception.
In case response is a capabilities, like in my use case it's a PITA.
I did not plan to commit, I will revert this point.

c2cgeoportal/models.py
arnaud-morvan: And it's another thing in 2.3

c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
arnaud-morvan: I also like the idea that we leave the choice to the app if the user does not want to force.
(=> set this only if you know what you are doing)
What about if you want to change the default value in future version ?

arnaud-morvan: In 2.3 we can add a description for the field (default to 1.1.1)

c2cgeoportal/models.py
sbrunner: Yes but as I know you didn't move the fields in the model to change the order, or yes?

c2cgeoportal/scaffolds/update/CONST_alembic/main/versions/72eaad5daf34_add_field_ogc_server_version_wms.py
sbrunner: OK for the description :-)

c2cgeoportal/models.py
arnaud-morvan: As I remember, we did not change any order, we show fields in sqlalchemy model order.
Except for some relations, we create schema manually and insert them imperatively at the right place.
Concerning dimensions, as they are embedded in layer form, layer field should be excluded.
So you're right, we could put field before layer.

sbrunner: cool :-)

arnaud-morvan:
Replaced by #4112

PR #3842, opened 2018-06-20
Don't remove the generated alt applications

sbrunner:
And fix all his deployment

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
+++ b/geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
@@ -499,6 +499,8 @@ class C2cUpgradeTool:
 
     def is_managed(self, file_):
         default_project_file = self.get_upgrade('default_project_file')
+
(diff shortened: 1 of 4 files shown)


geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
```diff
@@ -499,6 +499,12 @@ def step7(self, step):
 
     def is_managed(self, file_):
         default_project_file = self.get_upgrade('default_project_file')
+
+        for pattern in default_project_file['hidden']:
+            if re.match(pattern + '$', file_):
+                print("File '{}' hidden by migration config pattern '{}'.".format(file_, pattern))
+                return True
```
pvalsecc: Isn't it weird that a hidden file is managed? Or I don't understand the meaning of hidden in this context...

sbrunner: hidden meens never managed by the upgrade, but I will change that ...

PR #3843, opened 2018-06-21
Set gunicorn timeout to 60s

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -60,6 +60,7 @@ vars:
     geoportal:
       environment:
         <<: *geo-run-env
+        GUNICORN_PARAMS: --bind=:8080 --worker-class=gthread --threads=10 --workers=5 --timeout=30
```
sbrunner: Why using gthread?

PR #3846, opened 2018-06-22
Get the qgisserver plugin from geomapfish_qgisserver repo

sbrunner:
https://github.com/camptocamp/geomapfish_qgisserver

### DIFF
--- a/docker/qgisserver/Dockerfile
+++ b/docker/qgisserver/Dockerfile
@@ -1,3 +1,4 @@
 FROM camptocamp/qgis-server:3
+LABEL maintainer Camptocamp "info@camptocamp.com"
 
 COPY * /var/www/plugins/geomapfish_qgisserver/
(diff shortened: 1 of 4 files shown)


docker/qgisserver/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM camptocamp/qgis-server:3
+LABEL maintainer Camptocamp "info@camptocamp.com"
+
+COPY * /var/www/plugins/geomapfish_qgisserver/
```
pvalsecc: How is the plugin enabled in QGIS?

pvalsecc:
Shouldn't you change the push target to add this image? Add the target to build it? Change the docker-compsoe to use it?

docker/qgisserver/Dockerfile
sbrunner: As I remember QGIS take everything that's in this folder

sbrunner:
The goal is to add what's in the previous repository with minimal changes, the integration with the CI will comm later :-)

PR #3848, opened 2018-06-22
Be able to run as non-root

sbrunner:
Fix for kubernetes (RunAsGroup is not yet available)


geoportal/c2cgeoportal_geoportal/scaffolds/create/bin/eval-templates
```diff
@@ -3,7 +3,7 @@
 export VISIBLE_WEB_HOST_RE_ESCAPED=`python3 -c "print(__import__('re').escape('${VISIBLE_WEB_HOST}'))"`
 export VISIBLE_ENTRY_POINT_RE_ESCAPED=`python3 -c "print(__import__('re').escape('${VISIBLE_ENTRY_POINT}'))"`
 
-find /etc /mapcache /usr/local/tomcat/webapps/ -name '*.tmpl' -print | while read file
+find /etc /usr/local/tomcat/webapps/ -name '*.tmpl' -print | while read file
```
pvalsecc: Why removing mapcache?

sbrunner: Because now mapcache is in /etc

PR #3852, opened 2018-06-25
Windows cannot run get-pip-dependencies without shell & update pip

kalbermattenm:
because pip == 10 is much better integrated for installing  (& compiling) wheels

Thanks for any review (and eventually merging this if Travis & co. are happy)


sbrunner:
@kalbermattenm I just change the base to be 2.3, than you should do a rebase ...

kalbermattenm:
Oups... I am not so gifted for rebases ;-)

PR #3853, opened 2018-06-25
Windows cannot run get-pip-dependencies without shell & update pip

kalbermattenm:
because pip == 10 is much better integrated for installing (& compiling) wheels

Thanks for any review (and eventually merging this if Travis & co. are happy)

Replaces #3852

(sorry about that @sbrunner)


kalbermattenm:
I just realized that `networkx` will not install itself through the common pip installation process.

Maybe I am going to add this as an exception to this PR.

PR #3855, opened 2018-06-25
Print can be disable again

sbrunner:
Removes no more used MAPSERVER var

Fix: https://jira.camptocamp.com/browse/GSGMF-495


doc/integrator/make.rst
```diff
@@ -84,7 +84,6 @@ The following variables may be set in the makefiles:
 * ``CGXP_INTERFACES``: List of CGXP interfaces, default is empty.
 * ``NGEO_INTERFACES``: List of ngeo interfaces, default is ``mobile desktop``.
 * ``PRINT``: Mapfish print is enable, default to ``TRUE``.
```
jwkaltz: enable -> enabled
default to -> default is

sbrunner: Thanks, updated :-)

PR #3864, opened 2018-06-26
Mapcache as nonroot

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
```diff
@@ -41,6 +41,7 @@ ${service_defaults('tinyows', 80)}\
 
   mapcache:
     image: camptocamp/mapcache:1.6
+    user: www-data
     volumes_from:
       - config:ro
 ${service_defaults('mapcache', 80)}\
```
pvalsecc: You should change the port, I guess...

sbrunner: Thanks :-)

PR #3870, opened 2018-06-27
Don't use the tinyows entrypoint

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose.yaml.mako
```diff
@@ -37,6 +37,7 @@ ${service_defaults('mapserver', 80)}\
     image: camptocamp/tinyows
     volumes_from:
       - config:ro
+    entrypoint: []
```
pvalsecc: Why?

sbrunner: To be sure that nothing in the default entrypoint create a side effect ...

PR #3871, opened 2018-06-27
Update httplib2

kalbermattenm:
Because the current version does not support properly SSL v3. Hence we get a handsake failure when trying to access URLs using SSL v3, like https://wms.geo.admin.ch, through the ogcproxy route (used by the WMS loading plugin). The obtained error using the current version:

```
SSLHandshakeError: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:590)
```

Thanks

### DIFF
--- a/c2cgeoportal/scaffolds/update/CONST_versions.txt
+++ b/c2cgeoportal/scaffolds/update/CONST_versions.txt
@@ -13,7 +13,7 @@ GeoAlchemy2==0.2.4
 GeoFormAlchemy2==2.0.dev3
 geojson==1.0.9
 http-parser==0.8.3
-httplib2==0.9
+httplib2==0.11.3
 isodate==0.5.1
 Jinja2==2.7.3
 js.jqgrid==4.4.4


jwkaltz:
@kalbermattenm I have this problem in a different context, and I am wondering: did you try this in your installation, and it solved your problem? (I am wondering, because updating httplib2 did not help in my context).

kalbermattenm:
Yes it did... From https://sitn.ne.ch, I was not able to reach the geo.admin WMS getCapabilities anymore. Updating this wheel resolved the problem:

capture

I read quite a lot of stuff regarding this error. Many hints on Stackoverflow suggested updating openSSL. With our sysadmin we monitored the executable activities during an gmf ogcserver proxy request and realized that openSSL was not used on Windows. I do not know if this is the case on Linux

After this monitoring I found out that:

capture2

So updating httplib2, updated `cacerts.txt` which seemed to resolve the problem.

I am really not a SSL specialist. I just get the basic concepts, but it looks like some stuff was added to this file between the two httplib2 versions which seems to resolve the problem. Again, our servers are running on Windows (and Python 2.7.10), I do not know if this helps on Linux.

Finally, I read also that httplib2 was kind of a dev thing, making it possible to dig deeply into http concepts and that it might be easier to use a more "standard" library (urllib, urllib2) for production systems.

fredj:
@kalbermattenm, @sbrunner, @jwkaltz : what's the status of this pull request ?

sbrunner:
@kalbermattenm?

kalbermattenm:
It has been updated in our project and I would find it good to merge it into c2cgeoportal for legacy purposes (although the codacy and Travis failure are quite obscure for me...)

fredj:
ok, I've rebased the change into origin/1.6. Let's see if travis is happy now

kalbermattenm:
Thanks  a lot @fredj !

PR #3876, opened 2018-06-27
Be able to create a new project with exta internet access

sbrunner:


pvalsecc:
Do you mean "with no internet access"?

PR #3880, opened 2018-06-28
Fix deep recursion detection in groups

arnaud-morvan:


arnaud-morvan:
@yjacolin : here is your fix for deep recursion

Issue #1826, opened 2015-11-27
https - add options to ignore certificate error on localhost


ochriste:
we need to ignore certificate error (verify=False) for all request on localhost
because any request done using something like https://localhost/... will fail at the moment

jwkaltz:
This issue is actually a problem in a current upgrade in a specific environment. I re-open and will try to do a PR.

sbrunner:
https://jira.camptocamp.com/browse/GSGMF-549

PR #3884, opened 2018-06-29
configurable http

jwkaltz:
Addresses https://github.com/camptocamp/c2cgeoportal/issues/1826 (so far, only for lingua_extractor functionality, because currently only needed there).

### DIFF
--- a/c2cgeoportal/lib/lingua_extractor.py
+++ b/c2cgeoportal/lib/lingua_extractor.py
@@ -370,12 +370,7 @@ class GeoMapfishThemeExtractor(Extractor):  # pragma: no cover
                 callback(item, messages)
 
     def _get_http(self):
-        http_options = self.config.get("http_options")
-        disable_ssl_certificate_validation = False
-        if http_options is not None:
-            disable_ssl_certificate_validation = http_options.get("disable_ssl_certificate_validation", False)
-        http = httplib2.Http(disable_ssl_certificate_validation=disable_ssl_certificate_validation)
-        return http
+        return httplib2.Http(**self.config.get("http_options", {}))
 
     def _import_layer_wms(self, layer, messages):
         server = layer.ogc_server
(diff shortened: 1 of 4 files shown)


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -369,6 +369,14 @@ def _import(object_type, messages, callback=None):
             if callback is not None:
                 callback(item, messages)
 
+    def _get_http(self):
+        http_options = self.config.get("http_options")
+        disable_ssl_certificate_validation = False
+        if http_options is not None:
+            disable_ssl_certificate_validation = http_options.get("disable_ssl_certificate_validation", False)
+        http = httplib2.Http(disable_ssl_certificate_validation=disable_ssl_certificate_validation)
+        return http
```
sbrunner: simpler:
```
return httplib2.Http(**self.config.get("http_options", {}))
```
```

File: c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl [hunk 2]
```diff
@@ -245,6 +245,10 @@ vars:
     # architecture.
     external_themes_url:
 
+    # Available options for http sessions created within c2cgeoportal
+    http_options:
```
sbrunner: Should be added in CONST_config-schema.yaml

c2cgeoportal/lib/lingua_extractor.py
jwkaltz: Indeed, thanks

c2cgeoportal/scaffolds/update/CONST_vars.yaml_tmpl
jwkaltz: OK

jwkaltz:
@sbrunner can you please review once more?

sbrunner:
Please when you have this kind oh hugly commits squash your commits, it makes the merge harder ...

PR #3891, opened 2018-07-02
make checker work on non-standard environments (via parameterization)

jwkaltz:
use http options throughout c2cgeoportal and use config parameter to decide whether to force to localhost on checker

### DIFF
--- a/c2cgeoportal/scaffolds/create/project.yaml.mako_tmpl
+++ b/c2cgeoportal/scaffolds/create/project.yaml.mako_tmpl
@@ -1,7 +1,10 @@
 project_folder: {{project}}
 project_package: ${package}
 host: ${host}
+# Legacy configuration option for checker
 checker_path: /${instanceid}/wsgi/check_collector?
+# New configuration option for checker
+#checker_url: https://${host}/${instanceid}/wsgi/check_collector?
 template_vars:
     package: {{package}}
     srid: {{srid}}
--- a/c2cgeoportal/scripts/c2ctool.py
+++ b/c2cgeoportal/scripts/c2ctool.py
@@ -214,21 +214,32 @@ class C2cTool:
             return yaml.safe_load(f)
 
     def test_checkers(self):
-        http = httplib2.Http()
+        # Read project options for checker http session connection
+        checker_url_base = self.project.get("checker_url")
+        # If parameter not present, use legacy parameter
+        if checker_url_base is None:
+            print("test_checkers missing configuration setting checker_url, using legacy configuration")
+            checker_url_base = "http://localhost{}".format(self.project["checker_path"])
+
+        http = httplib2.Http(disable_ssl_certificate_validation=True)
         for check_type in ("", "type=all"):
+            checker_url = "{}{}".format(checker_url_base, check_type)
+            checker_headers = {"Host": self.project["host"]}
+            print("Calling checker via URL {}, headers {}".format(checker_url, checker_headers))
             resp, _ = http.request(
-                "http://localhost{}{}".format(self.project["checker_path"], check_type),
+                checker_url,
                 method="GET",
-                headers={
-                    "Host": self.project["host"]
-                }
+                headers=checker_headers
             )
             if resp.status < 200 or resp.status >= 300:
                 return False, "\n".join([
                     "Checker error:",
-                    "Open `http://{}{}{}` for more informations."
+                    "Run `curl {} '{}'` for more information."
                 ]).format(
-                    self.project["host"], self.project["checker_path"], check_type
+                    ' '.join([
+                        '--header={}={}'.format(*i) for i in checker_headers.items()
+                    ]),
+                    checker_url
                 )
 
         return True, None
--- a/c2cgeoportal/views/checker.py
+++ b/c2cgeoportal/views/checker.py
@@ -32,7 +32,6 @@ import os
 import urllib
 import httplib
 import logging
-from httplib2 import Http
 from json import dumps, loads
 from time import sleep
 from urlparse import urlsplit, urlunsplit
@@ -41,7 +40,7 @@ from subprocess import check_output, CalledProcessError
 from pyramid.view import view_config
 from pyramid.response import Response
 
-from c2cgeoportal.lib import add_url_params
+from c2cgeoportal.lib import add_url_params, get_http
 
 log = logging.getLogger(__name__)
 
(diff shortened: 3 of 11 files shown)


c2cgeoportal/scripts/c2ctool.py
```diff
@@ -214,10 +214,29 @@ def get_project():
...
-        http = httplib2.Http()
+        # Read project options for checker http session connection
+        try:
+            checker_protocol = self.project["checker_protocol"]
+        except KeyError:
+            checker_protocol = "http"
+            print("No checker_protocol defined in project, using default value {}".format(checker_protocol))
+        try:
+            checker_disable_ssl_certificate_validation \
+                = self.project["checker_disable_ssl_certificate_validation"]
```
sbrunner: In 2.3 we never check the certificate => non needs to add an option for that.

c2cgeoportal/scripts/c2ctool.py
```diff
@@ -214,10 +214,29 @@ def get_project():
             return yaml.safe_load(f)
 
     def test_checkers(self):
-        http = httplib2.Http()
+        # Read project options for checker http session connection
+        try:
+            checker_protocol = self.project["checker_protocol"]
+        except KeyError:
+            checker_protocol = "http"
+            print("No checker_protocol defined in project, using default value {}".format(checker_protocol))
```
sbrunner: Don't use an exception handling just to add a massage, just display the URL in any cases...

c2cgeoportal/scripts/c2ctool.py
```diff
@@ -214,10 +214,29 @@ def get_project():
             return yaml.safe_load(f)
 
     def test_checkers(self):
-        http = httplib2.Http()
+        # Read project options for checker http session connection
+        try:
+            checker_protocol = self.project["checker_protocol"]
```
sbrunner: On 2.3 we just add a `checker_url` I think that better to already use the same things, and keep the old parameters it not present...

c2cgeoportal/views/checker.py
```diff
@@ -51,17 +50,20 @@ def build_url(name, url, request, headers=None):
...
-        headers["Host"] = urlfragments.netloc
-    else:
-        url_ = url
-
     settings = request.registry.settings.get("checker", {})
+    rewrite_as_http_localhost = settings.get("rewrite_as_http_localhost", True)
+    log.info("build_url rewrite_as_http_localhost: {}".format(rewrite_as_http_localhost))
+
+    url_ = url
+    if rewrite_as_http_localhost:
```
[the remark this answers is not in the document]
jwkaltz: Some sites do not allow access to http://localhost.
So the goal is to, with a configuration option, be able to override the hard-coded "redirection" of build_url to http://localhost.

c2cgeoportal/scripts/c2ctool.py
jwkaltz: The goal of the exception handling here is not to add a message, but to provide a default value in case the configuration option "checker_protocol" is not present in the project.

sbrunner: => `checker_protocol = self.project.get("checker_protocol", "http")` :-)

c2cgeoportal/views/checker.py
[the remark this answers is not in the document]
jwkaltz: security policy of the site in question

c2cgeoportal/scripts/c2ctool.py
jwkaltz: yes, but this is about 2.2.
So either we add this as an option for 2.2, or we deactivate the certificate check completely. In this PR it has already been implemented as an option, but I can remove the option and deactivate the certificate check everywhere in 2.2 if you prefer.

c2cgeoportal/scripts/c2ctool.py
jwkaltz: OK, I will try to backport the "checker_url" part from 2.3 in this PR.

c2cgeoportal/views/checker.py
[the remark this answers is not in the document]
sbrunner: Witch kind of security policy, what does he check?

c2cgeoportal/scripts/c2ctool.py
sbrunner: +1 to deactivate the certificate check in any cases in 2.2 :-)

sbrunner: But not every where, for the runtime call it should be enable by default.

c2cgeoportal/views/checker.py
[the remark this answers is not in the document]
jwkaltz: http://localhost is not allowed on that server (Connection refused). Security policy is the one of the canton in question, we do not have control over it, and they will not open http://localhost. But if we simply keep the url instead of forcing it to use http://localhost, there is no problem.

c2cgeoportal/scripts/c2ctool.py
jwkaltz: So I am disabling the certificate check in the c2ctool test_checkers (always; no configuration), but keeping the certificate check in the rest (configurable).

c2cgeoportal/scaffolds/create/project.yaml.mako_tmpl
```diff
@@ -1,9 +1,8 @@
 project_folder: {{project}}
 project_package: ${package}
 host: ${host}
-checker_path: /${instanceid}/wsgi/check_collector?
-checker_protocol: http
-checker_disable_ssl_certificate_validation: False
+# Configuration option for checker (starting GMF 2.2.5)
+checker_url: https://${host}/${instanceid}/wsgi/check_collector?
```
sbrunner: Why changing this file?

c2cgeoportal/scripts/c2ctool.py
```diff
@@ -226,9 +235,9 @@ def test_checkers(self):
             if resp.status < 200 or resp.status >= 300:
                 return False, "\n".join([
                     "Checker error:",
-                    "Open `http://{}{}{}` for more informations."
+                    "Open `{}` for more information."
                 ]).format(
-                    self.project["host"], self.project["checker_path"], check_type
+                    checker_url
```
sbrunner: If host is not localhost this message becomes wrong, if you want to change it you should do something like in 2.3, with the curl command...

c2cgeoportal/scaffolds/create/project.yaml.mako_tmpl
jwkaltz: Instead of checker_path, checker_url should now be used in project.yaml (but checker_path will still work like before when used in project configuration)

sbrunner: But it add something to migrate with possible side effect :-/

c2cgeoportal/scripts/c2ctool.py
jwkaltz: OK, new commit coming

c2cgeoportal/views/checker.py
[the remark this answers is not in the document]
sbrunner: I don't like when we don't know more :-/, but so keep it ...

jwkaltz:
@sbrunner , OK now?

c2cgeoportal/scripts/c2ctool.py
```diff
@@ -214,21 +214,32 @@ def get_project():
...
+        # If parameter not present, use legacy parameter
+        if checker_url_base is None:
+            print("test_checkers missing configuration setting checker_url, using legacy configuration")
+            checker_url_base = "http://localhost{}".format(self.project["checker_path"])
+
+        http = httplib2.Http(disable_ssl_certificate_validation=True)
         for check_type in ("", "type=all"):
+            checker_url = "{}{}".format(checker_url_base, check_type)
+            checker_headers = {"Host": self.project["host"]}
+            print("Calling checker via url {}, headers {}".format(checker_url, checker_headers))
```
sbrunner: url => URL

jwkaltz:
Thanks @sbrunner for the review, merging.

PR #3892, opened 2018-07-03
Add mechanism to edit views

gberaudo:


doc/administrator/editing.rst
```diff
@@ -132,3 +132,13 @@ Example:
...
    * Only the geometry will be copied, the attributes will not be.
+
+Edit views
+----------
+
+To be able to edit PostgreSQL views a primary key must be manually configured.
+Add a layer metadata ``geotable_primary_key`` with value the name of the column to use as primary key.
+That column must be of type ``Integer``.
+
+Example:
```
sbrunner: Shouldn't you add an extra new line to be a liste?

c2cgeoportal/lib/dbreflection.py
```diff
@@ -161,16 +161,25 @@ def get_table(tablename, schema=None, session=None):
...
+                autoload=True,
+                autoload_with=engine,
+            )
+        else:
+            table = Table(
+                tablename, metadata,
+                schema=schema,
+                autoload=True,
+                autoload_with=engine,
+            )
```
sbrunner: easier to maintain: 
```
args = [tablename, metadata]
if primary_key is not None:
  # Ensure we have a primary key to be able to edit views
  args.append(Column(primary_key, Integer, primary_key=True))
table = Table(
  *args,
  schema=schema,
  autoload=True,
  autoload_with=engine,
)
```

c2cgeoportal/lib/dbreflection.py
```diff
@@ -161,16 +161,20 @@ def get_table(tablename, schema=None, session=None):
...
+        if primary_key is not None:
+            # Ensure we have a primary key to be able to edit views
+            args.append(Column(primary_key, Integer, primary_key=True))
         table = Table(
-            tablename, metadata,
+            *args,
             schema=schema,
             autoload=True,
-            autoload_with=engine,
+            autoload_with=engine
```
sbrunner: ? for me it's better with a coma...

gberaudo: Strangely I get a 500 error with the coma.

gberaudo: What you wish is not supported in python prior to 3.6.
See https://stackoverflow.com/questions/33350454/syntaxerror-with-passing-kwargs-and-trailing-comma.

sbrunner: I never see that but it's true ...

PR #3893, opened 2018-07-03
Tune the sqlalchemy connections

pvalsecc:
To avoid too many errors when the DB is bounced and to have an optimal
number of connections possible.

https://jira.camptocamp.com/browse/GSGMF-501

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -155,8 +155,16 @@ vars:
   # Database information
   sqlalchemy:
     url: postgresql://{dbuser}:{dbpassword}@{dbhost}:{dbport}/{db}
+    pool_recycle: 30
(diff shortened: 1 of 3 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -155,8 +155,16 @@ vars:
...
+    pool_recycle: 30
+    pool_size: 5
+    max_overflow: 25
+    use_batch_mode: true
   sqlalchemy_slave:
     url: postgresql://{dbuser}:{dbpassword}@{dbhost_slave}:{dbport}/{db}
+    pool_recycle: 30
+    pool_size: 5
+    max_overflow: 25
+    use_batch_mode: true
```
sbrunner: Shouldn't they be added in the CONST_config-schema.yaml?

pvalsecc: Adding it

PR #3894, opened 2018-07-03
Centralize the get_layer_class code in a unique place

gberaudo:
In addition, this should ensure all created classes for a layer are consistent.

### DIFF
--- a/c2cgeoportal/lib/dbreflection.py
+++ b/c2cgeoportal/lib/dbreflection.py
@@ -185,7 +185,7 @@ def get_class(tablename, session=None, exclude_properties=None, primary_key=None
     if exclude_properties is None:
         exclude_properties = []
     tablename, schema = _get_schema(tablename)
-    cache_key = (schema, tablename, ",".join(exclude_properties))
+    cache_key = (schema, tablename, ",".join(exclude_properties), primary_key)
 
     if cache_key in _class_cache:
         return _class_cache[cache_key]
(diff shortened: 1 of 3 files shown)


sbrunner:
Looks good but is it necessary to do it in branche 2.2?

gberaudo:
Yes unfortunately we are not migrating yet the EPFL backend to version 2.3.

c2cgeoportal/lib/dbreflection.py
```diff
@@ -185,7 +185,7 @@ def get_class(tablename, session=None, exclude_properties=None, primary_key=None
     if exclude_properties is None:
         exclude_properties = []
     tablename, schema = _get_schema(tablename)
-    cache_key = (schema, tablename, ",".join(exclude_properties))
+    cache_key = (schema, tablename, ",".join(exclude_properties), primary_key)
```
sbrunner: Relay needed?

c2cgeoportal/views/layers.py
```diff
@@ -127,21 +126,21 @@ def _get_layers_for_request(self):
                 )
             )  # pragma: no cover
 
-    def _get_layer_for_request(self):
+    def get_layer_for_request(self):
```
sbrunner: Why this?

c2cgeoportal/lib/dbreflection.py
gberaudo: Yes I think it makes sense since if the primary_key is changed the class will be different.

c2cgeoportal/views/layers.py
gberaudo: My error. I thought that `_` meant private.

ochriste: shouldnt it ?

PR #3902, opened 2018-07-04
Add print scale notice

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-500

### DIFF
--- a/doc/integrator/print.rst
+++ b/doc/integrator/print.rst
@@ -82,3 +82,9 @@ For this, create a makefile with:
 
    BUILD_RULES = test-packages print
    TEST_PACKAGES = main print
+
+Notes about the scales
+----------------------
+
+The web map and the map printing do not use the same scales. Therefore, if you change the configuration
+of min/max scaling, you should also test map printing to be sure that it also works as expected.


doc/integrator/print.rst
```diff
@@ -82,3 +82,9 @@ For this, create a makefile with:
 
    BUILD_RULES = test-packages print
    TEST_PACKAGES = main print
+
+Notes about the scales
+----------------------
+
+The web map and the print did not use the same scales, then you configure the min/max scale denom you
+should also test it in the print to be sure that is works as expected.
```
jwkaltz: If I understand correctly:

"The web map generation and the map printing do not use the same scales. Therefore, if you change the configuration of min/max scaling, you should also test map printing to be sure that it also works as expected."

sbrunner: What the web map **generation**?

jwkaltz: To distinguish in the processes,
- printing (a map)
- generating (a map) for display in a web browser
These processes use different scaling configurations / mechanisms.

(my understanding - feel free to adapt)

PR #3904, opened 2018-07-05
Add some secure headers

sbrunner:
See also, before:
https://securityheaders.com/?hide=on&source=chromeplugin&q=https://geomapfish-demo.camptocamp.com/2.3
After:
https://securityheaders.com/?hide=on&source=chromeplugin&q=https://geomapfish-demo.camptocamp.com/docker

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
@@ -427,6 +427,14 @@ vars:
 
   lingua_extractor: {}
 
+  global_headers:
+    X-Frame-Options: SAMEORIGIN
+    X-Xss-Protection: 1; mode=block
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: origin
+    Strict-Transport-Security: max-age=31536000; includeSubDomains
+    Content-Security-Policy: script-src 'self'; frame-src 'self'; form-action 'self'
+
   # Control the HTTP headers
   headers:
     index: &header
@@ -521,7 +529,7 @@ vars:
 
   # What web page is authorized to use the API
   authorized_referers:
-    - '{web_protocol}://{host}{entry_point}'
+    - '{web_protocol}://{host}/'
 
   # Hooks that can be called at different moments in the life of the
   # application. The value is the full python name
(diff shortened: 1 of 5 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -519,7 +527,7 @@ vars:
 
   # What web page is authorized to use the API
   authorized_referers:
-    - '{web_protocol}://{host}{entry_point}'
```
pvalsecc: Why removing the entry_point?

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
```diff
@@ -425,6 +425,14 @@ vars:
 
   lingua_extractor: {}
 
+  global_headers:
+    X-Frame-Options: SAMEORIGIN
+    X-Xss-Protection: 1; mode=block
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: origin
+    Strict-Transport-Security: max-age=31536000; includeSubDomains
```
pvalsecc: Then you break localhost access when we develop, no?

Is it possible to remove entries from there with a vars.yaml?

geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -45,7 +45,7 @@ def create_authentication(settings):
         callback=defaultgroupsfinder,
         cookie_name=settings["authtkt_cookie_name"],
         timeout=timeout, max_age=timeout, reissue_time=reissue_time,
-        hashalg="sha512", http_only=True
+        hashalg="sha512", http_only=True, secure=True,
```
pvalsecc: Again, you break development environments where there is no SSL.

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
pvalsecc: Ah, you are saved by that note:
```
Note: The Strict-Transport-Security header is ignored by the browser when your site is accessed using HTTP; this is because an attacker may intercept HTTP connections and inject the header or remove it.  When your site is accessed over HTTPS with no certificate errors, the browser knows your site is HTTPS capable and will honor the Strict-Transport-Security header.
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

geoportal/c2cgeoportal_geoportal/lib/authentication.py
pvalsecc: And this time, I cannot find exceptions saving you...

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_vars.yaml_tmpl
sbrunner: Yes it's possible to remove with the vars.yaml :-)

PR #3912, opened 2018-07-06
Fix the QGIS server plugin

sbrunner:


docker/qgisserver/geomapfish_plugin/accesscontrol.py
```diff
@@ -126,6 +136,9 @@ def layerFilterSubsetString(self, layer):  # NOQA
         """ Return an additional subset string (typically SQL) filter """
         QgsMessageLog.logMessage("layerFilterSubsetString {}".format(layer.dataProvider().storageType()))
 
+        if self.layers is None:
+            raise Exception("The plugin is not initialised")
```
ger-benjamin: These 3 lines appears 4 times. It's not better to have little `assertLayerNotNone(layer)` method ?

PR #3925, opened 2018-07-18
Removes the venv

sbrunner:
https://jira.camptocamp.com/browse/GSGMF-400


sbrunner:
https://jira.camptocamp.com/browse/GSGMF-400

PR #3932, opened 2018-07-19
[Doc] Fix and add target to use to update po files (not automatic)

yjacolin:


doc/integrator/internationalization.rst
```diff
@@ -34,14 +34,20 @@ Build your application.
...
-* ``demo/locale/<lang>/LC_MESSAGES/demo-server.po`` for the server part (should be empty for ngeo interfaces)
+* ``geoportal/<package>/locale/<lang>/LC_MESSAGES/demo-client.po`` for ngeo client
+* ``geoportal/<package>/locale/<lang>/LC_MESSAGES/demo-server.po`` for the server part (should be empty for ngeo interfaces)
 
 .. note::
 
    All the ``#, fuzzy`` strings should be verified and the line should be removed
    (if the line is not removed, the localisation will not be used).
 
+To update the po files you should run the specific target:
```
jwkaltz: Should we say something about which part is no longer automatic / when exactly it is necessary to run this target?

yjacolin: | which part is no longer automatic

What do you mean here?

jwkaltz: If I understand correctly, these updates are no longer "automatic" in the sense that the GMF build will include this. Instead, the user must do a build with a specific target to rebuild the files when the texts have changed?

sbrunner: Shouldn't we add something here?
https://github.com/camptocamp/c2cgeoportal/blob/2.3/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl#L34

yjacolin: @jwkaltz yes, this is not automatic. @sbrunner updated!

sbrunner: Thanks :-)

doc/integrator/internationalization.rst
```diff
@@ -34,14 +34,22 @@ Build your application.
...
-* ``demo/locale/<lang>/LC_MESSAGES/demo-server.po`` for the server part (should be empty for ngeo interfaces)
+* ``geoportal/<package>_geoportal/locale/<lang>/LC_MESSAGES/demo-client.po`` for ngeo client
+* ``geoportal/<package>_geoportal/locale/<lang>/LC_MESSAGES/demo-server.po`` for the server part (should be empty for ngeo interfaces)
 
 .. note::
 
    All the ``#, fuzzy`` strings should be verified and the line should be removed
    (if the line is not removed, the localisation will not be used).
 
+To update the po files you should run the specific target each time you change
```
jwkaltz: I think it would be clearer to say it the other way around:

Each time you change or add a new layer (mapfile, layer item in admin interface, etc.) or add a new feature, you should update the ``po`` files. You can do this by running a specific build target, as follows:

sbrunner: Note that also conserne the print, the fulltextsearch.layername the application (javascript and HTML), the layer enumerations, the translatable metadata (disclamer), the editable layer (db structure, data for enums).

jwkaltz:
This PR is replaced by https://github.com/camptocamp/c2cgeoportal/pull/3995.

PR #3963, opened 2018-08-03
Run as non root, without confd

sbrunner:


geoportal/tests/functional/test_entry.py
```diff
@@ -483,7 +483,7 @@ def test_notmapfile(self):
             "The layer '__test_public_layer' (__test_pub",
             "The layer '__test_layer_in_group' (__test_l",
             "The layer 'test_wmsfeaturesgroup' (test_wms",
-            "GetCapabilities from URL http://mapserver/?"
+            "GetCapabilities from URL http://mapserver:8"
```
pvalsecc: `:8`?

sbrunner: Cutted version of `GetCapabilities from URL http://mapserver:8080/?...` :-)

PR #3974, opened 2018-08-07
Clear the cache on project load

sbrunner:


docker/qgisserver/geomapfish_plugin/accesscontrol.py
```diff
@@ -73,6 +73,10 @@ def __init__(self, server_iface):
 
         server_iface.registerAccessControl(self, int(os.environ.get("GEOMAPFISH_POSITION", 100)))
 
+    def clear_cache(self):
```
pvalsecc: You don't have to annotate it as a slot?

PR #3978, opened 2018-08-08
Add some wheel installation exception when using Windows

kalbermattenm:
because these ones cannot be installed straight from pypi using pip, but I think that they are "dev" packages and they are not blocking any "simple" production instance (behind a simple Apache "mod_wsgi" server and in an non-Docker environment).

Please review and eventually merge (if Travis let me through for once...)


kalbermattenm:
Thanks @sbrunner !

PR #4000, opened 2018-08-14
Try to fix Jenkins 

sbrunner:


docker/test-external-db/config.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+sed -ri "s/#log_statement = 'none'/log_statement = 'all'/g" /var/lib/postgresql/data/postgresql.conf
```
pvalsecc: You can do all that by simply changing the command in you docker-compose:
```yaml
command: postgres -c log_statement=all
```

travis/create-new-project.sh
```diff
@@ -28,6 +28,6 @@ git clean -fX
 
 # Build
 ./docker-run make --makefile=travis.mk build docker-build-testdb
-./docker-compose-run bash -c 'wait-for-db; PGHOST=external-db; wait-for-db;'
+./docker-compose-run bash -c 'wait-db; PGHOST=externaldb PGDATABASE=test wait-db;'
```
pvalsecc: Shouldn't you put a `&&` instead of the `;`?

sbrunner:
Thanks, fixed :-)

PR #4001, opened 2018-08-14
Introduce a PG statement_timeout of 30s

pvalsecc:
Since Gunicorn has no request timeout, this is needed to avoid having
workers busy for too long if a query goes crazy.


sbrunner:
Should be rebased...

PR #4005, opened 2018-08-15
Use requests instead of httplib2 and urllib

sbrunner:


geoportal/tests/functional/test_entry.py
```diff
@@ -1141,8 +1141,7 @@ def test_layer(self):
             ("REQUEST", "GetCapabilities"),
         )
         url = mapserv_url + "?" + "&".join(["=".join(p) for p in params])
-        http = httplib2.Http()
-        http.request(url, method="GET")
+        requests.get(url)
```
pvalsecc: You don't do anything with the result? Don't forget the requests doesn't raise an exception in case of non 2xx status. You have to call `raise_for_status` on the response...
And there seem to be a lot of locations in this PR with this kind of problem.

PR #4010, opened 2018-08-15
Be able to have an http cookie

sbrunner:


geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -40,12 +40,16 @@ def create_authentication(settings):
     timeout = None if timeout is None else int(timeout)
     reissue_time = settings.get("reissue_time")
     reissue_time = None if reissue_time is None else int(reissue_time)
+    https_only = settings.get("auth_https_only")
+    https_only = True if https_only is None else https_only != "False"
```
kalbermattenm: Should that be `https_only = True if https_only is None else False`

Otherwise I do not get what value has to be passed to `auth_https_only` in order for the condition to return `False` ?

(Sorry, it might be a stupid question, but I do not get the `else` part...)

sbrunner: For the else part:
If `https_only` is exactly "False" we get `False`
For all other values we get `True`

kalbermattenm: Got it, sorry for the mess in my head !

sbrunner: :-)

PR #4081, opened 2018-09-04
better error message for lingua_extractor

jwkaltz:


c2cgeoportal/lib/lingua_extractor.py
```diff
@@ -482,7 +482,7 @@ def _layer_attributes(self, url, layer):
                         "parse the GetCapabilities document.",
                         RED))
                     print(colorize(str(e), RED))
-                    print(u"URL: {0!s}\nxml:\n{1!s}".format(wms_getcap_url, content))
+                    print(u"URL: {0!s}\nxml:\n{1!s}i, Layer: {2!s}".format(wms_getcap_url, content, layer))
```
sbrunner: i ...

sbrunner: `{2!s}` => `{}`

sbrunner: Why adding the layer?

jwkaltz: So we can see which layer is causing the error.

sbrunner: But its not relay related to a layer it's more related to an OGC_server ...

jwkaltz: In this method, it is processing a specific layer. So in our case of debugging a build problem SZ GMF 2.2, it helped to understand what was happening when seeing which layer it was trying to translate. But I agree that strictly speaking, showing the getcapabilitiesURL in the error output should be enough to debug the build error, so this PR is optional.

PR #4086, opened 2018-09-05
fix WFS access for QGIS3 clients (and others relying on metadata)

jwkaltz:


c2cgeoportal/scaffolds/create/mapserver/c2cgeoportal.map.mako_tmpl
```diff
@@ -70,7 +70,7 @@ MAP
         METADATA
             "wms_title" "changeme"
             "wms_abstract" "changeme"
-            "wms_onlineresource" "http://${host}/${instanceid}/wsgi/mapserv_proxy"
+            "ows_onlineresource" "http://${host}/${instanceid}/wsgi/mapserv_proxy"
```
sbrunner: On 2.3 I use ``ows_service_onlineresource`` witch is right?

jwkaltz: @yjacolin do you know more? I find both in the docs (ows_service_onlineresource and ows_onlineresource)

sbrunner: Here:
https://mapserver.org/ogc/wms_server.html
https://mapserver.org/ogc/wfs_server.html
I just see one strange mention of `ows_onlineresource` but is it's tested and working we can go ahead like it :-)

jwkaltz: @sbrunner I retested my demo instance with ows_onlineresource, this works, and with ows_service_onlineresource, but this does not work. So I will merge this PR here, and I think the 2.3 branch should be adapted to ows_onlineresource.

sbrunner: Thanks for the tests :-)

PR #4112, opened 2018-09-12
Qgis dimensions

arnaud-morvan:
Add dimension.field for handling dimensions through OGC filters with QGIS server.


arnaud-morvan:
Replaced by  #4125 because travis is failing, cloning camptocamp/c2cgeoportal instead of arnaud-morvan/c2cgeoportal.

PR #4121, opened 2018-09-17
Return None if no data

julsbreakdown:
https://jira.camptocamp.com/browse/GSGMF-629


geoportal/c2cgeoportal_geoportal/views/profile.py
```diff
@@ -113,6 +113,7 @@ def _compute_points(self):
                 if value is not None:
                     values[ref] = value
                     has_one = True
+                else: value[ref] = None
```
sbrunner: value = => values, and finally => 
```
                values[ref] = value
                if value is not None:
                    has_one = True
```

julsbreakdown: ok thanks

sbrunner:
Regarding the tests results the CSV export is broken...

geoportal/tests/test_raster.py
```diff
@@ -211,8 +211,8 @@ def test_profile_csv(self):
             '"coordinates":[[548009.5,215990],[547990,216009.5]]}'
         response = profile.csv()
         self.assertEqual(response.body.decode("utf-8"), """distance,dem,dem2,dem4,x,y
-0.0,1166,1166,-9999,548009.5,215990.0
-9.2,1175,1175,-9999,548003.0,215996.5
+0.0,1166,1166,None,548009.5,215990.0
+9.2,1175,1175,None,548003.0,215996.5
```
sbrunner: This shouldn't change...

julsbreakdown: I do think is has to change => https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/c2cgeoportal_geoportal/views/profile.py#L66
=> i will remove the code that sets -9999

geoportal/c2cgeoportal_geoportal/views/profile.py
```diff
@@ -64,12 +64,6 @@ def csv(self):
         result = _("distance") + "," + ",".join(layers) + ",x,y"
         template = ",".join("%s" for l in layers)
         for point in points:
-            # Handles cases when a layer is undefined, thus when not all raster
-            # have the same geographical coverage
-            for l in layers:
-                if l not in point["values"]:
-                    point["values"][l] = -9999
```
sbrunner: This is wanted ...

sbrunner: For the CSV the client needs -9999 for no data in the export, for the Json we do what we want, and I prefer null, it's clearer ...

julsbreakdown:
@sbrunner is this ok with you ?

PR #4136, opened 2018-09-21
fix if ms:character in describfeaturetype

julsbreakdown:
@sbrunner

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/entry.py
+++ b/geoportal/c2cgeoportal_geoportal/views/entry.py
@@ -1484,10 +1484,10 @@ class Entry:
                                 if 'alias' in c.attrib:
                                     attrib[c.attrib['name']] = c.attrib['alias']
                             types[child.attrib['name']] = attrib
-
                     attributes = {}
                     for name, type_ in elements.items():
-                        attributes[name] = types[type_]
+                        if type_ in types:
+                            attributes[name] = types[type_]
 
                 result["ogcServers"][ogc_server.name] = {
                     "url": url,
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -1484,10 +1484,10 @@ def themes(self):
                                 if 'alias' in c.attrib:
                                     attrib[c.attrib['name']] = c.attrib['alias']
                             types[child.attrib['name']] = attrib
-
                     attributes = {}
                     for name, type_ in elements.items():
-                        attributes[name] = types[type_]
+                        if type_ in types:
+                            attributes[name] = types[type_]
```
sbrunner: I think that we should add a warning message ...

julsbreakdown: in the console?

sbrunner: Yes :-)

sbrunner: In the else :-)

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -1484,10 +1484,13 @@ def themes(self):
...
+                        else:
+                            log.warning("{} does not exist in {}, this is probably due"
+                            "to the use of ms:character in the describeFeatureType".format(type_, types))
```
sbrunner:
```
log.warning(
    "The provided type '{}' does not exist, available types: {}.".format(
        type_, ', '.join(types.keys())
    )
)
```

julsbreakdown: No mention about ms:character?

PR #4151, opened 2018-09-26
Add the missing pg_service.conf.tmpl file

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/qgisserver/pg_service.conf.tmpl
```diff
@@ -0,0 +1,6 @@
+[demo_geomapfish]
```
arnaud-morvan: May be something more generic.

sbrunner: https://github.com/camptocamp/c2cgeoportal/pull/4158 will fix that :-)

PR #4167, opened 2018-10-01
Editing columns order

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-616

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/xsd.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/xsd.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2013-2018, Camptocamp SA
+# All rights reserved.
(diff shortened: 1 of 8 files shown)


arnaud-morvan:
@sbrunner : Seems there is a lot of dead code concerning key "edit_columns" in themes view.
See : https://github.com/camptocamp/c2cgeoportal/blob/2.2/c2cgeoportal/tests/functional/test_themes_edit_columns.py
From my point of view this is not used as it is done by layers.metadata view.
git grep "edit_columns" in ngeo give no results.
If this is not used, I would like to remove it completely.

geoportal/c2cgeoportal_geoportal/lib/xsd.py
```diff
@@ -0,0 +1,85 @@
+
```
sbrunner: Missing copyright

PR #4171, opened 2018-10-03
Merge 2.2 into 2.3

ger-benjamin:
I had a conflict with this file:
2.2: https://github.com/camptocamp/c2cgeoportal/commit/1d28b290cc5ef106fd972e5fb58ebee3188f3489
2.3: https://github.com/camptocamp/c2cgeoportal/blob/2.3/commons/c2cgeoportal_commons/alembic/main/21f11066f8ec_trigger_on_role_updates_user_in_static.py

It's not in this merge, I think we had to move it / adapt it manually. That's why it's already on 2.3

### DIFF
--- a/commons/c2cgeoportal_commons/alembic/main/7530011a66a7_trigger_on_role_updates_user_in_static.py
+++ b/commons/c2cgeoportal_commons/alembic/main/7530011a66a7_trigger_on_role_updates_user_in_static.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2017-2018, Camptocamp SA
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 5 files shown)


arnaud-morvan:
As said in #4118 I think should better creating an alembic merge migration.
Because each one is already in a released version of GMF, so we may have some databases with the two revisions.

arnaud-morvan:
See : https://alembic.zzzcomputing.com/en/latest/branches.html#merging-branches

ger-benjamin:
@arnaud-morvan , so, I can simply remplace the current (2.3) down_revision from:
`down_revision = 'd48a63b348f1'` to `down_revision = ('d48a63b348f1', 'd8ef99bc227e')` ?

arnaud-morvan:
You should not alter the existing migrations as they may already be used (saved in databases), you should create a merge migration.
In some cases, for a migration backport, we may empty the duplicated migration coming from 2.3, but here, it is not necessary as it can run twice.

ger-benjamin:
@arnaud-morvan Thansk again for the precious help !
I've added the missing alembic file in this PR

PR #4174, opened 2018-10-04
--add-host flag

yjacolin:


arnaud-morvan:
you need to add it in docker_cmd ex:
```
for env options.env:
         docker_cmd.append("--env={}".format(env))
```
otherwise it will have no effect

yjacolin:
@arnaud-morvan add-host couldn't be used as a Python variable. Could you check if my option is correct?

sbrunner:
UP ?

PR #4194, opened 2018-10-17
Fix yaml syntax to match expected data format

pfirpfel:
Tested this with @yjacolin. The previous syntax resulted in data structures that did not match the expectations of the angular code.

To explain the unusual syntax:
```
- profiles:
  -
    label: Car
    profile: routed-car
  -
    label: Bike (City)
    profile: routed-bike
```

This results in:
`
[{label: "Car", profile: "routed-car"}, {label: "Bike (City)", profile: "routed-bike"}]`

The `ngeoNominatimSearchDefaultParams` just don't want an array. It is supposed to be just an object.

### DIFF
--- a/doc/integrator/routing.rst
+++ b/doc/integrator/routing.rst
@@ -55,12 +55,14 @@ To configure the routing feature the constants `ngeoRoutingOptions` and `ngeoNom
                     ngeoRoutingOptions:
                         - backendUrl: https://routing.osm.ch/
                         - profiles:
-                            - label: Car
+                            -
+                              label: Car
                               profile: routed-car
-                            - label: Bike (City)
+                            -
+                              label: Bike (City)
                               profile: routed-bike
                     ngeoNominatimSearchDefaultParams:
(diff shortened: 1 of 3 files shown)


pfirpfel:
This:
```
ngeoRoutingOptions:
            - profiles:
                - label: Car
                  profile: route
```
resulted in this:
routing-profile

yjacolin:
@sbrunner @pfirpfel I just tested the second way in the Stéphane example and it works as expected.

sbrunner:
It should also be changed in the following files:
- geoportal/c2cgeoportal_geoportal/scaffolds/create/vars.yaml_tmpl
- geoportal/c2cgeoportal_geoportal/scaffolds/nondockercreate/vars.yaml_tmpl

sbrunner:
UP?

pfirpfel:
I think this can be merged, right?

sbrunner:
If it's OK for you you can do it :-)

PR #4237, opened 2018-10-30
Correct typo 'authtkt' for secure and http_only

ger-benjamin:
Never see this "authtk" string, without the final `t`. I guess it was a typo.

Also, accept `false, False, No, no, 0` as "false value, and not only `False`.
Still `True` by default.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/authentication.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/authentication.py
@@ -40,10 +40,10 @@ def create_authentication(settings):
     timeout = None if timeout is None else int(timeout)
     reissue_time = settings.get("reissue_time")
     reissue_time = None if reissue_time is None else int(reissue_time)
-    http_only = settings.get("authtk_http_only")
-    http_only = True if http_only is None else http_only != "False"
-    secure = settings.get("authtk_secure")
-    secure = True if secure is None else secure != "False"
+    http_only = settings.get("authtkt_http_only", "True")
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -40,10 +40,10 @@ def create_authentication(settings):
     timeout = None if timeout is None else int(timeout)
     reissue_time = settings.get("reissue_time")
     reissue_time = None if reissue_time is None else int(reissue_time)
-    http_only = settings.get("authtk_http_only")
-    http_only = True if http_only is None else http_only != "False"
-    secure = settings.get("authtk_secure")
-    secure = True if secure is None else secure != "False"
+    http_only = settings.get("authtkt_http_only", "")
+    http_only = False if http_only.lower() in ("false", "no", "0") else True
```
sbrunner:
```suggestion
    http_only = not http_only.lower() in ("false", "no", "0")
```

ger-benjamin:
@sbrunner  I can't merge `The base branch restricts merging to authorized user` :-/

sbrunner:
Oups, it should be fixed :-)

geoportal/c2cgeoportal_geoportal/lib/authentication.py
arnaud-morvan:
```suggestion
    from pyramid.settings import asbool
    http_only = asbool(settings.get("authtkt_http_only", "False"))
```

PR #4313, opened 2018-11-15
Automatically create schema

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/testdb/11-schemas.sql
```diff
@@ -0,0 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS postgis;
```
pvalsecc: The name of the file is not matching its content

arnaud-morvan:
@sbrunner : I suppose it is not possible because www-data does not have rights to create schemas.

sbrunner:
In fact before that Alembic will create the version table in the schema and it failed ...

PR #4317, opened 2018-11-16
Simplify dynamic.js

sbrunner:


arnaud-morvan:
@sbrunner : It could be simpler for the integrators if nearly constant routes could be put in CONST_vars.yaml, maybe in another PR.

arnaud-morvan:
OK, this is not trivial due to `defaults: &interface`.

IMHO, I would like to have in vars.yaml : 
```
interfaces:
  defaults: &interface
    constants: &constants
       defaultTheme: Demo 
       defaultLang: en 
       gmfExternalOGCServers: 
         - name: Swiss Topo WMS 
           type: WMS 
           url: https://wms.geo.admin.ch/?lang=fr 
       ngeoWfsPermalinkOptions: 
         wfsTypes: 
           - featureType: fuel 
             label: display_name 
         defaultFeatureNS: http://mapserver.gis.umn.edu/mapserver 
         defaultFeaturePrefix: feature
```
others things may not talk to integrators.

as it works for now, `<<` and anchors are interpreted at yaml load time, but `extends` later at c2c.template time, so it is not possible to mix values coming from CONST_vars.

So ok, it was an idea, but it seems not possible or too complicated. Let's keeping it like this for the moment.

sbrunner:
OK than I should do the overrides manually (meaning not with yaml ref) and the vas.yaml will become in your case:
```
interfaces:
  defaults:
    constants:
       defaultTheme: Demo 
       defaultLang: en 
       gmfExternalOGCServers: 
         - name: Swiss Topo WMS 
           type: WMS 
           url: https://wms.geo.admin.ch/?lang=fr 
       ngeoWfsPermalinkOptions: 
         wfsTypes: 
           - featureType: fuel 
             label: display_name 
         defaultFeatureNS: http://mapserver.gis.umn.edu/mapserver 
         defaultFeaturePrefix: feature
```

and the `<<` will be removed in this case.

PR #4331, opened 2018-11-21
Don't exclude properties everywhere

sbrunner:
Fix regression introduced by https://github.com/camptocamp/c2cgeoportal/commit/496e2bb23194c23b99c72ed377714fc3610369dc

Mainly we shouldn't exclude the lastUpdateDateColumn and lastUpdateUserColumn on insert or update operation otherwise we can't set them.

### DIFF
--- a/c2cgeoportal/views/layers.py
+++ b/c2cgeoportal/views/layers.py
@@ -414,7 +414,7 @@ class Layers:
         layer = self._get_layer_for_request()
         if not layer.public and self.request.user is None:
             raise HTTPForbidden()
-        return get_layer_class(layer)
+        return get_layer_class(layer, with_exclude=True)
 
     @view_config(route_name="layers_enumerate_attribute_values", renderer="json")
     def enumerate_attribute_values(self):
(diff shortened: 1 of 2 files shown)


c2cgeoportal/views/layers.py
```diff
@@ -468,15 +468,18 @@ def query_enumerate_attribute_values(dbsession, layerinfos, fieldname):
...
-    last_update_date = Layers.get_metadata(layer, "lastUpdateDateColumn")
-    if last_update_date is not None:
-        exclude.append(last_update_date)
-    last_update_user = Layers.get_metadata(layer, "lastUpdateUserColumn")
-    if last_update_user is not None:
-        exclude.append(last_update_user)
+def get_layer_class(layer, with_exclude=False):
+    if with_exclude:
+        # Exclude the columns used to record the last features update
+        exclude = [] if layer.exclude_properties is None else layer.exclude_properties.split(",")
```
arnaud-morvan: layer.exclude_properties is not part of "the columns used to record the last features update"
It seems that this line shoud be outside of if block.
I suppose we don't need "excluded properties" in select / insert / update operations.

sbrunner: @arnaud-morvan I don't understand your comment, as I understand the start say the opposite of the end of your comment.

sbrunner: Note that `with_exclude=False` is mostly for insert / update operations, and `with_exclude=True` for metadata operations (Layertree, xsd, lingua)

PR #4339, opened 2018-11-22
[WIP] Use get_appsettings instead of get_app in scripts

arnaud-morvan:
Note that we may use a custom loader to use pyramid get_app and get_appsettings directly, exemple:
```
from plaster_pastedeploy import Loader
from c2c.template.config import config as configuration

class Loader(Loader):

    def _get_defaults(self, defaults=None):
        d = {}
        d.update(os.environ)
        d.update(defaults or {})
        settings = super()._get_defaults(d)
        configuration.init(settings.get('app.cfg'))
        settings.update(configuration.get_config())
        return settings

    def __repr__(self):
        return 'c2cgeoportal.Loader(uri="{0}")'.format(self.uri)
```
Such loader could be acessible through config_uri, for example : `c2c://production.ini`
Note that this will also allow usage of pyramid scripts, for example `proute`.

It also make duplication with things in c2cwsgiutils as it already pass environ as default.
After a little analyse in c2cwsgiutils, it seems that this could be completely removed:

https://github.com/camptocamp/c2cwsgiutils/blob/master/c2cwsgiutils_run
Orchestrator need to be notified when container crash.

https://github.com/camptocamp/c2cwsgiutils/blob/master/c2cwsgiutils/wsgi_app.py
sentry filter should be activated using settings, or config.

https://github.com/camptocamp/c2cwsgiutils/blob/master/c2cwsgiutils/wsgi.py
not more need to pass os.environ to get_app with custom logger.

Also note support for env vars was merged but reverted in plaster_pastedeploy due to a bug in pastedeploy concerning char % in defaults: 
https://github.com/Pylons/plaster_pastedeploy/pull/15/commits/b34a45d604f8c73bdc670b1c2dae7c9aa31f8b56

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scripts/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/scripts/__init__.py
@@ -29,8 +29,13 @@
 
 
 import os
-from pyramid.paster import get_app as paster_get_app
-from logging.config import fileConfig
+from plaster.uri import parse_uri
+from pyramid.paster import get_app as paster_get_app, get_appsettings as paster_get_appsettings
+from pyramid.scripts.common import get_config_loader
+import transaction
+from c2c.template.config import config as configuration
+from c2cgeoportal_commons.testing import (
+    get_engine, get_session_factory, get_tm_session, generate_mappers)
 
 
 def fill_arguments(parser):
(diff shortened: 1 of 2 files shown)


arnaud-morvan:
@sbrunner @pvalsecc : I'd like to discuss about this before going further.

pvalsecc:
> sentry filter should be activated using settings, or config.

It is: https://github.com/camptocamp/c2cwsgiutils/blob/f75dfaf23d14c171f4e9847189f8b673e388eea9/c2cwsgiutils/sentry.py#L65

pvalsecc:
> Orchestrator need to be notified when container crash.

Breaks backward compatibility. We have a lot of docker-compose out there that relies on that. But you can add that to https://github.com/camptocamp/c2cwsgiutils/issues/177

pvalsecc:
> not more need to pass os.environ to get_app with custom logger.
No need for geomapfish. But we have a lot of other projects that relies on that.

arnaud-morvan:
@pvalsecc : 

- there is settings to set filters in pastedeploy : https://pastedeploy.readthedocs.io/en/latest/#filter-composition, if my app use a filter, I would like to see it in settings or config.

- docker-compose have an autorestart functionnality, no ? same thing, if my container autorestart I would like to know it.

- we could put an base config loader in c2cwsgiutils for the env vars and extend it in c2cgeoportal for the yaml config, this will allow the use of pyramid command line tools.

- there is options in gunicorn to chose config_uri : http://docs.gunicorn.org/en/latest/settings.html#paste

- there also env var in gunicorn doc to pass arguments : http://docs.gunicorn.org/en/latest/settings.html#settings

Maybe I miss something, but in brief, I would like to reduce learning curve for experimented developpers with pyramid and other tools and avoid wrappers around wrappers that mask functionalitites from such powerful tools.

arnaud-morvan:
Superseeded by  #4639

PR #4345, opened 2018-11-23
Better class to edit the layer

sbrunner:
See:
https://github.com/camptocamp/c2cgeoportal/pull/4331#discussion_r235298715


geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
```diff
@@ -422,7 +422,7 @@ def _import_layer_wms(self, layer, messages):
         if layer.geo_table is not None and layer.geo_table != "":
             try:
                 from c2cgeoportal_geoportal.views.layers import get_layer_class
```
arnaud-morvan: Seems wreid to import from views here. Method get_layer_class might be in lib.db_reflexion.

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -469,22 +469,32 @@ def query_enumerate_attribute_values(dbsession, layerinfos, fieldname):
         return dbsession.query(distinct(attribute)).order_by(attribute).all()
 
 
-def get_layer_class(layer, with_exclude=False):
-    if with_exclude:
-        # Exclude the columns used to record the last features update
-        exclude = [] if layer.exclude_properties is None else layer.exclude_properties.split(",")
+def get_layer_class(layer, for_metadata=False):
```
arnaud-morvan: `with_last_update_columns=True` might make more sense here as it is also for LiguaExtractor, not only for metadata view.

Note that moving editingAttributesOrder and readonlyAttributes inside if block will create two distinct classes in cache for metadata and protocol, even if there is no last_update_columns, but this might not have a big impact on memory usage, unless there is thousands of editable layers.

sbrunner: Your right I will get them out for the cache ...

sbrunner: Done :-)

geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
sbrunner: It can't be it make initialization issue ...

PR #4357, opened 2018-11-27
QGIS publish docker package

sbrunner:
- Publish a version 3.4 of QGIS
- Don't publish minor versions


docker/qgisserver/Dockerfile34
```diff
@@ -0,0 +1,10 @@
+FROM camptocamp/qgis-server:3.4
```
pvalsecc: FYI: https://docs.docker.com/engine/reference/builder/#arg

PR #4381, opened 2018-12-04
Upgrade to Redis 5, add tcp keepalive to 30s.

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-lib.yaml_tmpl
```diff
@@ -93,6 +93,8 @@ services:
       - --maxmemory
       - 512mb
       - --maxmemory-policy
+      - --tcp-keepalive
+      - 30
```
pvalsecc: In a monoserver context this is useless.

PR #4385, opened 2018-12-04
[WIP] Add n-n relation between user and role

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-719


arnaud-morvan:
@sbrunner : Concerning User.role, when you've moved the user table to the static schema : https://github.com/camptocamp/c2cgeoportal/commit/bdbe97b5df38a4317e094f0bb56313e16324e869#diff-f39bf2dedcde4cab26ddb7fd167d31d5
You removed the relationship and replaced it by a python property, and you changed User.role_id by User.role_name.
I'm not sure to understand.
The role_id seems acceptable here and it is possible to use a relationship even without foreign key by giving the primaryjoin condition.
Now changing a role.name break the relations and we do not have the SQLAlchemy majic coming with the relationship anymore.

commons/c2cgeoportal_commons/alembic/static/ae5e88f35669_add_table_user_role.py
```diff
@@ -0,0 +1,70 @@
...
+        'SELECT id, role_name FROM {staticschema}.user'.
+        format(staticschema=staticschema)
+    )
+
+
+def downgrade():
+    schema = config['schema']
+    staticschema = config['schema_static']
+
+    op.drop_table('user_role', schema='main_static')
```
sbrunner:
```suggestion
    op.drop_table('user_role', schema=staticschema)
```

admin/c2cgeoportal_admin/templates/widgets/user_fields.pt
```diff
@@ -0,0 +1,21 @@
...
+  <div class="col-md-6 col-sm-12">
+    <div tal:repeat="child field.children" tal:omit-tag="">
+      <div
+         tal:condition="child.schema.name in ('roles', 'role_name')"
+         tal:replace="structure child.render_template(field.widget.item_template)">
+      </div>
+    </div>
+  </div>
+
+</div>
```
sbrunner: Missing enlline

sbrunner:
What's sure:
- We can't have an SQL constrains because we should be able to dump restore the schema individually.
- We want to have the same kind of relation for the (name or id) settings_role and the roles.

For the other I don't know, it's possible that can be better...

For me the role property and column should be renamed to e.-g. `settings_role`.

arnaud-morvan:
We have a trigger on role table that update user.role_role but this does not apply on production user table when changing a role_name on integration server.
IMHO we should create a new user.setting_role relationship based on role.id and remove the old one.

arnaud-morvan:
Hope this will not break custom integrations like luxembourg.

sbrunner:
No problem for me :-)

arnaud-morvan:
I still have in mind that it might be clearer for the user with: 

primary_role: sort of main role, used for settings and permissions
secondary_roles or additional_roles: only used for additional permissions

intead of:

settings_role: only used for settings, not for permissions
roles: only used for permissions

arnaud-morvan:
Also note that with primary_role and additional_roles we could put a required on primary_role.

sbrunner:
https://jira.camptocamp.com/browse/GSGMF-721
=> 
`default_settings` and `roles`
In your description `primary_role` is relay not primary and `secondary_roles` is more important than primary...

arnaud-morvan:
Note that on one side it can be more flexible to separate concerns (settings and permissions).
=> settings_role

But on the other side, in large majority of uses cases, user will have the permisions from the settings roles, and I'd like to reduce the number of clicks to get this.
Also note that it might be good to ensure user have the required permissions for his default values.
=> primary_role

Note that Linux choose the primary group / secondary groups strategy.

arnaud-morvan:
As discussed with @sbrunner : Getting permissions from primary_role and roles will complexify data request filters with multiple `join` and `distinct` and we do not want to bring such complexity in mapserver DATA request construction.
=> settings_role and role.

commons/c2cgeoportal_commons/alembic/main/338b57593823_remove_trigger_on_role_name_change.py
```diff
@@ -0,0 +1,80 @@
...
+RETURN NEW;
+END;
+$$
+LANGUAGE plpgsql""".format(schema=schema, staticschema=staticschema))
+    op.execute(
+        'CREATE TRIGGER on_role_name_change AFTER UPDATE ON {schema}.role FOR EACH ROW '
+        'EXECUTE PROCEDURE {schema}.on_role_name_change()'.format(
+            schema=schema
+        )
+    )
```
sbrunner: Any reason for multiplying the alembic script, I think that better to have only one ...

arnaud-morvan: Because we were used to upgrade only static schema during deployments.
=> one script for main schema
=> one script for static schema
Maybe this is not strictly needed anymore.

arnaud-morvan: this is needed because we need one version in each schema.
upgrade main and static on integration
deploy main
upgrade static on prod
Seems this is strictly needed.

sbrunner: I didn't see that's one for main and on for static, then your right it's strictly needed :-), thanks :-)

PR #4398, opened 2018-12-07
Get the theme for the roles

sbrunner:


geoportal/c2cgeoportal_geoportal/views/tinyowsproxy.py
```diff
@@ -135,13 +135,13 @@ def _get_headers(self):
             headers["Host"] = self.settings.get("tinyows_host")
         return headers
 
-    def _proxy_callback(self, operation, role_id, cache_control, *args, **kwargs):
+    def _proxy_callback(self, operation, role: static.Role, cache_control, *args, **kwargs):
```
arnaud-morvan: Shouldn't be user instead of role here ?

geoportal/c2cgeoportal_geoportal/views/tinyowsproxy.py
```diff
@@ -135,13 +135,13 @@ def _get_headers(self):
...
         response = self._proxy(*args, **kwargs)
         content = response.text
 
         if operation == "getcapabilities":
             content = filter_wfst_capabilities(
-                content, role_id, super(TinyOWSProxy, self)._get_wfs_url(), self.request
+                content, role, super(TinyOWSProxy, self)._get_wfs_url(), self.request
```
arnaud-morvan: Same thing, role => user ?

geoportal/c2cgeoportal_geoportal/views/tinyowsproxy.py
sbrunner: Yes, thanks :-)

PR #4404, opened 2018-12-07
Ignore the right file

sbrunner:


arnaud-morvan:
Superseeded by #4399 ?

sbrunner:
No, It's the inverse, I split #4399 on some pull request.

sbrunner:
> I'm surprised to see .env.mako here as I do not see any other file specific to docker version, like docker-compose*.

It's en error that the docker-compose* are not present, here, I will have a look to see if it's easy to add them :-)

PR #4425, opened 2018-12-12
Multi role continue - Fix typo

sbrunner:


arnaud-morvan:
Can we keep the linter before the tests, I prefer having the linter saying hey you've made an error at this line in 1 minutes than having a tests backtrace with 500 lines of code after 30 minutes.

PR #4442, opened 2018-12-17
Upgrade login for multi role

sbrunner:


sbrunner:
Not ready ...

sbrunner:
For me it can be merge as it :-)

doc/developer/webservices.rst
```diff
@@ -279,8 +279,10 @@ Connected JSON result
...
        "username": "<username>",
        "is_password_changed": "True"/"False", # If false the user should change his password
-       "role_name": "<role_name>",
-       "role_id": <role_id>
+       "roles": [{
+           "name": "<role_name>",
+           "id": <role_id>
+       }, ...],
        "functionality": {
            "<functionnality_name>": ["functionnality_value"],
```
jwkaltz: functionality_name and functionality_value
(I know, it's not really part of this PR...)

sbrunner: Fixed :-)

geoportal/c2cgeoportal_geoportal/lib/functionality.py
```diff
@@ -55,12 +56,22 @@ def _get_config_functionality(name, registered, types, request, errors):
...
+        ]
+        return [r for r in result if r is not None]
+
+    result = []
+    for role in user.roles:
+        result.extend([
+            get_typed(name, functionality.value, types, request, errors)
+            for functionality in role.functionalities
+            if functionality.name == name
+        ])
```
arnaud-morvan: You may have many times the same functionnality with same value:
```
result.extend([
        get_typed(name, functionality.value, types, request, errors)
        for functionality in set(
            [f for role in roles for f in role.functionalities if f.name == name]
        )
    ])
```

geoportal/tests/functional/test_functionalities.py
```diff
@@ -127,7 +128,7 @@ def test_functionalities(self):
                 "registered": {},
             },
             "admin_interface": {
-                "available_functionalities": ["__test_a", "__test_s"]
+                "available_functionalities": [{"name": "__test_a"}, {"name": "__test_s"}]
```
arnaud-morvan: Can it be something else than a name here ?

geoportal/c2cgeoportal_geoportal/lib/functionality.py
arnaud-morvan: or maybe better : 
```
result.extend([
        get_typed(name, value, types, request, errors)
        for value in set(
            [f.value for role in roles for f in role.functionalities if f.name == name]
        )
    ])
```

sbrunner: You try to find the more complex solution? For me we should avoid to use `[ ... for ... for ...]`, it's too complex to understand ...

geoportal/tests/functional/test_functionalities.py
sbrunner: Yes we can have: type, regex, error_message, or single, the more important one is single :-)

arnaud-morvan: Yes, we are in tests, but this is the definitions of functionalities.
My bad.

geoportal/c2cgeoportal_geoportal/lib/functionality.py
arnaud-morvan: Ok, so we might create a variable for the values
```
values = set()
for role in roles:
    values = values | set([f.value for f in role.functionalities])
result.extend([
    get_typed(name, value, types, request, errors)
    for value in values
])

Comment (arnaud-morvan) [reply geoportal/c2cgeoportal_geoportal/lib/functionality.py]: This might be more readable:
```
    result.extend([
        get_typed(name, value, types, request, errors)
        for value in set(
            [f.value
             for role in roles
             for f in role.functionalities
             if f.name == name]
        )
    ])
```

Comment (sbrunner) [reply geoportal/c2cgeoportal_geoportal/lib/functionality.py]: Updated with another more readable possibility...

File: geoportal/c2cgeoportal_geoportal/lib/functionality.py [hunk 4]
```diff
@@ -55,13 +56,28 @@ def _get_config_functionality(name, registered, types, request, errors):
...
+        ]
+        return [r for r in values if r is not None]
+
+    else:
+        functionalities = {
+            functionality.value
+            for role in user.roles
+            for functionality in role.functionalities
+            if functionality.name == name
+        }
```
arnaud-morvan: { => [
} => ]
and the first comment was we should use set(functionalities) to have distinct values

sbrunner: {...} directly creates a set ...

sbrunner:
Thanks :-)

PR #4445, opened 2018-12-18
Fix functionalities with milti_role

arnaud-morvan:
I need this to continue working on mapserver.
I did not handle the settings role or single / multivalues functionalities.


sbrunner:
Here I choos to pass the user instance of the role ...
https://github.com/camptocamp/c2cgeoportal/pull/4442

arnaud-morvan:
Thanks, I will rebase

PR #4446, opened 2018-12-18
[fix-master] Add lock for docker-compose-run 

sbrunner:
Add missing docker-compose file

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -108,27 +108,31 @@ dockerBuild {
                     // Test return code
                     sh './docker-run true'
                     sh 'if ./docker-run false; then false; fi'
-                    try {
-                        sh './docker-compose-run true'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -108,27 +108,31 @@ dockerBuild {
...
                     sh 'if ./docker-run false; then false; fi'
-                    try {
-                        sh './docker-compose-run true'
-                        sh 'if ./docker-compose-run false; then false; fi'
-                    } catch (Exception error) {
-                        sh 'docker-compose logs --timestamps'
-                        throw error
-                    } finally {
-                        sh 'docker-compose down'
+                    lock('docker-compose-run') {
```
pvalsecc: A lock is global accros the whole CI farm. I think you want something like that, instead:
```
lock("docker-compose-run-${env.NODE_NAME}") {
```

PR #4465, opened 2019-01-03
Remove version 1 and catalogue option

sbrunner:
Depends on #4464


sbrunner:
Yes, it's why I remove it :-)

PR #4468, opened 2019-01-03
Add all objects in memory view 

sbrunner:


geoportal/c2cgeoportal_geoportal/views/memory.py
```diff
@@ -28,65 +28,79 @@
...
 
-log = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
 
 
 @view_config(route_name="memory", renderer="fastjson")
 def memory(request):
-    del request
-    return _memory()
+    return _memory(
```
pvalsecc: I doubt this code has been tested. broadcast works only with named parameters...

sbrunner: :-), thanks

sbrunner:
Up @pvalsecc  :-)

PR #4477, opened 2019-01-08
Install binary wheels first on Windows

kalbermattenm:
On Windows we cannot install Pyproj through pip anymore. I added this dependency as a binary wheel. Moreover, as it is a dependency of OWSlib, we have to install it first (before OWSlib).

@sbrunner, I do not know if anybody else is using `nondocker-finalise.mk`, but I would be grateful if this PR could be merged (if Travis is happy, as always...)


kalbermattenm:
Thanks a lot @sbrunner &  @fredj !

PR #4479, opened 2019-01-08
Remove some deprecated columns

sbrunner:


commons/c2cgeoportal_commons/alembic/main/c75124553bf3_remove_deprecated_columns.py
```diff
@@ -0,0 +1,60 @@
...
+# revision identifiers, used by Alembic.
+revision = 'c75124553bf3'
+down_revision = 'dba87f2647f9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    schema = config['schema']
+    staticschema = config['schema_static']
```
fredj: variable is not used

commons/c2cgeoportal_commons/alembic/main/c75124553bf3_remove_deprecated_columns.py
```diff
@@ -0,0 +1,60 @@
...
+
+    op.drop_column('layergroup', 'is_internal_wms', schema=schema)
+    op.drop_column('layergroup', 'is_base_layer', schema=schema)
+
+
+def downgrade():
+    schema = config['schema']
+    staticschema = config['schema_static']
```
fredj: variable is not used

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -414,17 +414,6 @@ class LayerGroup(TreeGroup):
             'title': _('Expanded'),
             'column': 2
         }})  # shouldn't be used in V3
```
fredj: should the comment be removed ?

fredj:
there's maybe more code that can removed (`git grep is_internal_wms`)

commons/c2cgeoportal_commons/models/main.py
sbrunner: No because it's steel be the case, but we can't remove it because the script themev1tov2 use it :-)

sbrunner:
Yes, for that we need the multi role to avoid conflicts ...

PR #4484, opened 2019-01-10
Multi role fix last errors

sbrunner:


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -67,8 +66,8 @@
 from c2cgeoportal_geoportal.views.layers import get_layer_metadatas
 
 _ = TranslationStringFactory("c2cgeoportal")
-log = logging.getLogger(__name__)
-cache_region = get_region()
+LOG = logging.getLogger(__name__)
+CACHE_REGION = get_region()
```
arnaud-morvan: Using upper case for variables is not common, usually reserved for constants.

geoportal/tests/functional/test_dbreflection.py
```diff
@@ -27,6 +27,8 @@
 # of the authors and should not be interpreted as representing official policies,
 # either expressed or implied, of the FreeBSD Project.
 
+# pylint: disable=C0111,W0201,W0212
```
arnaud-morvan: I find the human readable name easier to understand, example : # pylint: disable=no-self-use

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: For my IDE it's a constant and I agree with him...

geoportal/tests/functional/test_dbreflection.py
arnaud-morvan: If we need it in all files, it might be possible to add it in config file or command line.

sbrunner: Done :-)

sbrunner: As I know it's not easily possible in the IDE, because I want it only for the test files...

geoportal/c2cgeoportal_geoportal/views/entry.py
arnaud-morvan: I was sure you was using vim, what is your IDE ?

sbrunner: VisualStudio Code :-)

PR #4490, opened 2019-01-11
Fix the redirect from mobile to desktop

sbrunner:


sbrunner:
Fix: https://jira.camptocamp.com/browse/GSGMF-759

geoportal/c2cgeoportal_geoportal/templates/dynamic.js
```diff
@@ -67,12 +67,21 @@ wfs_permalink.update(interface_config.get('wfs_permalink', {}))
...
             _query=no_redirect_query
         )
     else:
+        no_redirect_url = request.route_url(
+            interface_config['redirect_interface'],
+            _query=no_redirect_query
+        )
         url = request.route_url(
             interface_config['redirect_interface'],
             _query=no_redirect_query
```
arnaud-morvan: What is the difference between `no_redirect_url` and `url` ?
Maybe I miss something, but I do not see it.
As I see, at this point : `no_redirect_url === url`

sbrunner: no_redirect_url shoudb have the param `no_redirect=t`... => my PR is incorrect :-(

sbrunner: See  #4492

PR #4493, opened 2019-01-14
Use assertEqual instead of assertEquals

fredj:
`assertEquals` is deprecated, the following warning is raised in the tests:
```
/src/geoportal/tests/functional/test_layers.py:457: DeprecationWarning: Please use assertEqual instead.
    self.assertEquals(request.response.status_int, 400)
```


fredj:
rebased

PR #4497, opened 2019-01-14
Use QGIS server 3.4 by default

sbrunner:


arnaud-morvan:
I've make a short test with 3.4 in ADL and describeFeatureType is not filtered, this has effect on filters that show fields of first returned feature type instead of requested one.

sbrunner:
Thanks :-)

PR #4516, opened 2019-01-21
Fix SSL access from GDAL

pvalsecc:
For example when the profile is stored on S3...


geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile.mako
```diff
@@ -16,6 +16,7 @@ ENV NODE_PATH=/usr/lib/node_modules \
     TINYOWS_URL=http://tinyows/ \
     MAPSERVER_URL=http://mapserver/ \
     PRINT_URL=http://print:8080/print/
```
sbrunner: missing '\\'

PR #4518, opened 2019-01-21
Add path in routes for custom admin package.json

ger-benjamin:
So we can install and use admin's package.json with `npm install --global` in a custom project


ger-benjamin:
@sbrunner I've made a new change: move it to the 2nd last position because the last is always existing (and my link never used)

PR #4559, opened 2019-02-07
Remove sudo 

sbrunner:


docker/build/Dockerfile
```diff
@@ -26,7 +26,6 @@ RUN \
   touch /usr/local/lib/python3.6/dist-packages/c2c/__init__.py && \
   rm --recursive --force /tmp/* /var/tmp/* /root/.cache/*
 
-COPY sudoers /etc/
```
pvalsecc: I'm surprise not the see the deletion of this file...

sbrunner: Done :-)

PR #4581, opened 2019-02-12
Add chapter about geomapfish_qgisserver tests in server_side dev doc

arnaud-morvan:


sbrunner:
Should be rebased to fix Jenkins :-)

arnaud-morvan:
Ok, thanks for fixing master :-)

PR #4593, opened 2019-02-15
Fix themev2tov2 script

sbrunner:
Should fix https://jira.camptocamp.com/browse/GSGMF-779


rbovard:
Thanks @sbrunner, it's working.

sbrunner:
Thanks for the feedback :-)

PR #4626, opened 2019-02-20
Improve "Developing in Python" in developper doc

arnaud-morvan:
Note that I change:
```
GUNICORN_PARAMS="-b :80 --worker-class gthread --threads 1 --workers 1 --reload"
```
By:
```
GUNICORN_CMD_ARGS=--reload
```
As I found it really less intrusive.
Note that for now demo 2.4 use:
```
--bind=:8080 --worker-class=gthread --threads=10 --workers=1 --timeout=60 --max-requests=1000 --max-requests-jitter=100
```
And we should not change the port or I miss something.

### DIFF
--- a/doc/developer/debugging.rst
+++ b/doc/developer/debugging.rst
@@ -164,11 +164,24 @@ Just define an environment variable in the build:
 Developing in Python
 --------------------
 
-Add the following environment variable to the geoportal container:
+Create a development docker-compose.override.yml
+................................................
 
-``GUNICORN_PARAMS="-b :80 --worker-class gthread --threads 1 --workers 1 --reload"``
+Add a ``docker-compose.override.yml`` file with a ``geoportal`` service containing the following lines:
 
-TODO
+.. code:: yaml
+
+   ---
+
+   version: '2'
+
+   services:
+     geoportal:
+       volumes:
+         - ${PWD}/geoportal/<project>_geoportal:/app/<project>_geoportal
+       environment:
+         - GUNICORN_CMD_ARGS=--reload
+       # entrypoint: []  # do not eval templates when using webpack-dev-server
 
 You can also do a graceful restart of the running gunicorn:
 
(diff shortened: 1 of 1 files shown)


doc/developer/debugging.rst
```diff
@@ -164,11 +164,24 @@ Just define an environment variable in the build:
...
+
+   ---
+
+   version: '2'
 
-TODO
+   services:
+     geoportal:
+       volumes:
+         - ${PWD}/geoportal/<project>_geoportal:/app/aprona_geoportal
```
sbrunner:
```suggestion
         - ${PWD}/geoportal/<package>_geoportal:/app/<package>_geoportal
```

doc/developer/debugging.rst
```diff
@@ -164,11 +164,24 @@ Just define an environment variable in the build:
...
+       environment:
+         - GUNICORN_CMD_ARGS=--reload
```
sbrunner: Shouldn't we add `-b :8080` to be on the right port?

doc/developer/debugging.rst
```diff
@@ -164,11 +164,24 @@ Just define an environment variable in the build:
...
+       entrypoint: []  # do not eval templates when using webpack-dev-server
```
sbrunner: For me this will not work If we don't use webpack-dev-server, because we need the interpreted templates...

arnaud-morvan: We might put a # in front of this line

doc/developer/debugging.rst
arnaud-morvan: No, as I do not overwrite the c2cwsgiutils GUNICORN_PARAMS that is applied on the command line when running gunicorn.
This is the native gunicorn environment variable that allow to add parameters using environment variables, I found this method less intrusive with the base parameters given for production in GUNICORN_PARAMS.

sbrunner: +1, I miss that you use a different environment variable :-)

arnaud-morvan: Note that I do not apply "--threads 1 --workers 1" anymore, but this seems optional for me.

sbrunner: I get some issue with that, then it's good to remove that :-)

PR #4639, opened 2019-02-22
Don't create wsgi app for the scripts

arnaud-morvan:


geoportal/c2cgeoportal_geoportal/scripts/create_demo_theme.py
```diff
@@ -43,38 +45,41 @@ def main():
...
-    from c2cgeoportal_commons.models.main import Interface, OGCServer, Theme, LayerGroup, LayerWMS
+    config_uri = get_config_uri(options)
+    loader = get_config_loader(config_uri)
+    loader.setup_logging()
+    settings = loader.get_wsgi_app_settings()
 
-    session = DBSession()
+    generate_mappers()
+    engine = get_engine(settings)
+    session_factory = get_session_factory(engine)
```
sbrunner: All those are copied in "all" scripts, they should be in the `__init__` file.

In a function like `get_settings_[and_]session(...)`.

geoportal/c2cgeoportal_geoportal/scripts/manage_users.py
```diff
@@ -82,74 +84,77 @@ def main():
...
 
-    # Loading schema name from config and setting its value to the
-    # corresponding global variable from c2cgeoportal_geoportal
+    config_uri = get_config_uri(options)
+    loader = get_config_loader(config_uri)
+    loader.setup_logging()
+    settings = loader.get_wsgi_app_settings()
 
     # Ignores pyramid deprecation warnings
     warnings.simplefilter("ignore", DeprecationWarning)
```
sbrunner: This can be in all the scripts

arnaud-morvan: This will have no effect as we do not load the app anymore. see https://github.com/Pylons/pyramid/search?q=DeprecationWarning&unscoped_q=DeprecationWarning
This only appears in routes and views.

sbrunner: OK, thanks for investigation :-)

geoportal/c2cgeoportal_geoportal/scripts/themev1tov2.py
```diff
@@ -64,41 +67,43 @@ def main():
...
+        from c2cgeoportal_commons.models.main import (
+            OGCServer, Theme, LayerWMS, LayerWMTS, LayerV1, LayerGroup)
 
-        # List and create all distinct ogc_server
-        ogc_server(session, app.registry.settings)
+        if options.layers:
+            table_list = [LayerWMTS, LayerWMS, OGCServer]
+            for table in table_list:
+                print(("Emptying table {0!s}.".format(table.__table__)))
+                # must be done exactly this way otherwise the cascade config in the
```
sbrunner: Please start comments with capital :-)

arnaud-morvan: Forgot one during rebase :confused:

geoportal/c2cgeoportal_geoportal/scripts/create_demo_theme.py
```diff
@@ -43,38 +45,38 @@ def main():
...
     options = parser.parse_args()
 
-    get_app(options, parser)
-
-    from c2cgeoportal_commons.models import DBSession
-    from c2cgeoportal_commons.models.main import Interface, OGCServer, Theme, LayerGroup, LayerWMS
+    config_uri = get_config_uri(options)
+    loader = get_config_loader(config_uri)
+    loader.setup_logging()
+    settings = loader.get_wsgi_app_settings()
```
sbrunner: Those 4 lines still be in all the scripts...

arnaud-morvan: Yes, codacy also complains about thoses lines.
From my point of view, each line relates to something really different, so it make sense for me to keep them.
- First line is specific to c2cgeoportal for the parameters from scripts packages.
- Second line is a getter coming from pyramid.
- Third line is an action.
- Fourth line might be different in another script (for example get the app instead of settings).
It would be wreid to setup logging in a `get_settings` method.
IMHO we should keep thoses lines.

sbrunner: Yes, but it's relay not script specific, and finally in the script we just want to have the settings, and don't needs to know this complexity...

PR #4723, opened 2019-03-12
Add documentation for the timeAttribute

sbrunner:


doc/administrator/wmstime.rst
```diff
@@ -86,6 +86,11 @@ In mode ``range`` the layer will display data that is relative to the defined ra
 The ``disabled`` mode hides the time widget. No time parameter will be sent
 to the GetMap request in this case.
 
+Admin interface - query
+~~~~~~~~~~~~~~~~~~~~~~~
+
+To make the quierry working you also should add a ``timeAttribute`` metadata with the attribute that contains the time.
```
kalbermattenm:
```suggestion
By setting a ``timeAttribute`` metadata having as value the attribute used in the Mapfile as time item, you ensure that the WMS/WFS queries take into account the time filtering values and that no result is returned if an object is hidden by the time filter. 
```

sbrunner: Thanks :-)

PR #4729, opened 2019-03-12
Add docker-compose.override.yaml in Git ignore.

sbrunner:
Consider it as developer stuff that shouldn't be committed.


arnaud-morvan:
Note that we could propose a docker-compose.override.sample.yaml in scaffold.

sbrunner:
+1 @arnaud-morvan

PR #4734, opened 2019-03-13
Complete the changelog

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -13,6 +13,17 @@ Information
    a `.env.mako` file where we have the environment variables used in the Docker compose files,
    `COMPOSE_PROJECT_NAME` contains the default name of the project.
 
+Changes to apply
+----------------
+
+1. The ngeo modules import have change, the major chang that will consern you is that we dirrectly
```
jwkaltz: 1. The ngeo modules import has changed. The major change that will concern you is that we drrectly
export the module, therefore, in the angular module declaration, you should replace `<module>.module.name`
   by `<module>.name`.

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -13,6 +13,17 @@ Information
...
+   export the module, then in the angular module declaration you should replace `<module>.module.name`
+   by `<module>.name`.
+
+2. We migrate from Bootstrap 3 to 4 (https://getbootstrap.com/docs/4.0/migration/) that implyes
```
jwkaltz: 2. We have migrated from Bootstrap 3 to 4 (https://getbootstrap.com/docs/4.0/migration/). This implies
using Sass in place of Less. The major change is that the variables will be
   prefixed by '$' instead of '@' and we will use rem as recommended unit.

sbrunner:
Thanks @jwkaltz

PR #4739, opened 2019-03-13
Create a variable to skip translation process in enum config

kalbermattenm:
This PR was made for two purposes and should solve issues that we have with the layer enumeration configuration (YAML config file):
* It happens that we have None values or columns containing numbers (and defined as having to). This makes the lingua_extractor crash, thus I check for these exceptions and avoid to translate them (https://github.com/camptocamp/c2cgeoportal/compare/master...kalbermattenm:enum_translate?expand=1#diff-88967e2bdedfa9e803c46b29cd0a2e76R260)
* It happens that a DB column already contains the right translated strings, thus it is not useful to translate them. Moreover, for the SITN project and its filters, the translation process added more the 400 lines to the po file, because of that and all these translation additions are useless, we will not translate them (leaving the `msgstr` empty). I added a new config variable at the column level to specify if its content has to be translated (or not), by default it is translated and the variable (`translate`) is not required.

I am opened to any other suggestion in order to improve one or the other point.

Thanks for the review

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
@@ -257,7 +257,7 @@ class GeomapfishConfigExtractor(Extractor):  # pragma: no cover
             for fieldname in list(attributes.keys()):
                 values = self._enumerate_attributes_values(DBSessions, Layers, layerinfos, fieldname)
                 for value, in values:
-                    if value != "":
+                    if value is not None and isinstance(value, str) and value != "":
                         msgid = value
                         location = "/layers/{}/values/{}/{}".format(
                             layername,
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
```diff
@@ -288,6 +288,9 @@ class C:
     @staticmethod
     def _enumerate_attributes_values(dbsessions, layers, layerinfos, fieldname):
         dbname = layerinfos.get("dbsession", "dbsession")
+        translate = layerinfos.get("attributes").get(fieldname, {}).get("translate")
+        if translate is not None and translate is False:
+            return []
```
sbrunner:
```
        translate = layerinfos.get("attributes").get(fieldname, {}).get("translate", True)
        if not translate:
            return []
```

geoportal/c2cgeoportal_geoportal/lib/lingua_extractor.py
```diff
@@ -257,7 +257,7 @@ class C:
             for fieldname in list(attributes.keys()):
                 values = self._enumerate_attributes_values(DBSessions, Layers, layerinfos, fieldname)
                 for value, in values:
-                    if value != "":
+                    if value is not None and not isinstance(value, float) and not isinstance(value, int) and value != "":
```
sbrunner: Why not:?
```suggestion
                    if value is not None and isinstance(value, str) and value != "":
```

kalbermattenm: Yes, why not... ;-)

kalbermattenm:
I applied your review, but I just saw that Jenkins triggered an error... Lets see if he is happier this time...

kalbermattenm:
Thanks for the review & merge, @sbrunner

sbrunner:
:-)

PR #4765, opened 2019-03-19
Improve login

gberaudo:


gberaudo:
After a few restart of Jenkins it finally passed!...

PR #4785, opened 2019-03-22
Fix edit-qgis-project

sbrunner:
Shouldn't we add or be able to add the environment variable:
    AWS_DEFAULT_REGION: ch-dk-2
    AWS_S3_ENDPOINT: sos-ch-dk-2.exo.io


sbrunner:
And what about the environment variables?

PR #4790, opened 2019-03-25
Generate geoportal/alembic.yaml using /build/c2ctemplate-cache.json

arnaud-morvan:
For now alembic.yaml is not regenerated when user development vars file is modified because it does not depends on VARS_FILES (plural version).
As /build/c2ctemplate-cache.json should have correct dependencies, using the cache as dependendy and data source should fix the problem and improve performance at the same time.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -760,9 +760,9 @@ push-docker:
 	docker push $(DOCKER_BASE)-config:$(DOCKER_TAG)
 	docker push $(DOCKER_BASE)-geoportal:$(DOCKER_TAG)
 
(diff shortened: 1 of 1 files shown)


sbrunner:
This change don't only change the dependency, but it also don't use the vars_alembic.yaml, then the `testdb/*.sql` rules will not work anymore...

arnaud-morvan:
I do not understand why do we need a custom vars file for alembic ?

arnaud-morvan:
In such case we might create an ALEMBIC_YAML_FILES (plural version) to have correct dependencies.

sbrunner:
It's used to generate the `testdb/*.sql` to create a testdb, but finally it's not used than nearly all of them will be removed in 2.4...

sbrunner:
For me this is safely done in 2.4 can can be closed :-)

arnaud-morvan:
Ok thanks for feedback.

PR #4813, opened 2019-04-02
Fix role_ids sql condition in mapserver

fredj:


doc/administrator/mapfile.rst
```diff
@@ -225,7 +225,7 @@ If we do not need to restrict on an area, we can use the following
         FROM
             <schema>.<table> AS geo
         WHERE (
-            %role_ids% IN (
+            ARRAY[%role_ids%] && ARRAY(
```
sbrunner: @kalbermattenm, @gnerred Is it OK like that or should we put more int the `mapfile_data_noarea_subselect`?
I like this solution because the () still be symmetric...

gnerred: Ok for me. I tested, it works.

kalbermattenm: I like it that way !

I saw that the documentation was updated as well, which is excellent !

Wouldn't it be good to put something about it in the changelog file as well ? Because, this is surely something you will not be able to automatize...

sbrunner: @fredj can you add something in the change log?

PR #4817, opened 2019-04-03
Support passing values for elements in routes in interfaces configura…

arnaud-morvan:
…tions
As I cannot use dummy ``route_url`` as constructed in _request for my tests, I've used testConfig coming from pyramid testing to register routes. This might be generalized to other tests for consistency.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/dynamic.py
+++ b/geoportal/c2cgeoportal_geoportal/views/dynamic.py
@@ -90,7 +90,10 @@ class DynamicView:
             params.update(config.get('params', {}))
             for name, dyn in config.get('dynamic_params', {}).items():
                 params[name] = dynamic[dyn]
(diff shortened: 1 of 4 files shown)


geoportal/c2cgeoportal_geoportal/views/dynamic.py
```diff
@@ -90,7 +90,10 @@ def dynamic(self):
             params.update(config.get('params', {}))
             for name, dyn in config.get('dynamic_params', {}).items():
                 params[name] = dynamic[dyn]
-            constants[constant] = self.request.route_url(config['name'], _query=params)
+            constants[constant] = self.request.route_url(config['name'],
+                                                         *config.get('segments', []),
```
sbrunner: Why named it segments?
In the pyramid documentations it's named elements ...
https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_url

PR #4821, opened 2019-04-04
Run webpack-dev-server as nonroot

sbrunner:


arnaud-morvan:
Have you tried it, because I already tried this and www-data does not have read permission on node_module or something else.

sbrunner:
Effectively I see that, tomorrow...

PR #4822, opened 2019-04-04
Don't rename the file that concern the API

sbrunner:


arnaud-morvan:
Thanks, @sbrunner, but it does not works for me:
```
ls: cannot access 'geoportal/aprona_geoportal/static-ngeo/build/desktop.*.html': No such file or directory
ls: cannot access 'geoportal/aprona_geoportal/static-ngeo/build/mobile.*.html': No such file or directory
ls: cannot access 'geoportal/aprona_geoportal/static-ngeo/build/iframe_api.*.html': No such file or directory
```
I see desktop.html, mobile.html and iframe_api.html in build folder.

sbrunner:
Thanks @arnaud-morvan for the test, I updated the pull request and now it should work (tested on demo).

PR #4846, opened 2019-04-10
Upgrade vulnerable packages

sbrunner:
See: https://requires.io/github/camptocamp/c2cgeoportal/requirements/?branch=2.3
@arnaud-morvan does it look safe to upgrade colander?


arnaud-morvan:
I took a look at the change log. Only one thing took my attention: for the admin interface; we made a custom url validator on top on colander one, so this might be tested before merge.

arnaud-morvan:
Note that admin url validator is already tested by Jenkins with some urls (http, https, static and config).

sbrunner:
OK, excellent, thanks for the feedback :-)

PR #4859, opened 2019-04-14
[requires.io] dependency update on master branch

pvalsecc:


sbrunner:
@arnaud-morvan Is it safe to get those changes?

PR #4887, opened 2019-04-23
Fix security issue introdused by Transifex workaround

sbrunner:


sbrunner:
Not fixed in transifex side: https://github.com/transifex/transifex-client/issues/211

PR #4894, opened 2019-04-26
Do DescribeFeatureType query on internal server

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-932


kalbermattenm:
For what it is worth, I tested that live on our (future) prod server and this is working.

Thanks a lot @sbrunner!

PR #4939, opened 2019-05-15
[Backport 2.3] Allow static urls in url metadatas

arnaud-morvan:


commons/tests/validators_test.py
```diff
@@ -0,0 +1,17 @@
...
+from c2cgeoportal_commons.lib.validators import url
+
+
+class TestUrl():
+
+    @pytest.mark.parametrize("valid_url", [
+        'http://geomapfish.org',
+        'https://geomapfish.org/functionalities',
+        'geomapfish.org/functionalities',
+        'static://img/cadastre.jpeg',
```
rbovard: @arnaud-morvan Shouldn't you also test `static:///...` as described here https://github.com/camptocamp/c2cgeoportal/blob/2.3/doc/integrator/customise.rst#url

arnaud-morvan: Yes it would make sense. Note that the regex accept anything except whitespaces after "static://".

PR #4940, opened 2019-05-15
Fix the cache when we call the application from different URL

sbrunner:


rbovard:
Any chance to see this PR merged soon?

PR #4953, opened 2019-05-20
update mfp

jwkaltz:


sbrunner:
Why this?

jwkaltz:
> Why this?

Current motivation is that 3.10.7 has a bug, which (so far) is visible only pyramid_oereb users. So all GeoMapFish users which also use pyramid_oereb will have a problem when updating to gmf 2.3/2.4

And in general, it seems to me GeoMapFish should always use the latest bug-fix version of MapFish Print branches.

sbrunner:
I miss that you only update the patch version -> looks good to merge :-)

jwkaltz:
Unfortunately, the CircleCI fails (it does not seem to be setup for 2.3?)

fredj:
> Unfortunately, the CircleCI fails (it does not seem to be setup for 2.3?)

You can ignore CircleCI and merge

PR #4957, opened 2019-05-21
By able to eayier build the project throw a proxy

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/Dockerfile
```diff
@@ -1,11 +1,5 @@
-FROM debian:stretch
-LABEL maintainer Camptocamp "info@camptocamp.com"
-
-RUN \
-  apt-get update && \
-  apt-get install --assume-yes --no-install-recommends gettext-base python3 && \
-  apt-get clean && \
-  rm --recursive --force /var/lib/apt/lists/*
+ARG VERSION
```
fredj: is it working to not have the `FROM` on the first line ?

sbrunner: Yes :-)

arnaud-morvan: Yes, but only with a recent version of docker

sbrunner: It's released tow year ago... (from  17.05)

PR #4964, opened 2019-05-27
Build c2cgeoportal using Docker multistage build

sbrunner:
=> Remove the docker-run and docker-compose-run, move to Circle CI
Update the documentation
Update C2cupgrade messages
Update Docker containers diagrams
Reactivate call to updatepo
Reactivate clean DockerHub tags

I steel fave to fix the deployments


Dockerfile
```diff
@@ -0,0 +1,179 @@
+FROM camptocamp/c2cwsgiutils:2 AS base
```
pvalsecc: Can't you upgrade to v3?

Dockerfile
```diff
@@ -0,0 +1,179 @@
...
+  typedoc typescript \
+  --src=/usr/lib/node_modules/ngeo/package.json --src=package.json --dst=npm-packages
+
+COPY admin/package.json /app/c2cgeoportal/admin/
+WORKDIR /app/c2cgeoportal/admin
+RUN \
+  npm --no-optional --no-package-lock install && \
+  npm cache clear --force && \
+  rm -rf /tmp/*
+
```
pvalsecc: Missing splitting comment.

Dockerfile
sbrunner: +1 in another pull request :-)

Dockerfile
sbrunner: Added, thanks :-)

PR #5014, opened 2019-06-12
Update the documentation removes all docker-run

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-982


sbrunner:
> Why do you empty the CONST_CHANGELOG ?

Because we regularly empty it and just after an LTS is a good moment to do that :-)

PR #5024, opened 2019-06-13
Upgrade some Docker dependency

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/docker-compose-lib.yaml_tmpl
```diff
@@ -47,15 +47,15 @@ services:
         lra.restrictionarea_id = ra.id AND lra.layer_id = la.id AND la.name = "
 
   print:
-    image: camptocamp/mapfish_print:3.16
+    image: camptocamp/mapfish_print:3.18
```
jwkaltz: If updating, why not update to the latest version?

sbrunner: Oups it's an error, fixed :-)

PR #5098, opened 2019-06-27
Move the apihelp to the static folder and move the static folder to t…

sbrunner:
…he config


geoportal/tests/test_get_url.py
```diff
@@ -54,7 +54,7 @@ def static_url(path, **kwargs):
         request.static_url = static_url
 
         self.assertEqual(get_url2("test", "static://pr:st/icon.png", request, set()), "http://server.org/pr:st/icon.png")
-        self.assertEqual(get_url2("test", "static:///icon.png", request, set()), "http://server.org/my_project_geoportal:static/icon.png")
+        self.assertEqual(get_url2("test", "static:///icon.png", request, set()), "http://server.org//etc/geomapfish/static/icon.png")
```
fredj:
```suggestion
        self.assertEqual(get_url2("test", "static:///icon.png", request, set()), "http://server.org/etc/geomapfish/static/icon.png")
```
double slash in the url ?

geoportal/tests/functional/test_themes_metadata.py
```diff
@@ -239,7 +239,7 @@ def static_url(url, **kwargs):
                             "time": "10:36:28",
                             "datetime": "2003-09-25T10:36:28",
                             "url1": "http://example.com/hi?a=b#c",
-                            "url2": "http://dummy.org/tests_geoportal:static/path/icon.png",
+                            "url2": "http://dummy.org//etc/geomapfish/static/path/icon.png",
```
fredj:
```suggestion
                            "url2": "http://dummy.org/etc/geomapfish/static/path/icon.png",
```

geoportal/tests/test_get_url.py
sbrunner: It's just an artifact of the tests...

sbrunner: It should be like that or also changing the get static URL method ...

PR #5123, opened 2019-07-02
Strip python libs

sbrunner:


Dockerfile
```diff
@@ -9,14 +9,16 @@ ENV \
...
   apt-get clean && \
   rm --recursive --force /var/lib/apt/lists/*
 
 COPY requirements.txt /tmp/
 RUN \
   python3 -m pip install --disable-pip-version-check --no-cache-dir --requirement=/tmp/requirements.txt && \
-  rm --recursive --force /tmp/* /var/tmp/* /root/.cache/*
+  rm --recursive --force /tmp/* /var/tmp/* /root/.cache/* && \
+  strip --strip-all $(file $(find /usr/local/lib/python3.7/dist-packages -name *.so*) | \
+  grep "not stripped" | sed -ne 's%\(/usr/local/lib/python3.7/dist-packages/.*\):.*%\1%p')
```
gberaudo: @sbrunner, what is the point of stripping these? What is the procedure if you need to get a relevant stack tract / debug something in there?

PR #5124, opened 2019-07-02
Be able to configure the cookie same site

sbrunner:


geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -28,12 +28,15 @@
...
-from pyramid.authentication import AuthTktAuthenticationPolicy, \
-    BasicAuthAuthenticationPolicy
+import logging
+
+from pyramid.authentication import AuthTktAuthenticationPolicy, BasicAuthAuthenticationPolicy
 from pyramid_multiauth import MultiAuthenticationPolicy
 
 from c2cgeoportal_geoportal.resources import defaultgroupsfinder
 
+LOG = logging.getLogger(__name__)
```
pvalsecc: Not used...

geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -44,14 +47,17 @@ def create_authentication(settings):
...
+    if len(secret) < 64:
         raise Exception('"authtkt_secret should be at least 64 characters.'
                         'See https://docs.pylonsproject.org/projects/pyramid/en/latest/api/session.html')
 
     cookie_authentication_policy = AuthTktAuthenticationPolicy(
-        settings["authtkt_secret"],
+        secret,
         callback=defaultgroupsfinder,
         cookie_name=settings["authtkt_cookie_name"],
+        samesite=samesite is None if samesite == '' else samesite,
```
arnaud-morvan: samesite == '' => samesite is None == False
So it could be simplified to:
```
False if samesite == '' else samesite
```

sbrunner: Oups thanks :-)

sbrunner: => https://github.com/camptocamp/c2cgeoportal/pull/5127

arnaud-morvan: But False is not a correct value for samesite: 
```
samesite

    Default: 'Lax'. The 'samesite' option of the session cookie. Set the value to None to turn off the samesite option.
```
So it should be correct and really simpler with:
```
samesite=samesite or None
```

arnaud-morvan: Note: for boolean settings it would be better to use pyramid.settings.asbool

PR #5137, opened 2019-07-05
Add floors configuration

sbrunner:


arnaud-morvan:
I feel wreid to see those values in a CONST file in scaffold, here is custom config for demo_geomapfish.

arnaud-morvan:
Maybe having a list of 3 values here would do the job as default value for everyone.

sbrunner:
> I feel wreid to see those values in a CONST file in scaffold, here is custom config for demo_geomapfish.

In the past I do the choice to make it available for everyone, I don't know if it's good or not, If we want we can change it for 2.5...

sbrunner:
> Maybe having a list of 3 values here would do the job as default value for everyone.

+1 :-)

PR #5209, opened 2019-08-20
Fix options call

sbrunner:


sbrunner:
See also: https://sentry.camptocamp.com/geo/geomapfish-demo/issues/148911/?environment=prod-2-5

PR #5210, opened 2019-08-20
Fix favicon path

sbrunner:


sbrunner:
See also: https://sentry.camptocamp.com/geo/geomapfish-demo/issues/148905/?environment=prod-2-5

PR #5217, opened 2019-08-20
Fix the table title

sbrunner:


doc/integrator/database.rst
```diff
@@ -17,15 +17,14 @@ and ``production_static`` for production environment.
 For the data of the ``main`` schema, we recommend to have one schema for each version of the application.
 The following example shows how an update can be performed from a version ``2019`` to a version ``2020``:
 
-========================  =============  =============
-                           Integration    Production
-                           schema name    schema name
-========================  =============  =============
+========================  =============  ============
+Schema name                Integration    Production
```
jwkaltz: I think this is a bit confusing, it looks like Schema name = Initial state, Start a new version, ...
So I like the current table headers better.

sbrunner: => updated...

PR #5236, opened 2019-08-28
Do http request asynchronously 

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-971


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -905,6 +933,41 @@ def _wms_get_features_type(self, ogc_server_id, wfs_url):
...
+    def preload(self, errors):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        tasks = set()
+        for ogc_server in models.DBSession.query(main.OGCServer).all():
+            url_internal_wfs, _, _ = self.get_url_internal_wfs(ogc_server, errors)
+            tasks.add(self._async_wms_get_features_type(ogc_server.id, url_internal_wfs))
+            tasks.add(self._async_wms_getcap(ogc_server))
+
+        loop.run_until_complete(asyncio.gather(*tasks))
```
arnaud-morvan: Why not asyncio.run, python doc says:
Application developers should typically use the high-level asyncio functions, such as asyncio.run(), and should rarely need to reference the loop object or call its methods.

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -187,14 +200,21 @@ def _get_metadatas(self, item, errors):
         return metadatas
 
     def _wms_getcap(self, ogc_server):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        result = loop.run_until_complete(self._async_wms_getcap(ogc_server))
+        loop.close()
+        return result
```
arnaud-morvan: Why not asyncio.run, python doc says:
Application developers should typically use the high-level asyncio functions, such as asyncio.run(), and should rarely need to reference the loop object or call its methods.

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: It's the first time I use asyncio than it's possible that I do some little mistake, but here if I use `asyncio.run` in place of `loop.run_until_complete` I get:
`ValueError: a coroutine was expected, got <_GatheringFuture pending>`
...

sbrunner: If I have a look of the doc:
https://docs.python.org/3/library/asyncio-task.html#asyncio.gather
To use run I should write a new method, then I prefer like that :-)

arnaud-morvan: You should not do give the gather to asyncio.run but an async function. No need for a new method: 
```
async def preload(self):
    ...
    await asyncio.gather(*tasks)

def _themes(self):
    asyncio.run(self.preload())
```

geoportal/c2cgeoportal_geoportal/views/entry.py
arnaud-morvan: Suggested change: 
```
async def _wms_getcap(self, ogc_server):
    ...
    return wms

def _wms_layers(self, ogc_server):
    wms = asyncio.run(self._wms_getcap())
```

sbrunner: +1 :-)

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: Yes but it moves many things to the theme_ method :-/

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: but not in _wms_layers...

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -187,14 +200,21 @@ def _get_metadatas(self, item, errors):
         return metadatas
 
     def _wms_getcap(self, ogc_server):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        result = asyncio.run(self._async_wms_getcap(ogc_server))
+        loop.close()
+        return result
```
arnaud-morvan: It should work with:
```
def _wms_getcap(self, ogc_server):
    return asyncio.run(self._async_wms_getcap(ogc_server))
```
nothing else, asyncio.run create the loop for you.

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -865,6 +884,13 @@ def apihelp(self):
         return {}
 
     def _wms_get_features_type(self, ogc_server_id, wfs_url):
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        result = asyncio.run(self._async_wms_get_features_type(ogc_server_id, wfs_url))
+        loop.close()
+        return result
```
arnaud-morvan: Should work with:
```
def _wms_get_features_type(self, ogc_server_id, wfs_url):
    return asyncio.run(self._async_wms_get_features_type(ogc_server_id, wfs_url))
```

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -921,26 +981,15 @@ def themes(self):
 
         result = {}
         all_errors = set()
+        LOG.info("Start preload")
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        self.preload(all_errors)
+        loop.close()
+        LOG.info("End preload")
```
arnaud-morvan: should work with : 
```
LOG.info("Start preload")
asyncio.run(self.preload(all_errors))
LOG.info("End preload")
```

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -255,7 +270,9 @@ def _wms_getcap_cached(self, ogc_server, _):
             headers.pop("Host")
 
         try:
-            response = self.get_http_cached(url, headers)
+            response = await asyncio.get_event_loop().run_in_executor(
+                None, get_http_cached, self.http_options, url, headers
+            )
```
arnaud-morvan: What does this do exactly, is it not the same as : 
```
response = await get_http_cached(self.http_options, url, headers)
```

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -885,7 +911,9 @@ def _wms_get_features_type(self, ogc_server_id, wfs_url):
             headers.pop("Host")  # pragma nocover
 
         try:
-            response = self.get_http_cached(wfs_url, headers)
+            response = await asyncio.get_event_loop().run_in_executor(
+                None, get_http_cached, self.http_options, wfs_url, headers
+            )
```
arnaud-morvan: What does this do exactly, is it not the same as : 
```
response = await get_http_cached(self.http_options, url, headers)
```

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: No, we need a separate thread for get_http_cached...

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: same...

sbrunner:
Thanks @arnaud-morvan is it OK to merge?

arnaud-morvan:
I still does not understand why you need to do: 
```
response = await asyncio.get_event_loop().run_in_executor(
                None, get_http_cached, self.http_options, url, headers
            )
```
As far as I understand this execute in the current tread.
To run in a custom thread the Python doc says:
```
    # 2. Run in a custom thread pool:
    with concurrent.futures.ThreadPoolExecutor() as pool:
        result = await loop.run_in_executor(
            pool, blocking_io)
        print('custom thread pool', result)
```
So It seems that your code should do exactly the same as:
```
response = await get_http_cached(self.http_options, url, headers)
```
But let's merge if everything works fine.

sbrunner:
When we are in an `await asyncio.gather(...)` the task will not be parallelized If I don't do an `run_in_executor` because asyncio will execute everything in the same thread, and request will block the current thread...

PR #5262, opened 2019-09-04
Add view to get translations from Simple API after negociation

arnaud-morvan:
GSGMF-1103
Client part: https://github.com/camptocamp/ngeo/pull/5104

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/i18n.py
+++ b/geoportal/c2cgeoportal_geoportal/views/i18n.py
@@ -1,0 +1,22 @@
+from pyramid.httpexceptions import HTTPFound
+from pyramid.view import view_config
+
+from c2cgeoportal_geoportal.lib.cacheversion import get_cache_version
+from c2cgeoportal_geoportal.lib.caching import NO_CACHE, set_common_headers
+
+
+@view_config(route_name='localejson')
+def locale(request):
+    response = HTTPFound(
+        request.static_url(
+            '{package}_geoportal:static-ngeo/build/{lang}.json'.format(
+                package=request.registry.settings["package"],
+                lang=request.locale_name
+            ),
+            _query={
+                'cache': get_cache_version(),
+            }
+        )
+    )
+    set_common_headers(request, 'api', NO_CACHE, response=response)
+    return response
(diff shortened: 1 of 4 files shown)


geoportal/c2cgeoportal_geoportal/views/i18n.py
```diff
@@ -0,0 +1,22 @@
...
+
+from c2cgeoportal_geoportal.lib.cacheversion import get_cache_version
+
+
+@view_config(route_name='localejson')
+def locale(request):
+    return HTTPFound(
+        request.static_url(
+            '/etc/geomapfish/static/{lang}.json'.format(
+                package=request.registry.settings["package"],
```
sbrunner: package is not used...

geoportal/c2cgeoportal_geoportal/views/i18n.py
```diff
@@ -0,0 +1,22 @@
...
+                lang=request.locale_name,
+                _query={
+                    'cache': get_cache_version(),
+                }
+            )
+        ),
+        headers={
+            "Access-Control-Allow-Origin": "*",
```
sbrunner: Do we relay needs this header?

arnaud-morvan: For the API it seems we need it, as it will be requested from CMS urls.

I cannot use `c2cgeoportal_geoportal.caching.set_common_headers` directly with which takes a response object as parameter with an HTTPFound exception.

Ideally CORS headers should threated in a tween or view decorator, not with an imperative method as it is done with `set_common_headers`.

We can also create a response object with info.registry.queryAdapterOrSelf(result, IResponse) as it is done by pyramid : https://github.com/Pylons/pyramid/blob/3af1883bcd617d74eb8c9b134b5ac830f8cdd2a9/src/pyramid/viewderivers.py#L408

We may also have a look at https://cornice.readthedocs.io/en/latest/index.html

sbrunner: I agree with all of that but for the final file, do we also need that for the HTTPFound responce? Do you test it without this header?
When I don some search about that I found: https://stackoverflow.com/questions/18539403/chrome-cancels-cors-xhr-upon-http-302-redirect, and it seems that we can't have a redirect on an OPTION request...
For me we should do a test to be sure that it's working correctly ...

arnaud-morvan: I alreaddy tested it from ngeo make serve-api from another port than demo_geomapfish, and it does not works without CORS header.

sbrunner: OK thanks :-)

arnaud-morvan:
@sbrunner : I missed that HTTPFound effectively extends Response class.
Now locale view is consistent with other GMF XHR views.
Do you want a last look ?

PR #5274, opened 2019-09-06
Removes unneded dependency on commons

sbrunner:


commons/c2cgeoportal_commons/models/main.py
```diff
@@ -43,9 +43,21 @@
...
+    from c2cgeoform.ext.deform_ext import RelationSelect2Widget
+except ModuleNotFoundError:
+    drop = None
+    class GenericClass:
+        def __init__(self, *args, **kwargs):
+            pass
+    HiddenWidget = GenericClass
+    SelectWidget = GenericClass
+    TextAreaWidget = GenericClass
+    CollenderGeometry = GenericClass
```
arnaud-morvan: ColanderGeometry

sbrunner: Thanks :-)

PR #5275, opened 2019-09-06
Add some more rules in the makefile

sbrunner:


Makefile
```diff
@@ -34,6 +34,12 @@ build-runner:
 	docker build --target=runner --tag=camptocamp/geomapfish:$(DOCKER_TAG) \
 		--build-arg=MAJOR_VERSION=$(MAJOR_VERSION) --build-arg=VERSION=$(VERSION) .
 
+QGIS_VERSION=latest
```
arnaud-morvan: QGIS_VERSION ?= latest
or I miss something

sbrunner: :-/ Your right :-)

PR #5276, opened 2019-09-06
Don't use some sql files to create the database

sbrunner:


sbrunner:
> So docker/test-db/12-alembic.sql and docker/test-db/13-alembic-static.sql were missing dependencies on related alembic migrations.

I don't understand what you want to say ...

arnaud-morvan:
Apparently docker/test-db/12-alembic.sql and docker/test-db/13-alembic-static.sql were generated from alembic migrations (with option --sql) without dependencies in makefile on alembic migration scripts, this is probably why I got error last week, and asked you to use alembic directly.

PR #5271, opened 2019-09-06
No simple auth

sbrunner:
It's incoherent to have tow factor authentication and steel have the simple auth enable.
So we should be able to disable the simple auth, and by the way set it to disable by default.

PR #5281, opened 2019-09-09
Add extra information in the changelog

sbrunner:
fix: https://github.com/camptocamp/c2cgeoportal/pull/5271#discussion_r322129514


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -8,6 +8,8 @@ Information
 -----------
 
 1. From now the basic auth is disabled by default.
```
jwkaltz: Sentence is not correct. Suggestion: 
"Basic authentication is disabled by default from this version onward."

PR #5291, opened 2019-09-11
Add puppeteer docker image to run check-example.js

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-1090

Expected final usage in demo_geomapfish:
```
docker-compose run --rm geomapfish-puppeteer node check-example.js http://geoportal:8080
```

### DIFF
--- a/docker/puppeteer/Dockerfile
+++ b/docker/puppeteer/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:8.11.3
+
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


sbrunner:
I don't understand Why you create a new image, with the main image many libs are missing, but is there another issue?

arnaud-morvan:
This is mainly for the log, as I used it for debugging check-example.js.
I do not know exactly where we want to put puppeteer, in geomapfish-tools ?

sbrunner:
OK, It's used in the checkers, then in the geoportal container, if you want I can take over :-)

PR #5292, opened 2019-09-11
Support layer short names and WMSUseLayerIDs

arnaud-morvan:
* Support layer short names and WMSUseLayerIDs
* Support roles == "ROOT" in get_area

### DIFF
--- a/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
+++ b/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
@@ -311,6 +311,9 @@ class OGCServerAccessControl(QgsAccessControlFilter):
         - Access area as WKT or None
         """
         roles = self.get_roles()
+        if roles == 'ROOT':
+            return Access.FULL, None
+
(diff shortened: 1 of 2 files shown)


docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -182,17 +188,25 @@ def get_layers(self):
                 return self.layers
 
             self.project = QgsProject.instance()
+            self.user_layer_ids, ok = self.project.readBoolEntry("WMSUseLayerIDs", "/", False)
```
sbrunner:
```suggestion
            self.use_layer_ids, ok = self.project.readBoolEntry("WMSUseLayerIDs", "/", False)
```

arnaud-morvan:
@sbrunner : I've added some tests and fixed a bug with root in get_area.
Could you have a look.

arnaud-morvan:
After rebase on 2.4, I made some tests with stpierre_geomapfish, no problem detected, so let's merge into 2.4.

PR #5300, opened 2019-09-17
Add missing config.localeUrl in scaffold api config

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-1103


arnaud-morvan:
Sorry, #5262 not merged

PR #5313, opened 2019-09-23
Do not try to generate WMTS capabilities at build

arnaud-morvan:
As it is not possible at build and it is lately done at runtime.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
@@ -49,7 +49,6 @@ endif
 
 # Use TileCloud chain
 TILECLOUD_CHAIN ?= TRUE
-TILECLOUD_CHAIN_LOCAL ?= FALSE
 
 TEMPLATE_EXCLUDE += node_modules \
 	print/templates \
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -103,9 +103,6 @@ endif
 
 TILEGENERATION_CONFIG_FILE ?= tilegeneration/config.yaml.tmpl
 DEFAULT_BUILD_RULES += $(TILEGENERATION_CONFIG_FILE)
-ifeq ($(TILECLOUD_CHAIN_LOCAL), TRUE)
```
sbrunner: line 52 and 419 should also be removed

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_Makefile_tmpl
```diff
@@ -103,9 +103,6 @@ endif
...
-DEFAULT_BUILD_RULES += $(WMTSCAPABILITIES_FILE)
```
sbrunner: Lines 97 and 420 should also be removed ...

arnaud-morvan:
Done thanks

PR #5315, opened 2019-09-23
Fix alembic multi-role script

arnaud-morvan:


sbrunner:
It looks to mix up some brand, I think that you want to merge on 2.4...

arnaud-morvan:
Yes it was for 2.4, thanks

PR #5360, opened 2019-10-04
Allow adding custom checkers

arnaud-morvan:
By adding reference to healcheck object in registry.


sbrunner:
Should be rebased :-)

PR #5371, opened 2019-10-07
Fix the cache invalidation in the tests

sbrunner:
In the tests we many times invalidate the cache but it hasn't any effect because we also replace the cache region...
Now we never replace the cache region but we reconfigure it.

I also replace the custom caches by a memory dogpile cache.

And finally I fix a publication of private layers name.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/__init__.py
@@ -359,13 +359,14 @@ def includeme(config: pyramid.config.Configurator):
 
     # dogpile.cache configuration
     if 'cache' in settings:
-        caching.init_region(settings['cache'])
-        from c2cgeoportal_commons.models import InvalidateCacheEvent
+        for name, cache_config in settings['cache'].items():
+            caching.init_region(cache_config, name)
+            from c2cgeoportal_commons.models import InvalidateCacheEvent
 
-        @zope.event.classhandler.handler(InvalidateCacheEvent)
-        def handle(event: InvalidateCacheEvent):  # pylint: disable=unused-variable
-            del event
-            caching.invalidate_region()
+            @zope.event.classhandler.handler(InvalidateCacheEvent)
+            def handle(event: InvalidateCacheEvent):  # pylint: disable=unused-variable
+                del event
+                caching.invalidate_region()
 
     # Register a tween to get back the cache buster path.
     if 'cache_path' not in config.get_settings():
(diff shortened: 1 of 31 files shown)


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -78,6 +83,20 @@ def get_http_cached(http_options, url, headers):
     return response
 
 
+@CACHE_REGION_OBJ.cache_on_arguments()
+def _build_web_map_service(no_cache, ogc_server_id):
+    # The content is named no_cache to don't be concidered in the cache key
```
arnaud-morvan: not to be considered

note that we should consider adding parameters to the decorator to ignore some parameters, it would be more understandable, example:
```
@CACHE_REGION_OBJ.cache_on_arguments(ignore_args=['xml'])
def _build_web_map_service(xml, ogc_server_id):
```

arnaud-morvan: and more flexible when we would have two custom parameters to ignore.

arnaud-morvan:
Also not that the key generator could use hash(tuple(args)) to construct the key. And for the sqlalchemy objects, instead of 'id' , we could use object.identity which returns and unique tuple for each objects.

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -359,13 +359,14 @@ def includeme(config: pyramid.config.Configurator):
...
-        @zope.event.classhandler.handler(InvalidateCacheEvent)
-        def handle(event: InvalidateCacheEvent):  # pylint: disable=unused-variable
-            del event
-            caching.invalidate_region()
+        for name, cache_config in settings['cache'].items():
+            caching.init_region(cache_config, name)
+            from c2cgeoportal_commons.models import InvalidateCacheEvent
+
+            @zope.event.classhandler.handler(InvalidateCacheEvent)
+            def handle(event: InvalidateCacheEvent):  # pylint: disable=unused-variable
```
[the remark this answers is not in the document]
sbrunner: Yes for pyflakes

geoportal/c2cgeoportal_geoportal/views/entry.py
sbrunner: But it's not our decorator, it's used by keygen_function who hasn't a know leg of the method, if you have an idea it's welcome :-)

sbrunner:
To be considered but in another pull request :-)

PR #5378, opened 2019-10-08
Don't expose private layers

sbrunner:


sbrunner:
@arnaud-morvan To be able to write the test I've done many changes:
- Add test
- More comprehensible tests run
- Don't delete the geodata.testpoint
- Create the table main.testpointtime in the init script and rename it to geodata.testpointtime
- Fix testpoint table schema

Then can you review it again?

PR #5380, opened 2019-10-08
Revive admin dev environment

arnaud-morvan:


.dockerignore
```diff
@@ -6,7 +6,6 @@
 !geoportal
 !admin
 !scripts/clone_schema.sql
-admin/Makefile
```
sbrunner: Why this? For me we should keep that... and add:
```
admin/.env
admin/Dockerfile
admin/Makefile
admin/README.md
admin/config.yaml
admin/docker-compose.yaml
```
Or inverse the filter (probable should be preferred)
Remove `!admin`
and add
```
!admin/package.json
!admin/c2cgeoportal_admin
!admin/requirements.txt
!admin/setup.cfg
!admin/setup.py
```

admin/Makefile
```diff
@@ -1,80 +1,36 @@
...
-	${VENV}/bin/pserve --reload development.ini
-
-${BUILD_DIR}/venv.timestamp:
-	# Create a Python virtual environment.
-	virtualenv -p python3 ${VENV}
-	# Upgrade packaging tools.
-	${VENV}/bin/pip install --upgrade pip==9.0.1 setuptools==36.5.0
-	touch $@
+serve: ## Run c2cgeoportal_admin in development mode
+serve:
```
sbrunner: Duplicate serve !

admin/Makefile
```diff
@@ -1,80 +1,36 @@
...
-	cd .. ; \
-	docker build -t $(DOCKER_BASE)_admin:$(DOCKER_TAG) --build-arg GIT_HASH=$(GIT_HASH) -f admin/Dockerfile .
-
-.PHONY: docker-serve
-docker-serve:
-	docker-compose up
+.PHONY: clean
+clean: ## Stop composition, remove container and images
+	docker-compose down --remove-orphans
+	docker rm $(DOCKER_BASE)_admin:$(DOCKER_TAG)
```
sbrunner: Shouldn't be rmi? 
Personally I don't clean the images on a clean, should we generalize that?

admin/Makefile
```diff
@@ -1,80 +1,36 @@
...
-
-${BUILD_DIR}/venv.timestamp:
-	# Create a Python virtual environment.
-	virtualenv -p python3 ${VENV}
-	# Upgrade packaging tools.
-	${VENV}/bin/pip install --upgrade pip==9.0.1 setuptools==36.5.0
-	touch $@
+serve: ## Run c2cgeoportal_admin in development mode
+serve:
+	docker-compose up admin
```
sbrunner: I prefer using `docker-compose up -d` but I don't know ?

llienher:
@arnaud-morvan I tested, and it is running fine in localhost.

admin/requirements-dev.txt
```diff
@@ -1,6 +1,4 @@
-flake8
-pytest
-pytest-cov
-pytest-selenium
-WebTest>=1.3.1
+# Used for development: please do not remove this file
+
 waitress
+pydevd
```
sbrunner: What's this package (pypi description is missing...)

admin/development.ini
```diff
@@ -1,36 +1,29 @@
...
 [app:main]
 use = egg:c2cgeoportal_admin
-
 pyramid.reload_templates = true
 pyramid.debug_authorization = false
 pyramid.debug_notfound = false
 pyramid.debug_routematch = false
 pyramid.default_locale_name = en
-pyramid.includes =
-    pyramid_debugtoolbar
```
sbrunner: You remove the pyramid_debugtoolbar?

admin/requirements-dev.txt
```diff
@@ -1,6 +1,4 @@
-flake8
-pytest
-pytest-cov
-pytest-selenium
-WebTest>=1.3.1
+# Used for development: please do not remove this file
```
sbrunner: This file is used in the setup.py, for me it should be removed from him.

admin/Makefile
```diff
@@ -1,80 +1,36 @@
...
+	@echo
+	@echo "Possible targets:"
+	@grep -Eh '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "    %-20s%s\n", $$1, $$2}'
 
-.PHONY: flake8
-flake8: ${BUILD_DIR}/requirements-dev.timestamp
-	${VENV}/bin/flake8 c2cgeoportal_admin acceptance_tests *.py
+.PHONY: preparedev
+preparedev: ## Run this once before serve
+	make -C .. build-tools build-test-db
```
sbrunner:
```suggestion
	make -C .. build-runner build-test-db
```
Or the base image in the Dockerfile is wrong...

.dockerignore
arnaud-morvan: Good catch, I got lost with such contradictions (not all but admin but not ...) the second option seems clearer.

admin/Makefile
arnaud-morvan: The advantage here is that you directly get the logs, and you can easily restart with fresh container (Ctrl + C, up arrow, enter). For development it seems to comfortable to me.

admin/Makefile
arnaud-morvan: We could consider, I've done it recently in getitfixed. I found useful such tool to get free disk space without removing all docker images of all projects.

sbrunner: ok, but it should be rmi instead of rm

admin/Makefile
sbrunner: ok

admin/requirements-dev.txt
arnaud-morvan: It is a debugger (client part of eclipse pydev debugger, also works with IntelliJ / PyCharm) as this image is only used for development, I think we could install debuggers by default to ease development.

sbrunner: And it's working out of the box with pserve?

sbrunner: Related card: https://jira.camptocamp.com/browse/GSGMF-848

sbrunner:
For me this can be meged :-)

PR #5407, opened 2019-10-17
Better layer cache

sbrunner:
Don't download po files from Transifex on rebuild

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,11 +87,16 @@ RUN \
 RUN \
   curl --output /opt/jasperreport.xsd http://jasperreports.sourceforge.net/xsd/jasperreport.xsd
 
-COPY . /opt/c2cgeoportal/
+COPY internal.mk vars.yaml /opt/c2cgeoportal/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -87,11 +87,16 @@ RUN \
...
-COPY . /opt/c2cgeoportal/
+COPY internal.mk vars.yaml /opt/c2cgeoportal/
+COPY .tx/ /opt/c2cgeoportal/.tx/
 WORKDIR /opt/c2cgeoportal/
-RUN mv bin/import-ngeo-apps bin/eval-templates bin/wait-db bin/transifex-init bin/run bin/run-git /usr/bin/
 ARG MAJOR_VERSION
 ENV MAJOR_VERSION=$MAJOR_VERSION
+RUN make --makefile=internal.mk dependencies
+
+COPY . /opt/c2cgeoportal/
```
arnaud-morvan: Does not this overwrite dependencies (delete them ?)

sbrunner: Good question ant it don't override (delete) them :-)

PR #5435, opened 2019-10-25
Add intranet security on restriction area

sbrunner:
- in the theme
- the mapserv proxy
- the tinyows proxy
- the layers (editing)

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/__init__.py
@@ -42,6 +42,7 @@ from zope.interface import implementer
 from c2cgeoportal_geoportal.lib.caching import get_region
 
 LOG = logging.getLogger(__name__)
+CACHE_REGION = get_region("std")
 CACHE_REGION_OBJ = get_region("obj")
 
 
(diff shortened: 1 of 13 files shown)


geoportal/c2cgeoportal_geoportal/lib/__init__.py
```diff
@@ -255,6 +256,33 @@ def _get_intranet_networks(request):
...
+    if is_intranet(request):
+        result.append(get_role_id("intranet"))
+    if request.user is not None:
+        result.append(get_role_id("registered"))
+        result.extend([r.id for r in request.user.roles])
+    return result
+
+
+def get_roles_name(request):
+    result = ["annonymous"]
```
arnaud-morvan: anonymous

geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -273,11 +274,9 @@ def _get_metadatas(self, item, errors):
...
 
         # Add headers for Geoserver
-        if ogc_server.auth == main.OGCSERVER_AUTH_GEOSERVER and self.request.user is not None:
-            headers["sec-username"] = self.request.user.username
-            headers["sec-roles"] = (
-                "" if self.request.user is None else ";".join([role.name for role in self.request.user.roles])
-            )
+        if ogc_server.auth == main.OGCSERVER_AUTH_GEOSERVER:
+            headers["sec-username"] = "root"
+            headers["sec-roles"] = "root"
```
arnaud-morvan: Why do we put "root" in headers for GeoServer ?

sbrunner: To do the same as for others about 10 lines higher.
To get all the layer for the internal usage in the theme building (in fact it's possible to do that without root access but it breaks the cache).

PR #5443, opened 2019-10-28
Be able to have logs in JSON

sbrunner:
To be used in OpenShift


commons/c2cgeoportal_commons/alembic/env.py
```diff
@@ -28,13 +28,18 @@
...
+from alembic import context
 from c2c.template.config import config
+from sqlalchemy import engine_from_config, pool
+
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
+os.environ.setdefault("LOG_HOST", "localhost")
+os.environ.setdefault("LOG_PORT", "0")
 fileConfig(context.config.config_file_name)
```
pvalsecc: Are you sure it works?
I would say you need to change this line into:
```
fileConfig(context.config.config_file_name, defaults=dict(os.environ))
```

sbrunner: Travis say no :-)

PR #5448, opened 2019-10-29
Fix cache

sbrunner:
We shouldn't cache on headers where we have the request ID...

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/entry.py
+++ b/geoportal/c2cgeoportal_geoportal/views/entry.py
@@ -69,6 +69,7 @@ from c2cgeoportal_geoportal.lib.caching import (
     NO_CACHE,
     PRIVATE_CACHE,
     PUBLIC_CACHE,
+    NoCacheDecorator,
     get_region,
     set_common_headers,
 )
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/views/entry.py
```diff
@@ -88,25 +89,26 @@
 CACHE_REGION_OBJ = get_region("obj")
 
 
+@NoCacheDecorator([0, 2])
 @CACHE_REGION.cache_on_arguments()
 def get_http_cached(http_options, url, headers):
-    response = requests.get(url, headers=headers, timeout=300, **http_options)
+    response = requests.get(url, headers=headers.get(), timeout=300, **http_options.get())
     LOG.info("Get url '%s' in %.1fs.", url, response.elapsed.total_seconds())
     return response
```
pvalsecc: Could be easier like that:
```
def get_http_cached(http_options, url, headers):
    @CACHE_REGION.cache_on_arguments()
    def do_it(url):
        response = requests.get(url, headers=headers.get(), timeout=300, **http_options.get())
        LOG.info("Get url '%s' in %.1fs.", url, response.elapsed.total_seconds())
        return response
    return do_it(url)
```

sbrunner: Done in alternative pull request :-)

PR #5453, opened 2019-10-30
Move login and theme methods from entry to a specific files

sbrunner:


geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -0,0 +1,340 @@
...
+                                }
+                            ),
+                            headers=(("Content-Type", "text/json"),),
+                        ),
+                    )
+                otp = self.request.POST.get("otp")
+                if otp is None:
+                    raise HTTPBadRequest("The second factor is missing.")
+                if not self._validate_2fa_totp(user, otp):
+                    LOG.info("The second factor is wrong for user '%s'.", user)
```
pvalsecc: Risks of displaying the hashed password of the user, there... I don't know.
To be sure, you can put `user.username` instead...

geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -0,0 +1,340 @@
...
+            user = models.DBSession.query(static.User).filter(static.User.username == login).one_or_none()
+            if user and not user.deactivated:
+                if "consecutive_failed" not in user.tech_data:
+                    user.tech_data["consecutive_failed"] = "0"
+                user.tech_data["consecutive_failed"] = str(int(user.tech_data["consecutive_failed"]) + 1)
+                if int(user.tech_data["consecutive_failed"]) >= self.request.registry.settings.get(
+                    "authentication", {}
+                ).get("max_consecutive_failures", sys.maxsize):
+                    user.deactivated = True
+                    user.tech_data["consecutive_failed"] = "0"
```
pvalsecc: Easy DOS, here. Maybe you want to disable that for admin?

sbrunner: I think that we should know that someone try to get the admin password, and we will reset it directly in the database ...

geoportal/c2cgeoportal_geoportal/views/login.py
sbrunner: Yes, in another pull request :-)

PR #5464, opened 2019-11-01
Use the username instead the user object in the logs

sbrunner:


sbrunner:
I don't know, for me it's less usable, but finally I don't know.
But if we do that we should do it everywhere...

PR #5492, opened 2019-11-06
Simplify memory view

sbrunner:


geoportal/c2cgeoportal_geoportal/views/memory.py
```diff
@@ -29,9 +29,8 @@
 
 
 import logging
-from typing import Any, Dict, List
+from typing import Any, Dict
 
-from pympler import asizeof, muppy
```
pvalsecc: Is it a dependency you could remove?

sbrunner: +1 :-)

PR #5504, opened 2019-11-08
Faster and removes unneeded things int QGIS server

sbrunner:


pvalsecc:
I don't understand wh you do all that.

PR #5515, opened 2019-11-11
Don't use six

sbrunner:


fredj:
`six` should be removed from `requirements-dev.txt`

sbrunner:
@fredj no because as it's mentioned it's required by transifex...

fredj:
yes, it's because of https://github.com/transifex/transifex-client/issues/252

PR #5518, opened 2019-11-12
Gcc is required to install c2cwsgiutils

sbrunner:


docker/qgisserver/Dockerfile
```diff
@@ -2,6 +2,13 @@ ARG VERSION=latest
 FROM camptocamp/qgis-server:${VERSION}
 LABEL maintainer Camptocamp "info@camptocamp.com"
 
+RUN \
+  . /etc/os-release && \
+  apt-get update && \
+  apt-get install --assume-yes --no-install-recommends gcc && \
+  apt-get clean && \
+  rm --recursive --force /var/lib/apt/lists/*
```
pvalsecc: You should install it and uninstall it in the same RUN as you install c2cwsgiutils

sbrunner: Fixed :-)

PR #5545, opened 2019-11-19
Add missing volume arg in scaffolding commands

asaunier:
The master doc indicates that one can list the existing scaffolds typing:
```
docker run --rm --volume=$(pwd):/src \
    camptocamp/geomapfish-tools:<release|version> \
    pcreate -l
```
See https://camptocamp.github.io/c2cgeoportal/master/integrator/create_application.html#list-existing-scaffolds

On the other hand the scaffolding commands omit the ``--volume`` parameter:
```
docker run --rm -ti \
    camptocamp/geomapfish-tools:<release|version> \
    run $(id -u) $(id -g) /src \
    pcreate -s c2cgeoportal_create <project>
```
See https://camptocamp.github.io/c2cgeoportal/master/integrator/create_application.html#create-the-new-application
I think this prevents from making the project available in the local file system.

### DIFF
--- a/doc/integrator/create_application.rst.tmpl
+++ b/doc/integrator/create_application.rst.tmpl
@@ -70,7 +70,7 @@ To create the application, first apply the ``c2cgeoportal_create`` scaffold:
 
 .. prompt:: bash
 
-    docker run --rm -ti \
+    docker run --rm -ti --volume=$(pwd):/src \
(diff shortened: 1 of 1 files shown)


asaunier:
Thanks @sbrunner for reviewing => merging.

PR #5557, opened 2019-11-21
Start TLS before login on SMTP server

arnaud-morvan:
And do not send not encrypted credentials over network


sbrunner:
Yon can disable the failing rule, on 2.5 we do the inverse...

PR #5565, opened 2019-11-22
Faster rebuild on makefiles changes, ...

sbrunner:


sbrunner:
Thanks @arnaud-morvan :-)

PR #5583, opened 2019-11-28
Not working with tilecloud chain, not needed -> remove

sbrunner:


fredj:
is the `gunicorn_config.py` file still used elsewhere?

sbrunner:
No because he is needed in OpenShift :-)

PR #5593, opened 2019-12-06
Some memory usage optimisation

sbrunner:


geoportal/c2cgeoportal_geoportal/views/theme.py
```diff
@@ -221,6 +218,9 @@ def build_web_map_service(ogc_server_id):
                     "children": [l.name for l in wms_layer.layers],
                 }
 
+            del wms
+            gc.collect(2)
```
pvalsecc: Do you really need `gc.collect`? And why only the second pool?

geoportal/c2cgeoportal_geoportal/views/theme.py
```diff
@@ -914,14 +911,16 @@ def get_url_internal_wfs(self, ogc_server, errors):
...
 
         await asyncio.gather(*tasks)
 
-    def _get_features_attributes(self, ogc_server_id, url_internal_wfs):
+    @CACHE_REGION.cache_on_arguments()
+    def _get_features_attributes(self, url_internal_wfs):
         all_errors = set()
-        feature_type, errors = asyncio.run(self._wms_get_features_type(ogc_server_id, url_internal_wfs))
+        feature_type, errors = asyncio.run(self._wms_get_features_type(url_internal_wfs))
+        gc.collect(2)
```
pvalsecc: Same question...

geoportal/c2cgeoportal_geoportal/views/theme.py
sbrunner: Yes to try to free memory behind capabilities parsing... => runs all generations

PR #5594, opened 2019-12-06
Remove one haproxy warning

sbrunner:


sbrunner:
But here we are not in SSL...

PR #5602, opened 2019-12-09
Add `--env` and `--upgrade` options to build

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
```diff
@@ -1,55 +1,76 @@
...
+        if result.code != 200:
+            print("ERROR:")
+            print(result.read())
+            exit(1)
+        f.write(result.read().decode())
+    os.chmod("upgrade", os.stat("upgrade").st_mode | stat.S_IXUSR)
+    try:
+        subprocess.check_call(["./upgrade", full_version])
+    except subprocess.CalledProcessError:
+        exit(1)
```
arnaud-morvan: Doesn't this mask the error ?

sbrunner: It masks the trace, but the real error message should be displayed by the upgrade command...

PR #5610, opened 2019-12-11
Fix the API locales

sbrunner:


geoportal/c2cgeoportal_geoportal/views/i18n.py
```diff
@@ -9,9 +9,7 @@
 def locale(request):
     response = HTTPFound(
         request.static_url(
-            "{package}_geoportal:static-ngeo/build/{lang}.json".format(
-                package=request.registry.settings["package"], lang=request.locale_name
-            ),
+            "/etc/geomapfish/static/{lang}.json".format(lang=request.locale_name),
```
arnaud-morvan: Are all files behind /etc public ?

arnaud-morvan: It was a visual effect, but the static view should point on /etc/geomapfish/static
sorry for the noise.

PR #5623, opened 2019-12-16
Use pwgen to create the default authtkt_secret

sbrunner:
Cleanup the scaffold variables throw the `project.yaml`


sbrunner:
Thanks :-)

PR #5625, opened 2019-12-17
Fix the filter panel

sbrunner:
Fix: https://jira.camptocamp.com/browse/GSGMF-1158

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/theme.py
+++ b/geoportal/c2cgeoportal_geoportal/views/theme.py
@@ -209,7 +209,7 @@ class Theme:
                     "maxResolutionHint": float("{:0.2f}".format(resolution[1])),
                 }
                 if hasattr(wms_layer, "queryable"):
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/views/theme.py
```diff
@@ -209,7 +209,7 @@ def build_web_map_service(ogc_server_id):
                     "maxResolutionHint": float("{:0.2f}".format(resolution[1])),
                 }
                 if hasattr(wms_layer, "queryable"):
-                    info["queryable"] = wms_layer.queryable
+                    info["queryable"] = wms_layer.queryable == 1 if hasattr(wms_layer, "queryable") else False
```
jwkaltz: The code
if hasattr(wms_layer, "queryable") 
is redundant because the same "if" is above the statement

sbrunner: Effectively, thanks :-)

PR #5630, opened 2019-12-19
[WIP] Include translations when installing from source

kdeininger:
When installing from source, the locale directory is missing. This should be fixed by adding `*.po` files in MANIFEST.in.


kdeininger:
Does not solve the problem.

PR #5631, opened 2019-12-19
Fix access control in QGSI server - no more working

sbrunner:
- Fix published image build
- Better logger call
- More robust initialization on failure
- More robust disable all layer on initialization failure
- Fix database URI
- Better debugging config

### DIFF
--- a/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
+++ b/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
@@ -13,6 +13,7 @@ Contact: info@camptocamp.com
 from enum import Enum
 import json
 import os
+import re
 import sys
 from threading import Lock
 import traceback
@@ -20,14 +21,13 @@ import traceback
 from c2c.template.config import config
 import c2cwsgiutils.broadcast
 import geoalchemy2
+from qgis.core import Qgis, QgsDataSourceUri, QgsLayerTreeGroup, QgsLayerTreeLayer, QgsMessageLog, QgsProject
+from qgis.server import QgsAccessControlFilter
 from shapely import ops
 import sqlalchemy
 from sqlalchemy.orm import configure_mappers, scoped_session, sessionmaker
 import yaml
 import zope.event.classhandler
-
-from qgis.core import QgsDataSourceUri, QgsLayerTreeGroup, QgsLayerTreeLayer, QgsMessageLog, QgsProject
-from qgis.server import QgsAccessControlFilter
 
 
 class GMFException(Exception):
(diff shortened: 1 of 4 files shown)


docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -45,14 +44,20 @@ def __init__(self, server_iface):
...
         try:
             config.init(os.environ.get("GEOMAPFISH_CONFIG", "/etc/qgisserver/geomapfish.yaml"))
 
-            c2cwsgiutils.broadcast.init(None)
+            c2cwsgiutils.broadcast.init()
 
             configure_mappers()
-            engine = sqlalchemy.create_engine(config.get("sqlalchemy_slave.url"))
+            QgsMessageLog.logMessage(
+                "Connect to the database: {}".format(config.get("sqlalchemy_slave.url")),
```
pvalsecc: The password is part of the URL => not a good idea to log as is

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -45,14 +44,20 @@ def __init__(self, server_iface):
...
+                "GeoMapFishAccessControl",
+                level=Qgis.Info,
+            )
+            engine = sqlalchemy.create_engine(config["sqlalchemy_slave.url"])
```
pvalsecc: I recommend adding `pool_recycle=30` to every SQL engine. Now a day, idle connections are closed by the server after 120s.

pvalsecc: If you don't do that, you'll have errors!

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
sbrunner: +1, thanks :-)

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
sbrunner: done :-)

geoportal/c2cgeoportal_geoportal/scaffolds/create/qgisserver/geomapfish.yaml_tmpl
```diff
@@ -3,11 +3,18 @@ vars:
   schema: '{PGSCHEMA}'
   schema_static: '{PGSCHEMA_STATIC}'
   sqlalchemy.url: postgresql://{PGUSER}:{PGPASSWORD}@{PGHOST}:{PGPORT}/{PGDATABASE}
```
arnaud-morvan: It seems `sqlalchemy.url` is not used anymore

sbrunner: Probably yes but I like to keep it...

PR #5639, opened 2019-12-20
Add Python logger for QGIS server access control

sbrunner:


docker/qgisserver/geomapfish_qgisserver/__init__.py
```diff
@@ -9,17 +9,32 @@
...
 
     try:
-        from geomapfish_qgisserver.accesscontrol import GeoMapFishAccessControl
+        from .accesscontrol import GMFException, GeoMapFishAccessControl
 
         return GeoMapFishAccessControl(serverIface)
-    except GMFException as e:
-        QgsMessageLog.logMessage(str(e))
+    except GMFException:
+        LOG.error("".join(traceback.format_exception(*sys.exc_info())))
```
pvalsecc: This is very strange... why not?
```python
LOG.error("Cannot setup GeoMapFishAccessControl", exc_info=True)
```

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -414,82 +365,52 @@ def layerFilterSubsetString(self, layer):  # NOQA
...
-            )
+            LOG.debug("layerFilterSubsetString filter: %s", result)
             return result
         except Exception:
-            QgsMessageLog.logMessage(
-                "".join(traceback.format_exception(*sys.exc_info())),
-                "GeoMapFishAccessControl",
-                level=Qgis.Critical,
-            )
+            LOG.error("".join(traceback.format_exception(*sys.exc_info())),)
```
pvalsecc: Same for all the other usage of format_exception

docker/qgisserver/geomapfish_qgisserver/__init__.py
sbrunner: Quick conversion of QgsMessageLog :-) thanks

PR #5667, opened 2020-01-07
clarify editing prerequisite

jwkaltz:


sbrunner:
Should be back ported to master :-)

jwkaltz:
> Should be back ported to master :-)

Done in #5668

PR #5679, opened 2020-01-08
Add a precision about the API layers

sbrunner:


doc/integrator/api.rst
```diff
@@ -34,3 +34,9 @@ In the file ``geoportal/demo_geoportal/static-ngeo/api/index.js`` you can custom
    config.queryableLayers = [<layer a>, <layer b>, ...];
 
 The API help is in the folder ``geoportal/<package>_geoportal/static/apihelp/``.
+
+To be able to have a layer in the api, the layer should be in the themes::
```
jwkaltz: Should this read:
"In order for a layer to be accessible via API, the layer must be in a group"

(below we are talking about "group", not "theme")

sbrunner: in the themes (imply that the group should already be in a visible theme) or do you have a better proposition?

doc/integrator/api.rst
```diff
@@ -34,3 +34,10 @@ In the file ``geoportal/demo_geoportal/static-ngeo/api/index.js`` you can custom
    config.queryableLayers = [<layer a>, <layer b>, ...];
 
 The API help is in the folder ``geoportal/<package>_geoportal/static/apihelp/``.
+
+In order for a layer to be accessible via API, the layer must be in the themes::
+
+ * Add the layer
+ * Select the api interface
```
jwkaltz: API

doc/integrator/api.rst
```diff
@@ -34,3 +34,10 @@ In the file ``geoportal/demo_geoportal/static-ngeo/api/index.js`` you can custom
...
+ * Add the layer to a group (the group should already be visible by the api,
```
jwkaltz: Suggestion:

* Add the layer to a group which is visible in the API.

Note: a group is visible in the API if it is in a theme whose API configuration option is activated.

doc/integrator/api.rst
```diff
@@ -34,3 +34,10 @@ In the file ``geoportal/demo_geoportal/static-ngeo/api/index.js`` you can custom
    config.queryableLayers = [<layer a>, <layer b>, ...];
 
 The API help is in the folder ``geoportal/<package>_geoportal/static/apihelp/``.
+
+In order for a layer to be accessible via API, the layer must be in the themes::
```
jwkaltz: Suggestion:
, the layer must be in the themes::
->
, proceed as follows::

doc/integrator/api.rst
sbrunner: Adapted and updated, what do you think?

jwkaltz: I like this except that the "the" should be removed in 
"whose the 'api' interface"

sbrunner: OK, updated :-), thanks

PR #5681, opened 2020-01-08
Add qgis db project reloader

RemiDesgrange:


arnaud-morvan:
As we talked about with Stéphane at the GMF daily, in some GMF projects we may have multiple QGIS projects.
In such case, the project path is given using the request MAP parameter and the access control config file list available projects and corresponding OGC server name in admin.

RemiDesgrange:
@arnaud-morvan found a better solution based on `QgsProject.lastModified` method. We'll provide another PR since it will also work with project saved as file.

PR #5683, opened 2020-01-10
Upgrade c2cgeoform

arnaud-morvan:


admin/c2cgeoportal_admin/templates/layout.jinja2
```diff
@@ -32,16 +32,25 @@
...
+    </script>
+    {% endfor %}
+
     <!-- Deform JavaScript -->
     {% if deform_dependencies %}
-    <script type="text/javascript" src="{{request.static_url('deform:static/scripts/deform.js')}}" type="text/css" />
+    <script src="{{request.static_url('deform:static/scripts/deform.js')}}" type="text/css" />
     {% for js in deform_dependencies['js'] %}
-    <script type="text/javascript" src="{{request.static_url(js)}}"></script>
+    <script src="{{request.static_url(js)}}"></script>
```
sbrunner: Why this? is it deprecated?

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -67,6 +71,13 @@ def __init__(self, *args: Any, **kwargs: Any):
 _schema: str = config["schema"] or "main"
 _srid: int = cast(int, config["srid"]) or 3857
 
+# Set some default values for the admin interface
+_admin_config: Dict = config.get_config().get("admin_interface", {})
+_map_config: Dict = {**_admin_config.get("map", {}), **default_map_settings}
```
sbrunner: default_map_settings should be first
```
>>> {**{'dd': 2, 'ee': 3 }, **{'dd': 4, 'gg': 5}}
{'dd': 4, 'ee': 3, 'gg': 5}`
```

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -128,7 +130,17 @@ def add_admin_interface(config):
         )
         config.add_view(c2cgeoportal_geoportal.views.add_ending_slash, route_name="admin_add_ending_slash")
         config.add_route("admin_add_ending_slash", "/admin", request_method="GET")
-        config.include("c2cgeoportal_admin", route_prefix="/admin")
+        config.include("c2cgeoportal_admin")
```
sbrunner: This is no more working with get it fixed?

admin/c2cgeoportal_admin/templates/layout.jinja2
arnaud-morvan: It is the default value (see with @tonio )

commons/c2cgeoportal_commons/models/main.py
arnaud-morvan: good catch, thanks

admin/c2cgeoportal_admin/templates/layout.jinja2
tonio: https://developer.mozilla.org/fr/docs/Web/HTML/Element/script#attr-type

geoportal/c2cgeoportal_geoportal/__init__.py
arnaud-morvan: We made changes in c2cgeoform to support multi applications.
The routes now looks like : 
c2cgeoform_index: /{application}/{table}
So the admin app segment is not a route_prefix anymore, it is an option of the application registered in c2cgeoportal_admin.routes module.

sbrunner: OK, thanks :-)

sbrunner:
:-)

PR #5697, opened 2020-01-13
Base makefile

sbrunner:
One build base on Makefile will be like that:

I'm not completely convinced because:
- It add a dependency that can be an issue on Windows (In fact I don't know)
- The generation of the `.env` file can be worth...

What do you think?

### DIFF
--- a/doc/integrator/requirements.rst
+++ b/doc/integrator/requirements.rst
@@ -7,6 +7,7 @@ To install a GeoMapFish application, you need to have the following
 components installed on your system:
 
 * **Git**
+* **Makefile**
 * **Docker** >= 1.12 (>= 17 is recommended)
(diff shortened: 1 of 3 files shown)


arnaud-morvan:
I'm convinced more than you by this move.
In my projects, I do not know exactly what I will do in my project, but it should be something like adding targets in Makefile without changing ./build script:

```
build:
    ./build $(ENV_FILES)

build: docker-build-basemaps
```

Is c2cgeoportal 2.5 / docker still compatible with Windows ?

sbrunner:
That looks to be a good solution :-)

> Is c2cgeoportal 2.5 / docker still compatible with Windows ?

On Windows 10 desktop yes, On server it's currently not yet released...

PR #5702, opened 2020-01-15
Use env DOCKER_BASE in docker images names in ./build

arnaud-morvan:
To be sure we run the same images we build.


arnaud-morvan:
Superseeded by #5696

PR #5704, opened 2020-01-15
Ignore startup quote and spaces

sbrunner:
Fix: https://sentry.io/organizations/camptocamp/issues/1447931269/

### DIFF
--- a/docker/build/requirements.txt
+++ b/docker/build/requirements.txt
@@ -17,6 +17,7 @@ coverage==4.5.3  # Build coverage XML for Codacy
 dateutils==0.6.6  # geoportal pcreate
 deform==2.0.7  # commons, admin
 defusedxml==0.5.0  # geoportal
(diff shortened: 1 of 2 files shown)


docker/build/requirements.txt
```diff
@@ -17,6 +17,7 @@ coverage==4.5.3  # Build coverage XML for Codacy
 dateutils==0.6.6  # geoportal pcreate
 deform==2.0.7  # commons, admin
 defusedxml==0.5.0  # geoportal
+docutils==0.15.2  # doc
```
pvalsecc: In the stuff you modified in this PR, I don't see any usage for that module. Are you sure it is needed?

sbrunner: Yes, it's used in the documentation and the documentation can't be built with the new version 0.16 ⇒ CI break :-(

sbrunner:
UP :-)

PR #5705, opened 2020-01-15
Be able to do a ST_Union instead of ST_Collect

sbrunner:
Required when we have some restriction area that have an intersection


sbrunner:
It can be good, but I don't know where, should ve create a documentation page with all the used environment variables?

arnaud-morvan:
+1, maybe in another PR

PR #5725, opened 2020-01-23
Do not npm install in custom ngeo folder

arnaud-morvan:


arnaud-morvan:
What is the procedure today to port to master ? Do I have to create another PR ?

sbrunner:
Yes, when we get the GitHub actions (normally in beginning of February) it should be possible by adding a label :-)

PR #5740, opened 2020-01-27
GitHub actions

sbrunner:


.github/workflows/dependabot-auto-merge.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - unlocked
+
+jobs:
+  auto-merge:
+    name: Auto merge
+    runs-on: ubuntu-latest
```
fredj: probably not important but this one uses `ubuntu-latest` and the other `ubuntu-18.04`

PR #5751, opened 2020-01-31
Upgrade getitfixed (less intrusive on session)

arnaud-morvan:


sbrunner:
explicit_manager stil be here: https://github.com/camptocamp/getitfixed/blob/master/getitfixed/models/__init__.py#L59
Custom dbsession_factory looks strange for me, why do you needs that?
https://github.com/camptocamp/getitfixed/blob/master/getitfixed/models/__init__.py#L65

arnaud-morvan:
In getitfixed.includeme, I do not include getitfixed.models so this does not apply when getitfixed is used as a plugin (only in getitfixed.main or getitfixed.script.*).
Note that I do not know precisely why I need this, but it comes from the cookiecutter pyramid starter template and when I remove it completely, the tests suite fail.

sbrunner:
OK, thanks :-)

sbrunner:
Will be upgraded by: #5756

sbrunner:
This should rebased to get the changes in the `geoportal/c2cgeoportal_geoportal/__init__.py` file.

sbrunner:
Let's merge, the CI is OK :-)

PR #5773, opened 2020-02-04
[admin] Add missing field in dimension form template

arnaud-morvan:


arnaud-morvan:
Already done in 2.4, 2.5 and master.
Closing

PR #5778, opened 2020-02-04
Add missing ondelete="cascade" on table restricted_role_theme

arnaud-morvan:
Note that we might do the same with all n-n relationships


commons/c2cgeoportal_commons/alembic/main/be74e21d90c7_add_missing_cascade_on_restricted_role_.py
```diff
@@ -0,0 +1,100 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2019, Camptocamp SA
```
sbrunner: 2020 :-)

arnaud-morvan:
Superseeded by #6415

PR #5860, opened 2020-02-20
Install the application Python dependencies

sbrunner:


sbrunner:
Yes, it's effectively recommanded :-)

PR #5893, opened 2020-02-27
fix(qgisserver) edit tests are now called

julsbreakdown:
Rename the file so it is called by tests

### DIFF
--- a/docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
+++ b/docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
@@ -9,18 +9,13 @@ Contact: info@camptocamp.com
     option) any later version.
 """
 
-import os
-from unittest.mock import Mock, patch
-
 import pytest
 from geoalchemy2.shape import from_shape
 from qgis.core import QgsFeature, QgsGeometry, QgsProject, QgsVectorLayer
-from qgis.server import QgsServerInterface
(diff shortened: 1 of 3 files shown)


docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
```diff
@@ -19,8 +19,8 @@
 from shapely.geometry import LineString, box, shape
 
 from geomapfish_qgisserver.accesscontrol import Access, GeoMapFishAccessControl, OGCServerAccessControl
-from .accesscontrol_test import add_node_in_qgis_project, set_request_parameters
-
+from .accesscontrol_test import add_node_in_qgis_project, set_request_parameters, \
+    scoped_session, server_iface, qgs_access_control_filter
```
sbrunner: Is that right?
They are never used as python way, just in fixture...

julsbreakdown: Fixtures and their visibility are a bit odd in pytest. They don't require importing, but if you defined them in a test_*.py file, they'll only be available in that file.

sbrunner: Then it's required?

arnaud-morvan: It might sounds good to move them one level up in conftest.py

PR #5907, opened 2020-03-02
Getitfixed fixes

arnaud-morvan:


geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_vars.yaml_tmpl
```diff
@@ -30,6 +30,11 @@ vars:
     max_overflow: 25
     use_batch_mode: true
 
+  # Session backend
+  session:
+    type: 'ext:redis'
+    url: 'redis://{REDIS_HOST}:{REDIS_PORT}/{REDIS_DB}'
```
sbrunner: This is used only by getitfix, if not, what's exactly?

arnaud-morvan: This configure the user session for the whole application, but yes, I suppose user session is only used by getitfixed for now.

It is used in getitfixed when user post a form with an attached file. The file is stored in user session. If form validation fails, form is returned to the user with error messages next to form fields. When user post new form (with correction), if it pass validators, file is taken from session and stored in database.

arnaud-morvan: So we need to share the user session between webapp services using an external cache service like redis (for application scaling).

sbrunner: +1 Thanks for the responce :-)

PR #5911, opened 2020-03-03
Backport qgistests from 24

julsbreakdown:
Make the test work on 2.5
Also had to create an extra ADMIN user for tests

### DIFF
--- a/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
+++ b/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
@@ -271,7 +271,7 @@ class OGCServerAccessControl(QgsAccessControlFilter):
 
         parameters = self.serverInterface().requestHandler().parameterMap()
 
-        if parameters.get("USER_ID") == "0":
+        if parameters.get("USER_ID") == "0" or parameters.get("USER_ID") == "9999":
             return "ROOT"
 
         if "ROLE_IDS" not in parameters:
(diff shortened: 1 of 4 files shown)


docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
```diff
@@ -52,8 +56,12 @@ def test_data2(dbsession):
 
     user_no_access = User("user_no_access", roles=[role1])
     user_full_access = User("user_full_access", roles=[role2])
+    user_full_access.id = 2147483647  # root
```
arnaud-morvan: Here `user_full_access` seems not a root user, but a user with full access role.
Root user seems to be a role used for c2cgeoportal tests not for QGIS server plugin test.

docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
```diff
@@ -94,7 +104,8 @@ def test_data2(dbsession):
 
     dbsession.flush()
 
-    yield {"users": users, "roles": roles}
+    yield {"users": users, "roles": roles, "restriction_areas": restriction_areas,
+           "ogc_servers": ogc_servers}
```
arnaud-morvan: I do not like this form, I prefer all on same line or one line for each.

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -271,7 +271,7 @@ def get_roles(self):
 
         parameters = self.serverInterface().requestHandler().parameterMap()
 
-        if parameters.get("USER_ID") == "0":
+        if parameters.get("USER_ID") == "0" or parameters.get("USER_ID") == "2147483647":
```
arnaud-morvan: Seems not required as USER_ID=0 is passed in request parameters and should not require any role relationships.

PR #5971, opened 2020-03-11
Be able to preconfigure the project env files to avoid user error

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
```diff
@@ -67,7 +69,11 @@ def main():
             sys.exit(1)
         sys.exit(0)
 
-    env_files = args.env_file or load_env([".env"])["ENV_FILES"].split(" ")
+    project_env = yaml.load("project.yaml")["env"]
+    if len(args.env) != project_env["required_args"]:
+        print(project_env["help"])
+        sys.exit(1)
+    env_files = [e.format(*args.env) for e in project_env["files"]]
```
arnaud-morvan: As I understand, we may have in `project_env:

project_env:
```
required_args: 1
files:
    - env.default
    - env.project
    - env.{}
```
and `args.env` could be `['local']`
and we would get `env_files`:
```
[
    'env.default',
    'env.project',
    'env.local',
]
```

So we need to be less verbose on .build command ex: `./build local`

If I'm right, this need to be documented somewhere.

And the default value stored in .env (os.environ['ENV_FILES']) is not kept anymore.

sbrunner:
@jwkaltz can you have a look on the documentation part?

geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
sbrunner: Effectively done :-)

PR #5983, opened 2020-03-12
Typo finalise => finalize

sbrunner:


sbrunner:
As I see finalize is more correct and more common do you agree?

PR #6024, opened 2020-03-18
Require SSL to connect to the database

sbrunner:


sbrunner:
@danduk82 I just add a comment for that :-)

PR #6037, opened 2020-03-19
Add missing field in dimension form template

arnaud-morvan:
Already merged in 2.4


sbrunner:
Shouldn't be this regex:
https://github.com/camptocamp/c2cgeoportal/blob/master/.github/workflows/pr-title-lint.yaml#L24
Be?:
```
title-regex: "^(\\[.+\\] )?[A-Z0-9].*"
```

PR #6047, opened 2020-03-20
Use the slave connexion for mapserver and QGIS server

sbrunner:


danduk82:
@sbrunner , mmh what about WFS Write requests?

sbrunner:
@danduk82 on mapserver there is no WFS, on QGIS server you should use the connexion ``{{package}}_master`` :-)

danduk82:
ok thx

PR #6049, opened 2020-03-20
Better error if an env file does not exists

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
```diff
@@ -80,6 +81,10 @@ def main():
         sys.exit(1)
     env_files = [e.format(*args.env) for e in project_env["files"]]
     print("Use env files: {}".format(", ".join(env_files)))
+    for env_file in env_files:
+        if not os.path.exists:
+            print("Error: the env file '{}' does not exists.".format(env_file))
```
jwkaltz: does not exist

PR #6079, opened 2020-03-26
Be able to have a pullrequest title with a tag

sbrunner:


sbrunner:
@arnaud-morvan Does it look good to you?

sbrunner:
- The \\ is replaced by the yaml then the regexp is '^(\[[a-z]*\] )?[A-Z0-9].*'
- For the bachport test I can replace it by: '^(\[.*\] )?[A-Z0-9].*'
- "I'm really not sure we need such checks": Yes we need it because when we find huggly pull request title thy comes to the CHANGELOG (after the release)
- "... wait for the CI ...": this is wrong because it's another workflow.

PR #6108, opened 2020-03-31
 No rights if OGC server isn't found instead of error 

sbrunner:
See commits for details

### DIFF
--- a/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
+++ b/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
@@ -10,22 +10,24 @@ Contact: info@camptocamp.com
 """
 
 
-import c2cwsgiutils.broadcast
-import geoalchemy2
 import json
 import os
-import sqlalchemy
 import sys
 import traceback
+from enum import Enum
+from threading import Lock
+
+import geoalchemy2
+import sqlalchemy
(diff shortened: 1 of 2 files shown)


docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -101,9 +101,19 @@ def get_ogcserver_accesscontrol_config(self):
 
     def get_ogcserver_accesscontrol(self):
         if self.single:
+            if "MAP" in parameters:
+                raise GMFException("The map parameter should not be provided")
             return self.ogcserver_accesscontrol
         else:
             parameters = self.serverInterface().requestHandler().parameterMap()
+            if "MAP" not in parameters:
+                raise GMFException("The map parameter should be provided")
```
arnaud-morvan, line 110: What does QGIS return in such case ?
status 500 ?

arnaud-morvan: Note that you could return an empty OGCServerAccessControl to be consistent (no access).

sbrunner: It's possible but it is relay an error, We can't mix the tow comportment...

PR #6128, opened 2020-04-07
The functionnalities can't anymore be in the vars file

sbrunner:


doc/integrator/filter.rst.tmpl
```diff
@@ -3,57 +3,16 @@
...
-  for not-logged-in users), add the layer names as follows to your project vars file:
-
-.. code:: yaml
-
-    functionalities:
-        anonymous:
-            filterable_layers:
-            - my_layer_name_1
-            - my_layer_name_2
+You should add to your role the functionality names ``filterable_layers`` with the layer name as value.
```
arnaud-morvan, line 9: IMHO this is a little short.

Per default, layers are not filterable.
If you wish to provide filter functionality for some layers, proceed as follows:
- For each layer you want to be "filtrable" create a "functionality" record with "filtrable_layers" as name and the layer name as value.
- In any role (anonymous, registered, intranet or any other role) check the corresponding functionality records.

sbrunner: => https://github.com/camptocamp/c2cgeoportal/pull/6152

PR #6158, opened 2020-04-15
Start TLS before login on SMTP server

sbrunner:
And do not send not encrypted credentials over network

Backport #5557

### DIFF
--- a/commons/c2cgeoportal_commons/lib/email_.py
+++ b/commons/c2cgeoportal_commons/lib/email_.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2013-2019, Camptocamp SA
+# Copyright (c) 2013-2020, Camptocamp SA
 # All rights reserved.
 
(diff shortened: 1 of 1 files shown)


commons/c2cgeoportal_commons/lib/email_.py
```diff
@@ -81,10 +81,12 @@ def send_email(
...
     else:
         smtp = smtplib.SMTP(smtp_config["host"])
-    if smtp_config.get("user", False):
-        smtp.login(smtp_config["user"], smtp_config["password"])
     if smtp_config.get("starttls", False):
         smtp.starttls()
+    if smtp_config.get("user", False):
+        smtp.login(smtp_config["user"], smtp_config["password"])
+    if smtp_config.get("user", False):
+        smtp.login(smtp_config["user"], smtp_config["password"])
```
yjacolin, line 89: it seems there is the same group of lines twice.

sbrunner: Merge issue effectively, thanks :-)

PR #6188, opened 2020-04-21
Fix access control in multi project mode

sbrunner:


docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -182,42 +186,47 @@ class OGCServerAccessControl(QgsAccessControlFilter):
...
-            self.ogcserver = (
-                self.DBSession.query(OGCServer).filter(OGCServer.name == ogcserver_name).one_or_none()
-            )
-            if self.ogcserver is None:
-                LOG.warning("No OGC server found for '{}' => no rights".format(ogcserver_name))
+    def _init(self, ogcserver_name, map_file):
+        with self.lock:
+            try:
+                config_cache = QgsConfigCache.instance()
+                self.project = config_cache.project(map_file)
```
arnaud-morvan, line 215: This seems already done some lines above

arnaud-morvan:
@sbrunner: I suppose we should do the same in QGIS ProjectReloader (getting the project from QgsConfigCache intead of QgsProject.instance()).

arnaud-morvan:
PR for multi project is openned : https://github.com/camptocamp/QgisServerProjectReloader/pull/1/files

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
sbrunner: You comment an old commit...

sbrunner:
Yes :-)

docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
arnaud-morvan: Maybe I miss something, but I see the same lines at 208 and 209 and the page is refreshed. Maybe an error from github.

sbrunner: Now it's working...
Effectively I do that to remove a warning in the IDE, I need to initialize project in the `__init__` method and if I set it to None I get an error in the IDE because he considers that he can be None.
If you think that's durty I can remove it...

sbrunner:
I think

PR #6192, opened 2020-04-22
Be able to call MapServer without any roles_ids

sbrunner:


doc/administrator/mapfile.rst
```diff
@@ -221,6 +221,21 @@ The ``${MAPSERVER_DATA_SUBSELECT}`` is defined as follows:
...
+
+   VALIDATION
+     # For secured layers
+     "default_role_ids" ""
+     # Or
+     # "default_role_ids" "<id of the anonymus role>"
+     "role_ids" "^-?[0-9,]*$"
+   END
+
+If you set ``default_role_ids`` to an empty value then if you call mapserver dirrectly without any
```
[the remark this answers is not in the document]
sbrunner: In ``no access restrictions`` I remove the ``restrictions`` because for me it will say the inverse as we want...

sbrunner:
Thanks @jwkaltz :-)

doc/administrator/mapfile.rst
```diff
@@ -221,6 +221,23 @@ The ``${MAPSERVER_DATA_SUBSELECT}`` is defined as follows:
...
+   VALIDATION
+     # For secured layers
+     "default_role_ids" ""
+     # Or
+     # "default_role_ids" "<id of the anonymous role>"
+     "role_ids" "^-?[0-9,]*$"
+   END
+
+The ``default_role_ids`` define the restrictions when mapserver is called without any ``roles_ids``.
+In this situation, if you have set ``default_role_ids`` to an empty value for this layer, then there are
```
jwkaltz, line 237: then there are -> then there is

(But I have a question: does this make sense, is there a use case where you would want to define a layer without any access? Or do I understand the consequences wrongly?)

doc/administrator/mapfile.rst
```diff
@@ -262,22 +279,26 @@ The ``${MAPSERVER_DATA_NOAREA_SUBSELECT}`` is defined as follows:
...
+     # For secured layers
+     "default_role_ids" "-1"
+     # Or
+     # "default_role_ids" "<id of the anonymous role>"
+     "role_ids" "^-?[0-9,]*$"
+   END
 
-This variable is defined in the ``CONST_vars.yaml`` configuration file as follows:
+The ``default_role_ids`` define the restrictions when mapserver is called without any ``roles_ids``.
+In this situation, if you have set ``default_role_ids`` to ``-1`` for this layer, then there are
```
jwkaltz, line 295: there are -> there is

doc/administrator/mapfile.rst
sbrunner: You will have no right if you call MapServer directly without passing throw the proxy, without any ``role_ids``

PR #6318, opened 2020-05-06
Force the background to always be mixed

sbrunner:
Fix GSGMF-1270


danduk82:
@sbrunner  will this be in 2.5 backport?

sbrunner:
I want to test is in real some times before back port it...

PR #6557, opened 2020-06-22
Fix The 'isStretchWithOverflow' attribute is deprecated. Use the 'tex…

sbrunner:
…tAdjust' attribute instead

see also: https://github.com/mapfish/mapfish-print/pull/1360


arnaud-morvan:
Seems you forgot one in A4 Portrait : https://github.com/camptocamp/c2cgeoportal/blob/4f997b7106e1b0fcc77f9e40d4d52f5db436998f/geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/%2Bpackage%2B/A4_Portrait.jrxml#L73

sbrunner:
@arnaud-morvan Thanks :-)

PR #6565, opened 2020-06-23
[admin] Support undefined metadatas

arnaud-morvan:


arnaud-morvan:
I've changed the override by passing altered values to `super().serialize`. It is shorter and easier to understand.

PR #6566, opened 2020-06-23
Add check to avoid dependencies with security issue

sbrunner:


.github/workflows/audit.yaml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  schedule:
+    - cron: "30 2 * * *"
+
+jobs:
+  main:
+    runs-on: ubuntu-18.04
+    name: Audit
+    timeout-minutes: 5
+    if: "!startsWith(github.event.head_commit.message, '[skip ci] ')"
```
arnaud-morvan, line 14: Why no running on commit with [skip ci] in a cron job ?

.github/workflows/audit.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ matrix.branch }}
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - run: sudo python3 -m pip installsafety
```
arnaud-morvan, line 31:
```suggestion
      - run: sudo python3 -m pip install safety
```

PR #6605, opened 2020-06-29
Upgrade vulnerable dependencies

sbrunner:
See: https://github.com/camptocamp/c2cgeoportal/runs/815075996


admin/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "GeoMapFish administration interface",
   "author": "",
   "dependencies": {
-    "bootstrap": "3.4.0",
+    "bootstrap": "4.5.0",
```
sbrunner, line 7: @arnaud-morvan  Do you think that this change is OK?

arnaud-morvan: I'm pretty sure this will break near to everything on UI side.

PR #6606, opened 2020-06-29
Be able to ignore some CWE

sbrunner:


.github/workflows/audit.yaml
```diff
@@ -31,21 +31,25 @@ jobs:
       - run: |
           for file in $(find -name requirements.txt)
           do
-            (cd $(dirname ${file}) && safety check --full-report --file=requirements.txt)
+            echo Audit ${file}
+            (cd $(dirname ${file}) && safety check --full-report --ignore=$(cat safety-ignore|true) \
```
arnaud-morvan, line 35: Seems --ignore can be specified many times, but should not support comma separated list of ids.

arnaud-morvan: https://github.com/pyupio/safety#--ignore--i

sbrunner:
pyenv looks to be used my pipenv to install alernate python version

.github/workflows/audit.yaml
sbrunner: Yes, in the safety-ignore I should add comma separated list of ids :-)

PR #6735, opened 2020-07-17
Fix route conflict with getitfixed

arnaud-morvan:
Fix overlapping routes between admin and getifixed.


sbrunner:
I'm working on a fix for the CI...

sbrunner:
CI fixed with: https://github.com/camptocamp/c2cgeoportal/pull/6737

arnaud-morvan:
@sbrunner : Do you tell me that I should change the sha256 at top of Pipfile.lock that always create conflicts ?
I never index this change in my commits.
As far as I know there is no dependency cache based on this hash in our case (docker context).

PR #6823, opened 2020-08-13
The sentinel database should be 0

sbrunner:


sbrunner:
Merge to fix the demo...

PR #6880, opened 2020-08-28
Fix typo in manage-users script doc

asaunier:
Replace entrypoint ``manage-users`` by ``manage_users`` as the filename is actually ``manage_users.py`` and as it is the spelling actually used in the doc examples, see https://github.com/camptocamp/c2cgeoportal/blob/2.4/geoportal/c2cgeoportal_geoportal/scripts/manage_users.py#L42-L47

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scripts/manage_users.py
+++ b/geoportal/c2cgeoportal_geoportal/scripts/manage_users.py
@@ -39,12 +39,12 @@ def main():
     """
     Emergency user create and password reset script
     example, reset toto password to foobar:
-    ./docker-compose-run manage_users --password=foobar toto
+    ./docker-compose-run manage-users --password=foobar toto
     example, create user foo with password bar and role admin:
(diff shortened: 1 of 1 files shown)


sbrunner (changes requested):
It's the documentation who is wrong...

asaunier:
What's the problem with using "_" instead of "-" in the entrypoint name? This change is smaller than updating the doc and that's not consistent with the filename.

sbrunner:
The recommendation in Python is to have _ in filename, and the usage in bash is to have - in commands, and this change is a breaking change, not the doc...

asaunier:
@sbrunner Thanks! Updated

sbrunner (approved):

sbrunner:
Thanks :-)

asaunier:
The Jenkins check has been running for hours!
Feel free to merge the PR when it's OK if I have not done it yet :P

PR #6881, opened 2020-08-28
Add changelog and documentation about the ngeo constants

sbrunner:

PR #6892, opened 2020-08-31
Fix build on master

sbrunner:


sbrunner:
The last version of setuptools will break the build
see: https://github.com/camptocamp/c2cgeoportal/pull/6881/checks?check_run_id=1050902751

arnaud-morvan (approved):

PR #6901, opened 2020-09-01
Don't fail on unneeded apt repository fetch

sbrunner:
See e.-g.: https://github.com/camptocamp/c2cgeoportal/runs/1056864435?check_suite_focus=true


arnaud-morvan (approved):

sbrunner:
Thanks @arnaud-morvan :-)

PR #6944, opened 2020-09-07
Add this directory, as it needed when running c2cgeoportal in debug mode

jwkaltz:


sbrunner:
Good to start the EMPTY file with a dot to mark it as hidden :-)

jwkaltz:
> Good to start the EMPTY file with a dot to mark it as hidden :-)

Like this (see newest commit)?

sbrunner (approved):

PR #6953, opened 2020-09-08
Add target update-po and put translations in config

arnaud-morvan:
Note that SRC_FILES is empty by default and has to be configured in geoportal/Makefile in GMF projects if needed.

### DIFF
--- a/doc/integrator/internationalization.rst
+++ b/doc/integrator/internationalization.rst
@@ -41,8 +41,7 @@ To update your ``po`` files, you should proceed as follows.
 
 .. code:: bash
 
-    docker-compose exec tools update-po $(id --user) $(id --group) \
-        <lang_1> [<lang_2> [...]]
+    make update-po
 
 .. note::
 
--- a/doc/integrator/internationalization.rst
+++ b/doc/integrator/internationalization.rst
@@ -11,14 +11,13 @@ In the file ``<package>.mk``, define the supported languages with (default):
 
    LANGUAGES ?= en fr de
 
-In the file ``vars.yaml``, define the available and default locales:
+In the file ``vars.yaml``, define the default locale:
 
 .. code:: yaml
 
    vars:
         ...
         default_locale_name: fr
-        available_locale_names: [en, fr, de]
 
 In the file ``language_mapping``, define any desired locale variants, for example:
 
--- a/geoportal/c2cgeoportal_geoportal/lib/i18n.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/i18n.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2019-2020, Camptocamp SA
+# Copyright (c) 2020, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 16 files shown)


sbrunner (approved):

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile_tmpl
```diff
@@ -32,13 +32,11 @@ RUN true
 COPY --from=builder /usr/lib/node_modules/ngeo/dist/* /etc/static-ngeo/
 COPY --from=builder /etc/static-ngeo/* /etc/static-ngeo/
 COPY --from=builder /app/alembic.ini /app/alembic.yaml ./
-RUN chmod go+w /etc/static-ngeo/ \
-    /app/{{package}}_geoportal/locale/ \
-    /app/{{package}}_geoportal/locale/*/LC_MESSAGES/{{package}}_geoportal-client.po
+RUN chmod go+w /etc/static-ngeo/
```
sbrunner, line 35: Should  be added to the docker ignore...

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile_tmpl
```diff
@@ -32,13 +32,11 @@ RUN true
...
 
 RUN python3 -m pip install --disable-pip-version-check --no-cache-dir --editable=/app/ && \
     python3 -m compileall -q /usr/local/lib/python3.8 \
-        -x '/usr/local/lib/python3.8/dist-packages/(pydevd|ptvsd|pipenv)/' && \
+        -x '/usr/local/lib/python3.8/dist-packages/(pydevd|ptvsd|pipenv|.*pydev.*)/' && \
```
sbrunner, line 39: For curiosity why do need to do this change?

arnaud-morvan: Did not seen that we already have `pydevd` in the regex (can be removed as .*pydev.* also match).
Pydevd comes with some other modules : 

_pydev_bundle
_pydev_imps
_pydev_runfiles
_pydevd_bundle
_pydevd_frame_eval

And `compileall` fails on these folders.

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile_tmpl
arnaud-morvan: Good catch

arnaud-morvan:
@sbrunner : 
Concerning the command to run to update the translations catalogs.
I have in mind to recreate a `Makefile` with target `update-po` at scaffold root.
See : https://github.com/camptocamp/demo_geomapfish/commit/511eba0944ac43d9dfd701957427fb0072366d4d#diff-b67911656ef5d18c4ae36cb6741b7965R21

=> `make udpate-po`
Instead of having a long command in documentation, regularly changing with GMF versions.

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile_tmpl
sbrunner: We can keep that but it is curious why it's failing now...

sbrunner:
Why not simply modify the update-po command?

arnaud-morvan:
Because in c2cgeoportal we don't know the list of files with translations in a customized GMF project.

And note that we have the languages list is available in CONST_Makefile.

It is simpler for me to type `make udpate-po` than `docker-compose exec tools update-po $(id --user) $(id --group) en fr de`

geoportal/c2cgeoportal_geoportal/lib/i18n.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2013-2020, Camptocamp SA
```
sbrunner, line 3:
```suggestion
# Copyright (c) 2020, Camptocamp SA
```

geoportal/c2cgeoportal_geoportal/scripts/theme2fts.py
```diff
@@ -48,11 +49,11 @@ def main():
         description="Tool to fill the tsearch table (full-text search) from the theme information.",
     )
 
-    locale_path_1 = os.path.join("{package}_geoportal", "locale", "")
-    locale_path_2 = os.path.join("geoportal", locale_path_1)
-    locale_path = locale_path_2 if os.path.exists("geoportal") else locale_path_1
+    locale_path = "/etc/geomapfish/locale/"
```
sbrunner, line 52: Should be imported?

doc/integrator/internationalization.rst
```diff
@@ -41,8 +40,7 @@ To update your ``po`` files, you should proceed as follows.
 
 .. code:: bash
 
-    docker-compose exec tools update-po $(id --user) $(id --group) \
-        <lang_1> [<lang_2> [...]]
+    make update-po
```
sbrunner, line 43: Call to missing rule ...

sbrunner (approved):
Just some small issues :-)

sbrunner (approved):

arnaud-morvan:
@sbrunner 

I added some tests for method `i18n.available_locale_names` which detect available locales in locale folder.

Now I see that there is already "fr", "de", "en" and "it" languages created in that folder by the create template.
So it might be useful to be able to limit the available languages through vars.

In geoportal includeme I've put this: 
```
        if "available_locale_names" not in settings:
            settings["available_locale_names"] = available_locale_names()
```
But I do not really know how to handle existing and/or non existing keys for "available_locale_names" in CONFIG_VARS in Dockerfile.

We might discuss this by meet.

arnaud-morvan:
Finally the locales are added only at create time, and not at upgrade time because locale folder in included in default_project_file:
https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/c2cgeoportal_geoportal/scaffolds/update/%2Bdot%2Bupgrade.yaml_tmpl#L13
So we can safely delete a locale in locale folder and it won't reappear at next upgrade.

PR #6993, opened 2020-09-18
[admin] Create tree items directly from WMS GetCapabilities

arnaud-morvan:
GSGMF-1379

I've added a "Synchronize" button in OGC Server form that point to another form `/admin/ogc_servers/{id}/synchronize`.

This new form have a description of the synchronization process and a unique "Synchronize" button.
Once clicked on this button, the page reload with a green message that the synchronization was successful and the synchronization report.

See the result for the synchronization of test mapserver:
Screenshot_2020-10-14 GeoMapFish Administration Interface


commons/c2cgeoportal_commons/models/main.py
```diff
@@ -552,6 +556,18 @@ def __init__(
...
         return self.name or ""  # pragma: no cover
 
+    def url_description(self, request: Request) -> str:
+        errors: Set[str] = set()
+        url = get_url2(self.name, self.url, request, errors)
+        return url or "\n".join(errors)
+
+    def url_wfs_description(self, request: Request) -> Optional[str]:
+        if not self.url_wfs:
+            return None
```
sbrunner, line 566: Shouldn't be:?
```suggestion
            return self.url_description(request)
```

sbrunner (approved):
Is it useful to have all those commits shouldn't we stash them?

PR #7039, opened 2020-09-29
Fix: update-po doesn't use PGSCHEMA defined in .env file

sbrunner:
Fix GSGMF-1395

The PGSCHEMA should come from the Docker image for the geoportal service and from the .env for the tools service...


arnaud-morvan (commented):
1. Is is not possible to add "- PGSCHEMA" in geoportal (in env alias).
2. It would make sense, if we set a PGSCHEMA in .env to use it also for geoportal.

sbrunner:
1. no: https://github.com/yaml/yaml/issues/48
2. I don't like it because if the image isn't done correctly we will have an issue when we will push the project to OpenShift...

sbrunner:
This solution doesn't work, I propose a new solution...

arnaud-morvan (approved):

PR #7069, opened 2020-10-06
Fix pdfreport reading content

jwkaltz:


jwkaltz:
@sbrunner the current code for reading the response in pdfreport was changed here, I think this never worked:
https://github.com/camptocamp/c2cgeoportal/commit/85ee9386849c1879d2cf6e64d0ac113d493419ba

sbrunner (approved):

PR #7178, opened 2020-10-30
Add editing enumerations order metdata gsgmf gsgmf 1389

ger-benjamin:
For GSGMF-1389

Add layer metadata `editingEnumerationsOrder`
Expected metadata format: `enum1:columnx;enum2:columny;...` (whiteout space, I'm not sure If it's safe to trim them.)

I still have to test localy but looks reviewable


geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -538,6 +542,21 @@ def get_layer_class(layer, with_last_update_columns=False):
...
+def format_edit_enum_order_metadata(raw_metadata, layername):
+    """
+      Return the editingEnumerationsOrder metadata formatted as a dict.
+      The syntaxe of the metadata must be 'enum1:columnx,enum2:columny'
+    """
+    if not raw_metadata:
+        return None
+    enumerations_order_config = [items.split(":") for items in raw_metadata.split(",")]
+    if False in [len(e) == 2 for e in enumerations_order_config]:
+        LOG.warning(f'Bad synthax for metadata "editingEnumerationsOrder" on layer {layername}')
```
sbrunner, line 554:
```suggestion
        LOG.warning('Bad synthax for metadata "editingEnumerationsOrder" on layer %s', layername)
```

ger-benjamin: `make checks` says me the same :-)

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -538,6 +542,20 @@ def get_layer_class(layer, with_last_update_columns=False):
...
+def format_edit_enum_order_metadata(raw_metadata, layername):
+    """
+      Return the editingEnumerationsOrder metadata formatted as a dict.
+      The syntaxe of the metadata must be 'enum1:columnx,enum2:columny'
+    """
+    if not raw_metadata:
+        return None
+    enumerations_order_config = [items.split(":") for items in raw_metadata.split(",")]
+    if False in [len(e) == 2 for e in enumerations_order_config]:
+        LOG.warning('Bad synthax for metadata "editingEnumerationsOrder" on layer %s.', layername)
```
jwkaltz, line 554: syntax

geoportal/c2cgeoportal_geoportal/lib/xsd.py
```diff
@@ -81,7 +81,15 @@ def add_association_proxy_xsd(self, tb, column_property):
...
-        query = DBSession.query(getattr(target_cls, association_proxy.value_attr))
+        if (
+            hasattr(cls, "__enumerations_order__")
+            and association_proxy.value_attr in cls.__enumerations_order__
+        ):
+            query = DBSession.query(getattr(target_cls, association_proxy.value_attr)).order_by(
+                getattr(target_cls, cls.__enumerations_order__[association_proxy.value_attr])
+            )
+        else:
+            query = DBSession.query(getattr(target_cls, association_proxy.value_attr))
```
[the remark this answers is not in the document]
arnaud-morvan: Note that association_proxy.value_attr is not the name of the enum in main table, but name of the column to use as a value in foreign table.
Currently, in association proxy you can only access target (relation) and value_attr (name column).
But the name of the relation is not trivial to find for admins, it depends on multiple things.
And the interesting column name is only stored in the left side of the relation.primaryjoin condition.
If it cannot be easily accessed, I would recommend to add a property to the association proxy with the main table foreign key column name at construction time.

arnaud-morvan (commented):
This does not really do what you expect, value_attr cannot identify the main table enum.

arnaud-morvan (commented):
You may forgot to add your new metadata in available_metadatas in CONST_vars.yaml

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -538,6 +542,20 @@ def get_layer_class(layer, with_last_update_columns=False):
     return cls
 
 
+def format_edit_enum_order_metadata(raw_metadata, layername):
+    """
+      Return the editingEnumerationsOrder metadata formatted as a dict.
+      The syntaxe of the metadata must be 'enum1:columnx,enum2:columny'
```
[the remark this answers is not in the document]
ger-benjamin: I didn't know that json is an option. I use it now, that means I didn't need this function anymore.

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -538,6 +542,20 @@ def get_layer_class(layer, with_last_update_columns=False):
     return cls
 
 
+def format_edit_enum_order_metadata(raw_metadata, layername):
```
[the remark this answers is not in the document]
ger-benjamin: I've removed this function

geoportal/c2cgeoportal_geoportal/lib/xsd.py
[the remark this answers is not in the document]
arnaud-morvan: Personnaly I would vote for putting directly the order attribute in AssociationProxy ex:
```
    sort_column = None
    if (
        hasattr(cls, "__enumerations_order__")
        and col.name in cls.__enumerations_order__
        and 
    ):
        sort_column = getattr(child_cls, cls.__enumerations_order__[col.name]
    setattr(cls, proxy, _AssociationProxy(rel, "name", nullable=nullable, sort=sort_column))
```

geoportal/tests/functional/test_xsd.py
```diff
@@ -172,13 +172,18 @@ def test_add_association_proxy_xsd(self):
...
+            '<xsd:enumeration value="zad" />'
             '<xsd:enumeration value="bar" />'
             "</xsd:restriction>"
             "</xsd:simpleType>"
             "</xsd:element>",
             tostring(e).decode("utf-8"),
         )
 
+        # Test child 2 with an order by name (enumerations_order)
+        self.cls.__enumerations_order__ = {"name": "name"}
```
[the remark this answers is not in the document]
ger-benjamin: You're right, I'll correct that

geoportal/c2cgeoportal_geoportal/lib/xsd.py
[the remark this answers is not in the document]
ger-benjamin: Between both formula, this one looks clearer yes.

But I don't understand why we would need this line.
I've updated my code. It looks working without that.

I've tried to add the `order_by` on the `query` only in the `if` statement (your first comment) but I doesn't work (and I know that works when we do things like `query.all()`). Perhaps that doesn't work here because we use the query as a iterator ?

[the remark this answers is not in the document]
ger-benjamin: (Tests are also better now)

geoportal/c2cgeoportal_geoportal/lib/xsd.py
```diff
@@ -81,7 +83,12 @@ def add_association_proxy_xsd(self, tb, column_property):
         association_proxy = getattr(cls, proxy)
         relationship_property = class_mapper(cls).get_property(association_proxy.target)
         target_cls = relationship_property.argument
-        query = DBSession.query(getattr(target_cls, association_proxy.value_attr))
+        if hasattr(cls, "__enumerations_order__") and column.name in cls.__enumerations_order__:
```
arnaud-morvan, line 86: IMHO we should move `if hasattr(cls, "__enumerations_order__")` in `dbreflexion._add_association_proxy` and add an `order_by` property on association proxy. It would be more understandable.
It does not really make sense here to suppose `cls` can have a `__enumerations_order__` attribute.

arnaud-morvan:
I note that it is possible now to configure the enumeration sort column, but it not yet possible to configure the name column.
I think we should consider changing this to be able to configure everything we the same metadata ex:
enumm_attributes : {

arnaud-morvan:
@sbrunner : Any opinion on my last comment here ^

ger-benjamin:
> but it not yet possible to configure the name column.

The documentation says effectively that's not possible to use another column than `name` for the value. But  that's not part of this ticket and a change should be discussed with the user group. Or am I wrong ?

ger-benjamin:
After discussions, we decide to:
Add metadata `editingEnumeration` with the following aspect:
```
editingEnumerations: {
    "child1_id": {
        "value": "column_1",
        "order_by": "column_2",
    }
}
```
Value column is now configurable (default still to `name`)
order_by column is now configurable (new default to the "value column")

doc/administrator/administrate.rst.tmpl
```diff
@@ -51,6 +51,8 @@ of metadata in the ``GmfMetaData`` definition in
 Some metadata items are used by the ``layers`` views (for editing) on the server side:
 
 * ``editingAttributesOrder``: Specify the order of the editable attributes in the edit interface.
+* ``editingEnumerationsOrder``: Specify the column to use to order values of attributes with enumeration in
```
arnaud-morvan, line 54: This might be removed

doc/administrator/editing.rst
```diff
@@ -66,6 +76,8 @@ corresponding to this field is *Related Postgres table* in the admin interface.
     It is recommended to limit the editing to 30 editable layers per user.
 
 
+.. _administrator_editing_editable:
```
arnaud-morvan, line 79: Pretty sure you should move back this anchor before "Adding or making layers editable".

arnaud-morvan (approved):
Minor comments on doc, but this looks really good now.

doc/administrator/administrate.rst.tmpl
ger-benjamin: Thanks, it's updated.

doc/administrator/editing.rst
ger-benjamin: You're right

PR #7181, opened 2020-10-30
Add header.html file

sbrunner:


bin/import-ngeo-apps
```diff
@@ -136,6 +136,15 @@ def main():
                 ),
                 data,
             )
+            data = _sub(
+                re.escape("<ng-include src=\"'desktop/header.html'\"></ng-include>"),
+                "<ng-include src=\"'{entry_point}static/{version}/header.html'\"></ng-include>".format(
```
arnaud-morvan, line 141:
```suggestion
                "<ng-include src=\"'{entry_point}static/{cache_version}/header.html'\"></ng-include>".format(
```

bin/import-ngeo-apps
```diff
@@ -136,6 +136,15 @@ def main():
...
+                    entry_point="<%= htmlWebpackPlugin.options.vars.entry_point %>",
+                    version="<%= htmlWebpackPlugin.options.vars.version %>",
```
arnaud-morvan, line 143:
```suggestion
                    cache_version="<%= htmlWebpackPlugin.options.vars.cache_version %>",
```

arnaud-morvan (approved):
Minor comment, not sure, but it is `cache_version` for favicon also in static folder.

bin/import-ngeo-apps
sbrunner: +1 thanks :-)

PR #7198, opened 2020-11-03
Fix geo_table with environment variables

sbrunner:
Fix GSMORGES-8


ochriste (approved):

ochriste:
@sbrunner whats the issue with the build failing ?

PR #7206, opened 2020-11-04
Mapserver group filtering only work with mapserver 7 6

ochriste:


sbrunner:
Should be cleaned up...

ochriste:
that PR was an error, I thought I had closed it already, sorry

PR #7207, opened 2020-11-04
mapserver group filtering only work with mapserver 7.6 at least

ochriste:


sbrunner:
I'm not sure that a good idea to change the Mapserver version on a stabilization branch, maybe it's better to ass a note in the documentation or in the changelog...

sbrunner (changes requested):
I'm not sure that a good idea to change the Mapserver version on a stabilization branch, maybe it's better to as a note in the documentation or in the changelog...

PR #7208, opened 2020-11-04
Add GPG of Cécile

jwkaltz:


jwkaltz:
@sbrunner in fact, I am not sure it is useful to have our GPG keys in the template? If not, we should remove all of them. If yes, here is a PR to update them.

sbrunner:
In the newer version it's removed, I'm not for remving them in the old branches...

jwkaltz:
OK since it is no longer needed, let's leave it like this.

PR #7255, opened 2020-11-13
Allow to debug upgrade procedure

arnaud-morvan:
For now it is a pain to debug the upgrade procedure (or I miss something).
Here is a PR to be able to debug the upgrade procedure easily.
Note sure this is the better way to do it, but it is a good base for discussion.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
+++ b/geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
@@ -253,7 +253,9 @@ class C2cUpgradeTool:
             self.print_step(
                 step + 1,
                 message="Here is the output of 'git status'. Please make sure to commit all your "
-                "changes before going further. All uncommitted changes will be lost.",
+                "changes before going further. All uncommitted changes will be lost.\n"
+                "Note that for debugging purpose it is possible to pass directly to step 2 "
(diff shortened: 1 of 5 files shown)


geoportal/c2cgeoportal_geoportal/scripts/c2cupgrade.py
```diff
@@ -253,7 +253,9 @@ def step0(self, step):
             self.print_step(
                 step + 1,
                 message="Here is the output of 'git status'. Please make sure to commit all your "
-                "changes before going further. All uncommitted changes will be lost.",
+                "changes before going further. All uncommitted changes will be lost."
+                "\nNote that for debugging purpose it is possible to pass directly to step 2 "
+                "ex: ./upgrade --debug ../c2cgeoportal {version} 2".format(version=os.environ["VERSION"]),
```
sbrunner, line 258:
```suggestion
                "changes before going further. All uncommitted changes will be lost.\n"
                "Note that for debugging purpose it is possible to pass directly to step 2 "
                "e.-g.: ./upgrade --debug=../c2cgeoportal {version} 2".format(version=os.environ["VERSION"]),
```

sbrunner (approved):
Looks good to me, just some cosmetics proposal

PR #7275, opened 2020-11-16
Update the used QGIS version

sbrunner:


.github/workflows/qgis.yaml
```diff
@@ -15,12 +15,10 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          # Old LTR
-          - '3.4'
           # LTR
           - '3.10'
           # Last
-          - '3.14'
+          - '3.16'
```
arnaud-morvan, line 21: Note that 3.16 is also an LTR

arnaud-morvan (approved):
Thanks

.github/workflows/qgis.yaml
sbrunner: +1, thanks :-)

.github/workflows/qgis.yaml
```diff
@@ -16,11 +16,9 @@ jobs:
       matrix:
         version:
           # Old LTR
-          - '3.4'
-          # LTR
           - '3.10'
-          # Last
-          - '3.14'
+          # Last LTR
+          - '3.16'
```
ger-benjamin, line 21: The last LTR is still the 3.10 no ?
3.16 is only a standard release

arnaud-morvan: @ger-benjamin : see https://www.qgis.org/en/site/getinvolved/development/roadmap.html#release-schedule

ger-benjamin: Thanks you are right.
They are not up to date on their whole website. Sometimes (like on the download page or simply the flag at the upper-left corner) they still say the 3.10 is the latest LTR.

sbrunner: @ger-benjamin strictly your right that the version 3.16 is the last LTR only from 2021-02-19...

PR #7292, opened 2020-11-19
Add PGSCHEMA env variable for qgisserver

danduk82:
The accesscontrol plugin needs to have the PGSCHEMA env variable at runtime.


danduk82:
ping @vuilleumierc FYI

sbrunner (changes requested):
Not the good solution, I will work on a better solution...

danduk82:
ok then

PR #7298, opened 2020-11-20
All database configuration from the config container got the QGIS plugin

sbrunner:
For less confusion


arnaud-morvan (commented):
Not sure to fully understand.
Where do $ variables comes from ?
Is it just that replacements are done at build time (in config build) and not at run time ?
Note that c2c.template README does not contains any mention of t
his $ variables.

sbrunner:
They are interpreted in the config container as the runtime, with the config container environment variables :-)

arnaud-morvan (approved):
If I correctly understand, those variables are interpreted by envsubst in config entry point (eval-templates) (this is a .yaml.tmpl_tmpl file :-).

sbrunner:
+1, thanks :-)

PR #7301, opened 2020-11-21
Use DateTimeFormatter in print templates

arnaud-morvan:
DateTimeFormatter is more flexible than SimpleDateFormat for print templates
as we can chain setting pattern, timezone and locale in a textFieldExpression.


jwkaltz (approved):
LGTM

sbrunner (approved):

PR #7308, opened 2020-11-24
Add streetview options

RenataMuellerC2C:


geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_vars.yaml_tmpl
```diff
@@ -100,6 +100,9 @@ vars:
                 radius2: 0
                 angle: 0
         ngeoProfileOptions: {}
+        ngeoStreetviewOptions: 
+          viewer: 'google'
+          key: ''
```
sbrunner, line 105: The default viewer shouldn't me Mapillary @ybolognini  what do you think?
and the key should be in the `vars.yaml_tmpl` file.

geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/vars.yaml_tmpl
```diff
@@ -92,6 +92,8 @@ vars:
               color: '#0000A0'
             srtm:
               color: '#00A000'
+        ngeoStreetviewOptions:
+          key: 'd1dNaFk4aDVoVVlZd0dEZG95Wm84QTpkYmRkOGQyMWRkMThiM2E2'
```
sbrunner, line 96: Should only be in the demo...

sbrunner (approved):

PR #7323, opened 2020-11-26
Fix dynamic for API

sbrunner:


arnaud-morvan (commented):
I do not really understand what you fix here.
Is there something that really change ?

sbrunner:
With the old code we try to get the route named api for the api who doesn't exists, see: 
https://sentry.io/organizations/camptocamp/issues/2057061059/?project=1851011&query=&statsPeriod=14d

arnaud-morvan:
Yes, but I do not understand what changed, you still call request.route_url with interface_name.

sbrunner:
Yes for the standard interface (desktop, mobile, iframe_api), but not for the api (JavaScript)...

arnaud-morvan (approved):
Sorry, api does not extends defaults. Looks good.

PR #7325, opened 2020-11-26
Give the priority to the translations of the root package

arnaud-morvan:
See https://jira.camptocamp.com/browse/GEO-3834 for details.


sbrunner (approved):
Looks good :-)

PR #7339, opened 2020-11-30
Handling wms get capabilities failure

vuilleumierc:
In one project we have the problem that one WMS GetCapabilities returns an error and as a result the layer and layers from other servers are not loaded. This seems to solve the issue but maybe there is a better solution.


sbrunner (changes requested):
This line looks to be correct, then the `None` value is not handled correctly somewhere...

vuilleumierc:
I agree, but I could not find where - any pointers?

PR #7340, opened 2020-11-30
Add used by column

sbrunner:
To easier follows what is used by version.


arnaud-morvan (approved):
Looks good, might be useful if maintained up to date.

PR #7353, opened 2020-12-01
Do not add layer to theme if GetCapabilities fails

vuilleumierc:
New attempt at solving the issue of layers not loading following a WMS GetCapabilties failure

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -14,6 +14,8 @@ Information
 2. We change the secret name from `GITHUB_GOPASS_CI_TOKEN` to `GOPASS_CI_GITHUB_TOKEN` because we can't
    anymore create create secret started with `GITHUB_`.
 
+3. Layers which have any errors are not added to the theme anymore.
+
 Changes to apply
 ----------------
 
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -49,6 +49,14 @@ services:
       - "MAPSERVER_JOIN_WHERE=rra.role_id in (%role_ids%) AND rra.restrictionarea_id = ra.id AND
         lra.restrictionarea_id = ra.id AND lra.layer_id = la.id AND la.name = "
 
+  tilecloudchain:
+    image: camptocamp/tilecloud-chain:1.13
+    user: www-data
+    restart: unless-stopped
+    environment:
+      - TILEGENERATION_CONFIGFILE=/etc/tilegeneration/config.yaml
+    volumes:
+      - ./docker/test-tilecloudchain:/etc/tilegeneration
 
   tests:
     image: camptocamp/geomapfish-tools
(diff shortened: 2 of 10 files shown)


sbrunner:
What's the original error?
Stack trace?

vuilleumierc:
Hi @sbrunner 
[logs_geoportal.txt](https://github.com/camptocamp/c2cgeoportal/files/5627473/logs_geoportal.txt)
The geoportal log is attached. There is an error within the owslib function that does the GetCapabilities request. The error is caught by GMF. But the layer and subsequent layers in the layer tree are not loaded (seemingly without errors). In GMF 2.3 even the layer with the GetCapabilities problem would be loaded (the layer is fine, there is just an usual xml-formatting in the GetCapabilities). By looking around I saw that this dictionary "layer_theme" is not properly filled for this problematic layer, that might be the reason for the crash.

sbrunner:
What kind of result do you get if you replace:
https://github.com/camptocamp/c2cgeoportal/blob/18184891e108309f116158ed7aa56b6947bda931/geoportal/c2cgeoportal_geoportal/views/theme.py#L283-L286
with:
```python
splitted_content_type = response.headers.get("Content-Type", "").split(";")
if len(splitted_content_type) == 0 or splitted_content_type[0].strip() not in [
    "application/vnd.ogc.wms_xml",
    "text/xml",
]:
```
?

vuilleumierc:
@sbrunner I get the same thing

vuilleumierc:
This is the value of splitted_content_type : ['text/xml', ' charset=utf-8']

sbrunner:
If you have ['text/xml', ' charset=utf-8'] in splitted_content_type you shouldn't get the error:
WARNING! an error ''NoneType' object has no attribute 'strip'' occurred while trying to read the mapfile and recover the themes.
or we should know from where it comes from...

vuilleumierc:
I think the request works fine, but the xml parsing done by owslib WebMapService fails because of the mark-up <Format/>
Do you see what I mean? Could we maybe talk quickly on Meet? meet.google.com/mxn-wgyz-yfj Or when you have some time, just let me know

sbrunner (approved):
Can you merge your commit before merging? :-)

vuilleumierc:
Of course, sorry about that. But some CI tests failed, shouldn't we fix this?

sbrunner:
You should fix the tests because we test those kinds of cases...

vuilleumierc:
OK, I am having look right now, I'll ask again if I am stuck :-)

sbrunner (commented):
Looks good, now you should add a message in the CONST_Changelog file :-)

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -14,6 +14,8 @@ Information
 2. We change the secret name from `GITHUB_GOPASS_CI_TOKEN` to `GOPASS_CI_GITHUB_TOKEN` because we can't
    anymore create create secret started with `GITHUB_`.
 
+3. Layers for which the GetCapabilities request fails are not added to the theme anymore.
```
sbrunner, line 17:
```suggestion
3. Layers for which have any error are not added to the theme anymore.
```

sbrunner:
Looks good thanks :-)

PR #7358, opened 2020-12-02
Fix the API configuration

sbrunner:
The CHANGELOG modifications will be in a separate pull request...


arnaud-morvan (approved):
Looks good

PR #7412, opened 2020-12-11
Use c2cciutils - Publish to GitHub Container Registry

sbrunner:


docker/qgisserver/tests/functional/accesscontrol_allow_to_edit_test.py
```diff
@@ -8,13 +8,12 @@
 # the License, or (at your option) any later version.
 
 
-from geoalchemy2.shape import from_shape
 import pytest
+from geoalchemy2.shape import from_shape
+from geomapfish_qgisserver.accesscontrol import OGCServerAccessControl
```
arnaud-morvan, line 13: Might be considered as a known_first_party

arnaud-morvan (approved):
LGTM, minor comment about geomapfish_qgisserver as a known first party, but it might be addressed in another PR if this pass CI. I don't want to block merge of such huge PR.

PR #7417, opened 2020-12-11
Some status changes

sbrunner:
Add repository mutualized print logs
Reorganize(GeoMapFish - GitHub actions - OpenShift and other)
Add some more needed workflows
Display the step name only for QGIS build in c2cgeoportal (The versions are not any more needed, now it's checked by c2cciutils)


arnaud-morvan (commented):
Looks good

PR #7438, opened 2020-12-14
Fix the publish

sbrunner:


.github/workflows/main.yaml
```diff
@@ -160,10 +160,21 @@ jobs:
...
       - run: (cd ${HOME}/workspace/geomapfish/; ./build)
+
       - name: Publish
         run: |
           c2cciutils-publish
+          c2cciutils-publish --group=full
+        if: >
+          github.ref != format('refs/heads/{0}', env.MAIN_BRANCH)
+          && github.repository == 'camptocamp/c2cgeoportal'
+      - name: Publish
```
arnaud-morvan, line 171: Is it really usefull to have two steps with the same name and same conditions (if)

arnaud-morvan (approved):
Trust you on the details, but it seems two steps can be merged.

.github/workflows/main.yaml
sbrunner: The conditions are different on time we have `==` and the other `!=` :-)

PR #7512, opened 2020-12-21
Script docker-run: fix home directory setting

jwkaltz:
Fix home directory setting in docker-run script, necessary for example when using sigdev with npm

### DIFF
--- a/docker-run
+++ b/docker-run
@@ -128,7 +128,7 @@ def main():
         docker_cmd.append("--volume={socket}:{socket}".format(socket="/var/run/docker.sock"))
 
     if options.home:
-        internal_home = "/home/{login}".format(login=login) if is_windows else os.environ["HOME"]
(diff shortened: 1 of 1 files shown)


jwkaltz:
FYI @danduk82 and @ger-benjamin

ger-benjamin (approved):
Looks good, but we lost flexibility. It is okay to not be able anymore to specify a direction via the `HOME` environment variable ?

sbrunner:
I just see that the line 131 should also be changed ...

jwkaltz:
> I just see that the line 131 should also be changed ...

@sbrunner I changed this as well, now

sbrunner:
@ger-benjamin we steel can use the HOME for the external home directory, this is essentially for the user sigdev where his home is '/var/sigdev' and the '/var' is readonly inside Docker...

PR #7570, opened 2021-01-05
Hide the timezone

sbrunner:
see also: https://github.com/camptocamp/ngeo/pull/6652


ger-benjamin (approved):
Thanks.
Or it is better to hide it only in desktop (perhaps we can see this input in desktop_alt, as example ?)

sbrunner:
The desktop alt is configured only in the demo...

sbrunner:
And it didn't make a real sense for me...

PR #7577, opened 2021-01-06
Ignore startup quote on all the search therm

sbrunner:
fix GSGMF-1425


ger-benjamin (approved):
Looks good, please merge, then I'll test on a real use case

arnaud-morvan (approved):
LGTM

PR #7591, opened 2021-01-07
Be able to easyer to switch from internal and mutialised print

sbrunner:


geoportal/c2cgeoportal_geoportal/scaffolds/create/env.default_tmpl
```diff
@@ -18,6 +18,8 @@ AUTHTKT_SAMESITE=Lax
 BASICAUTH=False
 GEOPORTAL_INTERNAL_URL=http://geoportal:8080
 MUTUALIZED_PRINT_URL=mutualized-print.apps.openshift-ch-1.camptocamp.com
+DISABLE_MUTUALIZED_PRINT=#
```
arnaud-morvan, line 21: Maybe a little comment here to explain the possible values ('' or '#')

arnaud-morvan (approved):
LGTM

geoportal/c2cgeoportal_geoportal/scaffolds/create/env.default_tmpl
ger-benjamin: Have we not escape (or double) the `#` ?

PR #7622, opened 2021-01-13
Be able to log in JSON from the Python plugin

sbrunner:
To have more readable and more rich messages in Kibana


arnaud-morvan (approved):
LGTM

PR #7623, opened 2021-01-13
Fix editingEnumerations

arnaud-morvan:
_AssociationProxy.value_attr need to be a str not a InstrumentedAttribute.
See getattr(target, self.value_attr) in _AssociationProxy.__get__ and _AssociationProxy.__set__.
Note it is the same in sqlalchemy.ext.associationproxy.AssociationProxy.

For consistency, do the same for _AssociationProxy.order_by (not InstrumentedAttribute but str).


sbrunner (approved):
Looks good :-)

PR #7633, opened 2021-01-14
Fix the header in desktop alt

sbrunner:


arnaud-morvan (approved):
Trust you on this

arnaud-morvan (commented):
Even I f I do not understand every thing here, this sound a little limited about interface names.
It might be adapted in customer project if need, and it is not trivial.
Maybe I miss something.

PR #7682, opened 2021-01-22
Use GDAL as base image

sbrunner:


Dockerfile
```diff
@@ -23,13 +26,21 @@ RUN \
     apt-get install --assume-yes --no-install-recommends 'nodejs=14.*' && \
     apt-get clean && \
     rm --recursive --force /var/lib/apt/lists/*
+
+RUN \
+    . /etc/os-release && \
+    apt-get update && \
+    apt-get install --assume-yes --no-install-recommends
```
arnaud-morvan, line 33: Here it install nothing but leave package cache in image ?

Dockerfile
```diff
@@ -39,15 +50,18 @@ ENV TEST=false
 # Finally used for all misk task, will not be used on prod runtime
 
 FROM base AS tools
+SHELL ["/bin/bash", "-o", "pipefail", "-ceux"]
```
arnaud-morvan, line 53: What does this ? Maybe add a comment.

arnaud-morvan (approved):
Looks good, just some things that I do not really understand.

Dockerfile
sbrunner: Oups thanks :-)

Dockerfile
sbrunner: => comment added:
# Fail on error on pipe, see: https://github.com/hadolint/hadolint/wiki/DL4006.
# Treat unset variables as an error when substituting.
# Print commands and their arguments as they are executed.

PR #7690, opened 2021-01-23
Fix urllogin authentication

arnaud-morvan:
UrlAuthentication need settings from config.yaml

UrlAuthentication should be checked before cookie.


sbrunner (approved):
I trust you :-)

PR #7705, opened 2021-01-26
Use WMS version from OGC server URL to GetCapabilities 

vuilleumierc:
This is to use the parameters that are given in the OGC server URL instead of overwriting with default values (-> use a specific WMS version)

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/theme.py
+++ b/geoportal/c2cgeoportal_geoportal/views/theme.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2011-2020, Camptocamp SA
+# Copyright (c) 2011-2021, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
@@ -189,9 +189,10 @@ class Theme:
         def build_web_map_service(ogc_server_id):
             del ogc_server_id  # Just for cache
 
+            version = urllib.parse.parse_qs(urllib.parse.urlsplit(url).query)['VERSION'][0]
             layers = {}
             try:
-                wms = WebMapService(None, xml=content)
+                wms = WebMapService(None, xml=content, version=version)
             except Exception as e:  # pragma: no cover
                 error = (
                     "WARNING! an error '{}' occurred while trying to read the mapfile and recover the themes."
--- a/geoportal/tests/functional/test_themes_mixed.py
+++ b/geoportal/tests/functional/test_themes_mixed.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2016-2019, Camptocamp SA
+# Copyright (c) 2016-2021, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
@@ -63,6 +63,18 @@ class TestThemesView(TestCase):
         ogc_server_external = OGCServer(
             name="__test_ogc_server_external", url="http://wms.geo.admin.ch/", image_type="image/jpeg"
         )
+        ogc_server_valid_wms_version = OGCServer(
+            name="__test_ogc_server_valid_wms_version",
+            url=mapserv_url + "?VERSION=1.3.0",
+            image_type="image/jpeg",
+            wfs_support = False,
+        )
+        ogc_server_invalid_wms_version = OGCServer(
+            name="__test_ogc_server_invalid_wms_version",
+            url=mapserv_url + "?VERSION=1.0.0",
+            image_type="image/jpeg",
+            wfs_support = False,
+        )
 
         layer_internal_wms = LayerWMS(name="__test_layer_internal_wms", public=True)
         layer_internal_wms.layer = "__test_layer_internal_wms"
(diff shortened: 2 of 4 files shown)


jwkaltz:
@vuilleumierc can you please provide an example URL, it will make the proposed change easier to understand, I think.

vuilleumierc:
OGC server URL: 
`https://www.geoportal.ch/services/wms/ktai?VERSION=1.3.0`
The GetCapabilities request in GMF:
`https://www.geoportal.ch/services/wms/ktai?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetCapabilities&ROLE_ID=0&USER_ID=0`
I also tried with
`https://www.geoportal.ch/services/wms/ktai?SERVICE=WMS&VERSION=1.3.0`
and obtained the same results

vuilleumierc:
It turns out that the WMS version has to be passed as argument to the OWSLib function - the value from the URL is not taken into account and it defaults to 1.1.1. I will try to update this pull request accordingly
The GetCapabilities from the AI server does work with version 1.3.0:
```
Python 3.7.9 (default, Aug 18 2020, 02:07:21) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from owslib.wms import WebMapService
>>> wms = WebMapService("https://www.geoportal.ch/services/wms/ktai?VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities&ROLE_ID=0&USER_ID=0", version='1.3.0')
https://www.geoportal.ch/services/wms/ktai?ROLE_ID=0&USER_ID=0
/home/cvuilleumier/.local/lib/python3.7/site-packages/owslib/map/wms130.py:122: UserWarning: Content metadata for layer "ch.geoportal.ver_entsorgung_kommunikation.843.0.abwasser_werkplan_ausser_betrieb_gde" already exists. Using child layer
...
```
without version as an argument:
```
>>> wms = WebMapService("https://www.geoportal.ch/services/wms/ktai?VERSION=1.3.0&SERVICE=WMS&REQUEST=GetCapabilities&ROLE_ID=0&USER_ID=0")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/cvuilleumier/.local/lib/python3.7/site-packages/owslib/wms.py", line 53, in WebMapService
    timeout=timeout, headers=headers, auth=auth)
  File "/home/cvuilleumier/.local/lib/python3.7/site-packages/owslib/map/wms111.py", line 87, in __init__
    self._buildMetadata(parse_remote_metadata)
  File "/home/cvuilleumier/.local/lib/python3.7/site-packages/owslib/map/wms111.py", line 96, in _buildMetadata
    self.identification = ServiceIdentification(serviceelem, self.version)
  File "/home/cvuilleumier/.local/lib/python3.7/site-packages/owslib/map/wms111.py", line 359, in __init__
    self.type = testXMLValue(self._root.find('Name'))
AttributeError: 'NoneType' object has no attribute 'find'

```

sbrunner:
For me, this pull request should not introduce an issue in GeoMapFish, but I can't determine if what you want to do will work...

But you should at least add a message in the CONT_CHANGELOG.

geoportal/c2cgeoportal_geoportal/views/theme.py
```diff
@@ -189,9 +189,10 @@ async def _wms_getcap(self, ogc_server, preload=False):
         def build_web_map_service(ogc_server_id):
             del ogc_server_id  # Just for cache
 
+            version = urllib.parse.parse_qs(urllib.parse.urlsplit(url).query)['VERSION'][0]
```
sbrunner, line 192: We call `urlsplit` and `parse_qs` 3 times on the same URL ...
Looks that we should refactor it a little on master...

vuilleumierc:
Thanks for your feedback @sbrunner. I am looking into adding new tests regarding this change, let me know if you have any thoughts on that.
I will squash the commits :)

vuilleumierc:
In the [mixed theme test](https://github.com/camptocamp/c2cgeoportal/pull/7705/files#diff-5aa6c5ddb678602b7d15f69e11f3c6f1f06f9112545d915ff8e5e64c0afe1251) there are two new layers, one with an OGC server with a valid WMS version (-> added to the theme) and the other with an invalid version (-> not added). Both are calling the internal MapServer.
I also edited the filter applied on the errors (regex) because I think it was working for wrong reasons (parenthesis not escaped).
Also I noticed the external OGC server (http://wms.geo.admin.ch/) in this same test is not used, maybe we should remove it from the code?

geoportal/tests/functional/test_themes_mixed.py
```diff
@@ -172,15 +200,23 @@ def _only_name(self, item, attributes=None):
...
         ]
+
+        regex = re.compile(
+            r"The (GeoMapFish|WMS) layer name '[a-z0-9_]*', cannot be two times in the same block \(first level group\)."
+        )
+        errors = [e for e in errors if not regex.match(e)]
+
         regex = re.compile(
-            r"The (GeoMapFish|WMS) layer name '[a-z0-9_]*', cannot be two times in the same block (first level group)."
+            r"^WARNING! an error \'The WMS version \(\d\.\d\.\d\) you requested is not implemented."
```
sbrunner, line 216: Shouldn't we test the exact message?

geoportal/tests/functional/test_themes_mixed.py
```diff
@@ -172,15 +200,23 @@ def _only_name(self, item, attributes=None):
...
         )
-        errors = [e for e in errors if regex.match(e)]
+        errors = [e for e in errors if not regex.match(e)]
```
sbrunner, line 218: For me, this method shouldn't be changed and after we should test somthing like:
```
self.assertEqual(self._get_filtered_errors(themes), {"WARNING! an error 'The WMS version (1.0.0) you requested is not implemented."})
```

ger-benjamin (commented):
Looks good as far as I know.
But I don't know how to test that

geoportal/tests/functional/test_themes_mixed.py
```diff
@@ -172,15 +200,23 @@ def _only_name(self, item, attributes=None):
...
         errors = themes["errors"]
+
         errors = [
             e
             for e in errors
             if e != "The layer '' (__test_layer_external_wms) is not defined in WMS capabilities"
         ]
+
+        regex = re.compile(
+            r"The (GeoMapFish|WMS) layer name '[a-z0-9_]*', cannot be two times in the same block \(first level group\)."
```
danduk82, line 211: +1 for catching `\( \)`

danduk82 (approved):
LGTM

sbrunner (approved):

PR #7719, opened 2021-01-29
Replace pipenv install by pipenv sync

arnaud-morvan:


Makefile
```diff
@@ -21,7 +21,7 @@ checks: ## Run the application checks
 checks: yamllint otherchecks
 
 pipenv.timestamp:
-	pipenv install
+	pipenv sync
```
sbrunner, line 24: The `touch $@` is missing or the `.PHONY` and should be renamed ...

sbrunner (approved):

Makefile
arnaud-morvan: Done and added dependency on Pipfile.lock.

arnaud-morvan: Note that in GMF 2.6 we install the whole Pipfile.lock just for yamllint.

PR #7765, opened 2021-02-05
Fix upgrade to an rc

sbrunner:


ger-benjamin (approved):
Let's try
(I've no clue if something else could go wrong)

PR #7785, opened 2021-02-08
Update the documentation to invalidate the cache

sbrunner:
Fix: GSGMF-1470


arnaud-morvan (commented):
I do not see any command to invalidate the cache in c2cwsgiutils README.

arnaud-morvan (approved):
Ok, I suppose we should open <application main URL>/c2c and we find it there.

PR #7805, opened 2021-02-11
Add some more generation commands

sbrunner:
To be able to generate the WMTS legend

See result
https://geomapfish-demo-2-6.camptocamp.com/tiles/admin/


ger-benjamin (approved):
I don't have enough context to really review, but I don't see any obvious error.
Except maybe your example that doesn't look to work

PR #7881, opened 2021-03-02
Remove default URL length limit in Gunicorn

asaunier:
We had several users reporting GMF crashes with long URLs (GMF has often permalink with a lot of params) such as:
```
Request Line is too large (5722 > 4094)
```

This is related to Gunicorn limiting by default URLs to 4094 chars: https://docs.gunicorn.org/en/stable/settings.html#limit-request-line

The PR suggests to get rid of the Gunicorn default limit.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/env.default_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/env.default_tmpl
@@ -33,7 +33,8 @@ MAPSERVER_URL=http://mapserver:8080/
 TINYOWS_URL=http://tinyows:8080/
 QGISSERVER_URL=http://qgisserver:8080/
 QGIS_VERSION=3.16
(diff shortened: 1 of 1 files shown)


jwkaltz (approved):

arnaud-morvan (approved):
LGTM.
Note that the gunicorn documentation says this limit is used to prevent DDOS attacks.

jwkaltz:
> this limit is used to prevent DDOS attacks.

Indeed, but since Gunicorn is normally not the front that is used

sbrunner (changes requested):
Not sour that completely remove this limit is a good choice ...

sbrunner:
I prefer to set it to 8190...

asaunier:
@sbrunner OK, I have updated the PR accordingly.
@jwkaltz @arnaud-morvan Still OK for you?

sbrunner (approved):
Thanks :-)

arnaud-morvan (approved):

jwkaltz (approved):

asaunier:
Setting the limit to 8190 was the good choice here as the shorttener raises an exception if the URL > 8190 char anyway
https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/c2cgeoportal_geoportal/views/shortener.py#L81-L82

PR #7978, opened 2021-03-16
Improve changelog about the relocation of the config to the vars.yaml

ger-benjamin:
For GSGMF-1526

I've tried to improve the explanation of the relocation of the config to the vars.yaml file.
I'm a little bit less concise, with some repetitions but if think that's more logical for a less experimented user.

Questions:
 - I use `geoportal/<project_package>_geoportal/` syntaxe, can we do better ? (like `geoportal/${project_package}_geoportal/` ?)
 - Could we add a link to a doc that explain yaml section template`&mysection`, `*mysection`

The only section I've removed is this one, thinking it's too technical and not necessary in the changelog:
```
Then configure the `vars/srid`, `vars/alternate_projections`, `vars/resolutions` and `vars/extent` they will
be dispatched using `c2ctemplate` or YAML link.
```

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
@@ -11,42 +11,74 @@ Client application configuration
 ................................
 
 Previous versions had configuration in the `*.html.ejs`, in the `Controler*.js`, and in the `vars.yaml` file.
-The goal of this change is to move this configuration to the `vars.yaml` file (with default values in
+The goal of this change is to group all this configuration to the `vars.yaml` file (with default values in
 the `CONST_vars.yaml`). This will make future upgrades easier and increase configuration possibilities
 in the simple application mode, see the documentation
 <https://camptocamp.github.io/c2cgeoportal/${MAIN_BRANCH}/integrator/create_application.html#simple-application>.
 
 
-When you apply the `ngeo.diff` you should first apply the changes in the `geoportal/vars.yaml` file.
+When you apply the `ngeo.diff` you will have to remove lines of configuration from your `*.html.ejs` and
+`Controler*.js` files. You must compare each removed line of configuration with its default value newly
+defined in the section `vars/interface_config` of your `geoporal/CONST_vars.yaml` file. If you use a custom
+configuration, you have to override the default value by adding a new one in your `geoportal/vars.yaml`.
 
-For the next step, you should consult the documentation about:
-- The interfaces_config documentation
+
+In the `vars/interface_config` section of `geoportal/vars.yaml` and `geoporal/CONST_vars.yaml` files, you
+define configuration per `interface` and then mainly per `constants` and `routes` (for paths). See all
+possibility in the `Dynamic.json view` part of the documentation
 <https://camptocamp.github.io/c2cgeoportal/${MAIN_BRANCH}/integrator/ngeo.html?highlight=ngeo#dynamic-json-view>
+
+
+You can find a list of existing options in the `geoporal/CONST_vars.yaml` and here with descriptions:
 - GMF constants definitions <https://camptocamp.github.io/ngeo/master/jsdoc/module-contribs_gmf_src_options.html>
 - ngeo constants definitions <https://camptocamp.github.io/ngeo/master/jsdoc/module-src_options.html>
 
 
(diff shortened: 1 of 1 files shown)


ybolognini (approved):
I can't tell about the technical aspect of these changes, but I feel like the fact it's more verbose is a good thing.

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -11,42 +11,74 @@ Client application configuration
...
 
 Previous versions had configuration in the `*.html.ejs`, in the `Controler*.js`, and in the `vars.yaml` file.
-The goal of this change is to move this configuration to the `vars.yaml` file (with default values in
+The goal of this change is to group all this configuration to the `vars.yaml` file (with default values in
 the `CONST_vars.yaml`). This will make future upgrades easier and increase configuration possibilities
 in the simple application mode, see the documentation
 <https://camptocamp.github.io/c2cgeoportal/${MAIN_BRANCH}/integrator/create_application.html#simple-application>.
 
 
-When you apply the `ngeo.diff` you should first apply the changes in the `geoportal/vars.yaml` file.
```
sbrunner, line 20: For me, this line should be kept because a part of the config structure is in the new vars.yaml

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
```diff
@@ -11,42 +11,74 @@ Client application configuration
...
+When you apply the `ngeo.diff` you will have to remove lines of configuration from your `*.html.ejs` and
+`Controler*.js` files. You must compare each removed line of configuration with its default value newly
+defined in the section `vars/interface_config` of your `geoporal/CONST_vars.yaml` file. If you use a custom
+configuration, you have to override the default value by adding a new one in your `geoportal/vars.yaml`.
```
sbrunner, line 23: For me, we can do a shortcut, if in the diff we see that we didn't change the default value we can apply this change without extra check...

sbrunner:
> I use geoportal/<project_package>_geoportal/ syntaxe, can we do better ? (like geoportal/${project_package}_geoportal/ ?)

You are in a _tmpl file then you can directly use {{package}} and it will be replaced :-) (Thanks for the correction :-) )

sbrunner:
> The only section I've removed is this one, thinking it's too technical and not necessary in the changelog: ...

Don't you think that's good to explain? For me, it's needed to be able to correctly configure them because it's a real exception...

sbrunner (approved):

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
ger-benjamin: I understand now, I'll keep it.

geoportal/c2cgeoportal_geoportal/scaffolds/update/CONST_CHANGELOG.txt_tmpl
ger-benjamin: Proposale:

When you apply the `ngeo.diff` you will have to remove lines of configuration from your `*.html.ejs` and
`Controler*.js` files. If your configuration diverge from the default configuration you have to override the default value (newly defined in the `geoportal/CONST_vars.yaml`) by adding a new one in your `geoportal/vars.yaml`.

sbrunner: +1 you just miss the precision: in the section `vars/interface_config` :-)

ger-benjamin: Thanks I've adapted the text (and little bit more than what I wrote here, but the same idea)

ger-benjamin:
> I use geoportal/<project_package>_geoportal/ syntaxe, can we do better ? (like geoportal/${project_package}_geoportal/ ?)
>You are in a _tmpl file then you can directly use {{package}} and it will be replaced :-) (Thanks for the correction :-) )

Thanks, done !

>The only section I've removed is this one, thinking it's too technical and not necessary in the changelog: ...
>Don't you think that's good to explain? For me, it's needed to be able to correctly configure them because it's a real exception...

I've added it again, but after the example. I think that's a little bit confusing before

sbrunner (approved):
Looks good to me :-)

PR #7994, opened 2021-03-24
Pull the base images before building the image

sbrunner:
Fix: GSGMF-1512


sbrunner:
Should we also run `docker-compose pull`?

arnaud-morvan (approved):
Would be simpler with "docker build --pull"

arnaud-morvan:
Not sure we need docker-compose pull on each build (note it is easy to do manually).
But for standard user it not trivial to pull the geoportal base image.

PR #8012, opened 2021-03-29
Fix admin i18n in simple application mode

sbrunner:
Replace: https://github.com/camptocamp/demo_geomapfish/pull/222/
Fix: GSGMF-1489

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
@@ -52,6 +52,8 @@ def main():
     parser.add_argument("--verbose", help="Display the docker build commands")
     parser.add_argument("--config", action="store_true", help="Build only the configuration image")
     parser.add_argument("--geoportal", action="store_true", help="Build only the geoportal image")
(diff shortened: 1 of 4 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
```diff
@@ -113,6 +115,11 @@ def main():
     default = not (args.config or args.geoportal)
 
     if default or args.config:
+        simple = not os.path.exists("geoportal/Dockerfile")
+        if args.simple:
+            simple = True
+        if args.not_simple:
+            simple = False
```
arnaud-morvan, line 122: Is it not necessary to pass this as a build arg to the docker build command for the config image ?

arnaud-morvan (approved):
Seems it miss the docker build arg for the config image.

geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl
sbrunner: Arg +1 :-)

PR #8036, opened 2021-04-01
Make standard roles customisable

sbrunner:
This can be required to have multiple organization in one container


arnaud-morvan (approved):

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -327,6 +327,7 @@ def includeme(config: pyramid.config.Configurator):
     get_user_from_request = create_get_user_from_request(settings)
     config.add_request_method(get_user_from_request, name="user", property=True)
     config.add_request_method(get_user_from_request, name="get_user")
+    config.add_request_method(lambda request, role_type: role_type, name="get_standard_role")
```
arnaud-morvan, line 330: Maybe desserve a comment as this does nothing right now.
If this is about organizations, should it appear in request method name ?
=> get_organization_role

arnaud-morvan: For me the standard role is "intranet" and this return organization tinted one.

PR #8040, opened 2021-04-01
Some documentation about editing

RenataMuellerC2C:


doc/administrator/editing.rst
```diff
@@ -76,6 +78,45 @@ corresponding to this field is *Related Postgres table* in the admin interface.
...
+Configure layer editing
+-----------------------
+An overview of the metadata for the configuration of the editing can be found here:
+:ref:`administrator_administrate_metadata`.
+
+Registering edit time and user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* To register the time of the change of an edited item, a column of type
+  ``timestamp without timezone`` is needed in the table of the layer. In
+  the configuration of the layer ``lastUpdateDateColumn`` has to be added
```
jwkaltz, line 90: This is also described in https://camptocamp.github.io/c2cgeoportal/2.6/administrator/administrate.html?highlight=lastupdatedatecolumn

Should it be removed from there, and a link provided there to this new section?

doc/administrator/editing.rst
```diff
@@ -8,6 +8,8 @@ This section explains how to configure the edit function in c2cgeoportal.
 Just like most administrative tasks, setting up editing in a c2cgeoportal application involves intervening
 in the database, through the c2cgeoportal administration interface.
 
+.. _requirements:
```
sbrunner, line 11:
```suggestion
.. _administrator_editing_requirements:
```

doc/administrator/editing.rst
```diff
@@ -76,6 +78,45 @@ corresponding to this field is *Related Postgres table* in the admin interface.
...
 
+Configure layer editing
+-----------------------
+An overview of the metadata for the configuration of the editing can be found here:
+:ref:`administrator_administrate_metadata`.
+
+Registering edit time and user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* To register the time of the change of an edited item, a column of type
+  ``timestamp without timezone`` is needed in the table of the layer. In
```
sbrunner, line 89: Are you sure about this? shouldn't be `timestamp with timezone`?

arnaud-morvan: As I see, `timestamp with timezone` is not supported as the setted value is not timezone aware : https://github.com/camptocamp/c2cgeoportal/blob/c87b857bc2f08051ca514e3a37b2c32099e71ed5/geoportal/c2cgeoportal_geoportal/views/layers.py#L384

arnaud-morvan: To be able to use a `timestamp with timezone` we would need `datetime.now(timezone.utc)`

sbrunner (approved):

PR #8045, opened 2021-04-06
Allows the attributes from the WFS DescribeFeatures to be None

sbrunner:


arnaud-morvan (approved):
May be more precise in pull request title as it appears in Change log.
This does not allow "something" (would like to know what) to be None.
```
                    if attributes is not None:
                        attributes = dict(attributes)
```
As if it is None, the change does nothing.

PR #8046, opened 2021-04-06
Allows nearly all variable to be from environment variable

sbrunner:


arnaud-morvan (approved):
I do not see any better solution to allow using env vars here.
So +1

arnaud-morvan:
Be careful that some variables may not be correctly parsed (typed) at runtime.

sbrunner:
@arnaud-morvan effectively if you do that you should do something like:
 ```
 runtime_postprocess:
  - expression: int({})
    vars:
       - ...
```
or
 ```
 runtime_postprocess:
  - expression: {}.lower() in ('true', 'yes', '1')
    vars:
       - ...
```
I see that some vars are already in, do you think that I should add the others?

arnaud-morvan:
Not sure, but I won't have found how to handle this myself, so it suppose it is the same for everyone except you.

PR #8056, opened 2021-04-13
Add Dependabot vulnerabilities and issues in status page

sbrunner:
By the way shouldn't it be moved to https://github.com/camptocamp/geospatial-dashboards?


arnaud-morvan (approved):
Looks good, "Add Dependabot vulnerabilities and issues" => "Add Dependabot vulnerabilities and issues in status page"

PR #8064, opened 2021-04-13
Fix the vars

sbrunner:


arnaud-morvan (approved):
Loosk good, maybe use a more precise tittle for the changelog.

PR #8068, opened 2021-04-14
Fix admin login

sbrunner:
The fautive: https://github.com/camptocamp/c2cgeoportal/pull/7690
Fix: GSGMF-1569


arnaud-morvan (approved):
LGTM

PR #8099, opened 2021-04-16
Fix markers URL in the new applications

sbrunner:


build.mk
```diff
@@ -125,6 +125,7 @@ $(APPS_PACKAGE_PATH)/static/apihelp: /usr/lib/node_modules/ngeo/api/dist/apihelp
 	sed -i -e 's#"data\.txt"#"../static/$${CACHE_VERSION}/apihelp/data.txt"#g' $@/index.html.tmpl_tmpl
 	sed -i -e "s#'data\.txt'#'../static/\$${CACHE_VERSION}/apihelp/data.txt'#g" $@/index.html.tmpl_tmpl
 	sed -i -e 's#img/#../static/$${CACHE_VERSION}/apihelp/img/#g' $@/index.html.tmpl_tmpl
+	sed -i -e "s#https://geomapfish-demo-2-[0-9].camptocamp.com/0/img/markers/#../static/0/apihelp/markers/#g" $@/data.txt
```
ochriste, line 128: why is there a reference to the demo in the c2cgeoportal build code ?

ger-benjamin: Previously, these markers was on the openlayers 2 example server (yes, the v2).
We have moved them on a server we own. This one was the better we have found.

ger-benjamin (approved):

build.mk
sbrunner: To get the apps and API help from ngeo...

ochriste: @ger-benjamin mmm so it's like an external online resource?
@sbrunner Im sorry, I dont understand your answer, but you will be able to explain to me more clearly during the daily monday

PR #8123, opened 2021-04-20
Add inotify as a dev dep, required by gunicorn to live reload

RemiDesgrange:
See https://docs.gunicorn.org/en/20.0.4/settings.html#reload


sbrunner:
I don't understand why you need that, for me, it's working, and the inotify package don't look to be well maintained, nearly no changes the 3 last years...
@arnaud-morvan what do you think?

sbrunner:
Stall -> closing

PR #8144, opened 2021-04-27
Separate with a dash like in GitHub workflows

sbrunner:


ger-benjamin (approved):

arnaud-morvan (commented):
I do not understand, but I suppose you know what you are doing.

PR #8162, opened 2021-04-30
Change the Tinyows Docker image tag to master which is regularly rebuilt

sbrunner:


arnaud-morvan (approved):
Maybe change PR title to:
Change the tinyows docker image tag to master which is regularly rebuilt

PR #8246, opened 2021-05-19
Fix the documentation about the static in dynamics

sbrunner:


ger-benjamin (approved):

doc/integrator/ngeo.rst.tmpl
```diff
@@ -99,8 +99,10 @@ The sub section is the interface name, and after that we have:
     (but can not be removed via interface configuration).
 * ``constants``: Directly define an ``angular`` constant in the vars file.
 * ``dynamic_constants``: Define an ``angular`` constant from a dynamic values.
-* ``static``: key: Constant name, value: name of the static view that we want to have the URL.
-* ``routes``: key: Constant name, value:
+* ``static``, the ``key`` is the constant name:
+    ``name``: The path of the ressource witch one we want to have the URL.
```
jwkaltz, line 103: I don't understand this sentence

doc/integrator/ngeo.rst.tmpl
```diff
@@ -99,8 +99,10 @@ The sub section is the interface name, and after that we have:
...
+    ``append``: Somthing we want to append.
```
jwkaltz, line 104: Typo: Something

Can we say more about what the Something might be?

jwkaltz (changes requested):

doc/integrator/ngeo.rst.tmpl
sbrunner: => https://docs.pylonsproject.org/projects/pyramid/en/1.10-branch/api/request.html#pyramid.request.Request.static_url ...

doc/integrator/ngeo.rst.tmpl
sbrunner: a text ...

doc/integrator/ngeo.rst.tmpl
jwkaltz: So is it 
"the relative path of the resource for which an URL is to be provided"
?

sbrunner:
@jwkaltz up

jwkaltz:
> @jwkaltz up

@sbrunner I still don't understand the sentence "``name``: The path of the ressource witch one we want to have the URL."

Is it "the relative path of the resource for which an URL is to be provided"?
Or "the path of the resource for which we want to provide the URL"?

jwkaltz (approved):
I don't want to block the merge, but I think the descriptions of "name" are not really clear, see my latest comment

doc/integrator/ngeo.rst.tmpl
sbrunner: Usualy it's an absolute path that starts with '/etc/geomapfish/static/' it can also start with a python package name like `<package>:<path_relative_to_the_package>`but it's not a common uasge...

sbrunner: I update it to:
The path of the resource witch one we want to have the URL e.g. ``/etc/geomapfish/static/contextualdata.html``.

PR #8272, opened 2021-05-26
Add a script to get the versions upgrades

sbrunner:
Fix: GSGMF-1603, See result:  #8271


ger-benjamin (approved):
Looks working well.
But we should not forget to add a note when a version of an image (like postgres) change. That's probably a manual task to do.

sbrunner:
> But we should not forget to add a note when a version of an image (like postgres) change. That's probably a manual task to do.

+1 :-)

PR #8275, opened 2021-05-27
Root should always have read access

sbrunner:


docker/qgisserver/geomapfish_qgisserver/accesscontrol.py
```diff
@@ -530,13 +530,16 @@ def layerPermissions(self, layer):  # NOQA
 
             session = self.DBSession()
             try:
+                roles = self.get_roles(session)
+
+                if roles == "ROOT":
```
jwkaltz, line 535: Is roles not a list (of roles)?

sbrunner: Not if we are root ...

arnaud-morvan (commented):
Is it not already the case with : https://github.com/camptocamp/c2cgeoportal/blob/master/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py#L402

sbrunner:
@arnaud-morvan As I understand no because it's done here:
https://github.com/camptocamp/c2cgeoportal/blob/0f4a8793ba092d3cd846261b65bbe5525a0ab83e/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py#L545-L550
And herer https://github.com/camptocamp/c2cgeoportal/blob/0f4a8793ba092d3cd846261b65bbe5525a0ab83e/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py#L540-L541
We return if the layer is not in the gmf layers...

Do you agree or I miss something?

arnaud-morvan (approved):
Ok, I missed this return statement.

PR #8297, opened 2021-06-02
Add optional suffix in l10n

sbrunner:


ger-benjamin (approved):
We have only to do that in this file ? Not in Makefile, not in update-po..?

Also, In which case that's useful ? Could we have a doc that explain we can add a suffix in some cases and on which cases ?

sbrunner:
I's not really comfortable, But I add some documentation :-)

ger-benjamin (approved):
Thanks

PR #8302, opened 2021-06-03
Fix encoding issue for QGIS projects in database

sbrunner:


arnaud-morvan (approved):
This is a little complex to understand for common integrators, but with no risk as it is done in vars file itself.
It might be proper to handle params directly in get_url2 but this might not handle all use cases of servers in all containers.
IMHO this is an acceptable fix.

sbrunner:
To be sure @arnaud-morvan, can you do a second review?

arnaud-morvan (approved):
LGTM

PR #8325, opened 2021-06-08
Fix issue with Different localisation sets

sbrunner:
Currently, the po files were removed on the first call => do only one call
Convert in Python to have the Python argument parser


arnaud-morvan (approved):
Look good, note that subprocess.run is recommended with  possible.

PR #8400, opened 2021-06-24
Send MapServer substitution only on MapServer with standard auth

sbrunner:
Fix some typo

Related to GEO-4467

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/mapserverproxy.py
+++ b/geoportal/c2cgeoportal_geoportal/views/mapserverproxy.py
@@ -80,14 +80,18 @@ class MapservProxy(OGCProxy):
 
         # Do not allows direct variable substitution
         for k in list(self.params.keys()):
-            if k[:2].capitalize() == "S_":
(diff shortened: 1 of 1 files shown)


ochriste:
@sbrunner whats the idea here? we can only use mapserver substitution with standart auth? does that mean nyon use non standart auth? or is the issue elsewhere?

sbrunner:
They use a no auth with arcgis, and they have an issue because the URL is too long for unneeded parameters...

ochriste:
@sbrunner  so the issue is that with arcgis server, there is an url length limitation that make authentication impossible because adding extra parameter to pass the auth info make the whole url too long?
shouldnt we document this kind of things somewhere?

ochriste:
@sbrunner what about qgis-server backend? it doesnt support authentication either?

sbrunner:
QGIS server access control plugin currently didn't support MapServer substitution :-P 
At therm, he should support column filtering but throw another way...

ochriste:
@sbrunner ok thank you

PR #8431, opened 2021-07-01
Fix Dependabot errors

sbrunner:
Also:
- clean the time
- put weekly if possible (daily for Python and for ngeo)
- remove no more used Python 3.6 for QGIS


arnaud-morvan (approved):
LGTM

PR #8440, opened 2021-07-02
Avoid having a completely empty project on project load error

sbrunner:
@arnaud-morvan What do you think?


arnaud-morvan (approved):
Not sure to understand what this change.
If project failed to load, can we access any layer ?

arnaud-morvan:
Not sure this is still required since we use correct settings with ignore bad layers.

sbrunner:
I agree, closing :-)

PR #8444, opened 2021-07-06
Avoid error with OGCServerSynchronizer.check_layers when quote in WMS layer name

arnaud-morvan:


sbrunner:
Does it relay avoid error, I just see that you add an error message, or QGIS will remove the layer that valid is `False`?

arnaud-morvan:
With a quote in the name, the line `capabilities.find(".//Layer[Name='{}']".format(name))` raise an error, so the complete `layers_check` fails.
I didn't find any solution to fix the `find` with a quote in the name.
And IMHO it's not a good idea to have quotes in a WMS layer name, so I just set a value for information in the admin interface, it has no other effect.
So yes it avoid an error in `layers_check`.

sbrunner (approved):
Thanks  to explain :-)

PR #8453, opened 2021-07-08
Don't cache HTTP errors

sbrunner:


arnaud-morvan (approved):
LGTM

arnaud-morvan:
Seems it could be backported to 2.6 but maybe you want to test in on 2.7 demo first

sbrunner:
Effectively, I prefer that's better tested :-)

PR #8520, opened 2021-07-30
Be able to have multiple instance of an application

sbrunner:
e.g. we can have a branch named 'int-2-6', currently the image stream should be names `int-2-6-c2cgeoportal-config:int-2-6` and the real names are: `int-org1-2-6-c2cgeoportal-config:int-2-6` and `int-org2-2-6-c2cgeoportal-config:int-2-6` we can call:
```
ci/trigger --image-stream-name={env}-org1-{version_only}-c2cgeoportal-{image}:{version} --image-stream-name={env}-org2-{version_only}-c2cgeoportal-{image}:{version}
```

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/ci/trigger
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/ci/trigger
@@ -9,6 +9,15 @@ import sys
 def main():
     parser = argparse.ArgumentParser(description="Publish Docker images")
     parser.add_argument("--image", dest="images", action="append", help="The image to be exported")
+    parser.add_argument(
+        "--image-stream-name",
+        dest="imagestreams",
+        action="append",
+        help=(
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/ci/trigger
```diff
@@ -9,6 +9,12 @@ import sys
 def main():
     parser = argparse.ArgumentParser(description="Publish Docker images")
     parser.add_argument("--image", dest="images", action="append", help="The image to be exported")
+    parser.add_argument(
+        "--image-stream-name",
+        dest="imagestreams",
+        action="append",
+        help="The image stream name template, can have the values: {version}, {image}, {env}, {version_only}.",
```
arnaud-morvan, line 16: The image stream names can use template values: {version}, {image}, {env}, {version_only}. Default to {version}-c2cgeoportal-{image}:{version}.

arnaud-morvan (approved):

PR #8522, opened 2021-07-31
Fix style

sbrunner:
The style can't by apply on files that not in Git

### DIFF
--- a/ci/test-upgrade
+++ b/ci/test-upgrade
@@ -47,6 +47,13 @@ function create {
     cd -
     cd $1/testgeomapfish
     git add --all
+    if [ $2 == ${MAJOR_VERSION} ]
+    then
+        c2cciutils-checks --fix --check=isort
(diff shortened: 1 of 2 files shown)


ci/test-upgrade
```diff
@@ -47,6 +47,13 @@ function create {
     cd -
     cd $1/testgeomapfish
     git add --all
+    if [ $2 == '2.7' ]
```
arnaud-morvan, line 50: Why this custome case for 2.7 ?
Might be removed at some time ? Before 2.7 release ?
Should be add a # TODO: ? or a better test that we are on master ?

arnaud-morvan (approved):

ci/test-upgrade
sbrunner: +1 => `${MAJOR_VERSION}` thanks :-)

PR #8543, opened 2021-08-09
Add clean option in OGCServerSynchronizer

arnaud-morvan:
In OGCServer synchronize, add checkbox to remove layers which does not exist in WMS capabilities and all empty groups.


sbrunner (approved):
I trust you on this :-)

PR #8547, opened 2021-08-10
Be able to be automatically logged in to a user for development

sbrunner:


geoportal/c2cgeoportal_geoportal/lib/authentication.py
```diff
@@ -153,6 +153,19 @@ def forget(self, request: pyramid.request.Request) -> List[Dict[str, str]]:  # p
...
 
 
+@implementer(IAuthenticationPolicy)
+class DevAuthenticationPolicy(CallbackAuthenticationPolicy):  # type: ignore
+    @staticmethod
+    def unauthenticated_userid(request: pyramid.request.Request) -> Optional[str]:
+        """
+        Get the user name from the configuration
+        """
+        # Conseder empty string as no config
```
ochriste, line 163: Consider

doc/developer/debugging.rst
```diff
@@ -61,6 +61,20 @@ With the default configuration in the file ``docker-compose.override.yaml``, you
...
+
+To be automatically logged in with a user present in the database you should set in the ``vars.yaml``:
+
+.. code:: yaml
+
+  vars:
+    authentication:
+      dev_autologin: <username>
+
+Then you will dirrectily be logged in to the user.
```
ochriste, line 75: directly

sbrunner:
Thanks @ochriste :-)

doc/developer/debugging.rst
```diff
@@ -61,6 +61,20 @@ With the default configuration in the file ``docker-compose.override.yaml``, you
...
+
+To be automatically logged in with a user present in the database you should set in the ``vars.yaml``:
+
+.. code:: yaml
+
+  vars:
+    authentication:
+      dev_autologin: <username>
+
+Then you will directly be logged in to the user.
```
arnaud-morvan, line 75: It might be better to put this in development.ini file, to be sure this is never used on production.

Or add a message that this should never be committed.

arnaud-morvan: Or we could use an environment variable (that will not be present on production)

sbrunner: +1 -> environment variable :-)

arnaud-morvan: So this could go directly to CONST_vars maybe.

geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_config-schema.yaml
```diff
@@ -198,6 +198,9 @@ mapping:
           oauth2_token_expire_minutes:
             type: scalar
             required: false
+          dev_autologin:
+            type: str
+            required: false
```
arnaud-morvan, line 203: This is not used anymore.

arnaud-morvan (approved):
Addition to config schema can be removed.

geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_config-schema.yaml
sbrunner: +1 thanks :-)

PR #8548, opened 2021-08-10
Don't annotate the transaction with the user

sbrunner:
In development mode like it's already don in production mode


arnaud-morvan (approved):
Probably dogpile is not properly configured when we enter pyramid_tm tween.

sbrunner:
Effectively, It's what I expected :-)

PR #8552, opened 2021-08-10
Add interface filter in the admin interface

sbrunner:


arnaud-morvan:
I've added one commiut with a form-inline for the layer tree toolbar : 

before-form-inline

Now it looks like : 

after-form-inline

IMHO it looks better with the interface dropdown on the left.

arnaud-morvan (approved):
LGTM thanks, I added a commit with some changes to the template.

PR #8567, opened 2021-08-13
OGC server synchronizer force ordering

arnaud-morvan:
+ add typing


sbrunner (approved):
I'm confident to you :-)

PR #8617, opened 2021-08-31
Fix the came_from on HTTP Forbidden

sbrunner:


geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -92,7 +92,11 @@ def loginform403(self):
 
         return {
             "lang": self.lang,
-            "login_params": {"came_from": self.request.path},
+            "login_params": {
+                "came_from": (
+                    f"{self.request.path}?{'&'.join(['='.join(e) for e in self.request.GET.items()])}"
```
arnaud-morvan, line 97: No need for urlencode here ?
=> `f"{self.request.path}?{urllib.parse.urlencode(self.request.GET.items())}"`

sbrunner: +1 :-)

arnaud-morvan (approved):

PR #8640, opened 2021-09-06
Add description in metadatas and functionalities

arnaud-morvan:
Based on https://github.com/camptocamp/c2cgeoportal/pull/8601


arnaud-morvan:
pushed in https://github.com/camptocamp/c2cgeoportal/pull/8601

PR #8641, opened 2021-09-06
Move Admin doc to SQLA model 2.4

arnaud-morvan:


arnaud-morvan:
Duplicated with https://github.com/camptocamp/c2cgeoportal/pull/8643

PR #8650, opened 2021-09-08
Add blacklist whitelist for headers

ger-benjamin:
For GEO-4639

Add blacklist/whitelist system for headers.
Add docs.

@sbrunner Are you sure you don't want to add this feature per url like the global_header system ? 

```
vars:
  global_headers:
    - pattern: <regex>
      headers:
        <header>: <value>
```
It looks risky, or possibly problematic to apply these lists on all request going through the proxy.
(@jwkaltz it's for Agrola, but I add you especially to review the documentation)

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/proxy.py
+++ b/geoportal/c2cgeoportal_geoportal/views/proxy.py
@@ -52,7 +52,9 @@ CACHE_REGION = get_region("std")
 class Proxy:
     def __init__(self, request):
         self.request = request
-        self.host_forward_host = request.registry.settings["host_forward_host"]
+        self.host_forward_host = request.registry.settings.get("host_forward_host", [])
+        self.headers_whitelist = request.registry.settings.get("headers_whitelist", [])
+        self.headers_blacklist = request.registry.settings.get("headers_blacklist", [])
         self.http_options = self.request.registry.settings.get("http_options", {})
 
     def _proxy(self, url, params=None, method=None, cache=False, body=None, headers=None):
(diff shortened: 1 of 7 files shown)


geoportal/c2cgeoportal_geoportal/views/proxy.py
```diff
@@ -100,11 +103,24 @@ def _proxy(self, url, params=None, method=None, cache=False, body=None, headers=
...
 
         if method in ("POST", "PUT") and body is None:  # pragma: no cover
             body = self.request.body
 
+        # Filters headers with a whitelist.
+        # Some default pyramid headers will be added back by pyramid.
+        if len(self.headers_whitelist) > 0:
+            for header in list(headers):
+                if header not in self.headers_whitelist:
+                    headers.pop(header)
```
sbrunner, line 116:
```suggestion
            headers = {key: value for key, value in headers.items() if key in self.headers_whitelist}
```

geoportal/c2cgeoportal_geoportal/views/proxy.py
```diff
@@ -100,11 +103,24 @@ def _proxy(self, url, params=None, method=None, cache=False, body=None, headers=
...
+
+        # Filters headers with a blacklist.
+        # Some default pyramid headers will be added back by pyramid.
+        for header in list(headers):
+            if header in self.headers_blacklist:
+                headers.pop(header)
```
sbrunner, line 122:
```suggestion
        headers = {key: value for key, value in headers.items() if key not in self.headers_blacklist}
```

sbrunner (commented):
Looks good, just propose use functional syntax

sbrunner:
> Are you sure you don't want to add this feature per url like the global_header system?

I don't think that necessary...

ger-benjamin:
> The real error is here...

It's fixed in my 2nd commit. I don't know why I've that. The error append also on branch `2.5`

sbrunner (approved):
Looks good :-)

PR #8674, opened 2021-09-15
Fix the db in dev mode

sbrunner:


arnaud-morvan (approved):
As far as I understand, it is needed to specify PGUSER and PGDATABASE for psql. But PGPASSWORD is not really useful anymore as `psql` will authenticate in trust mode.
Ex: `docker-compose run --rm db psql`

PR #8679, opened 2021-09-17
Exit tool container on CTRL+C signal

sbrunner:


Dockerfile
```diff
@@ -55,7 +55,7 @@ FROM base AS tools
 # Print commands and their arguments as they are executed.
 SHELL ["/bin/bash", "-o", "pipefail", "-cux"]
 
-CMD ["sleep", "infinity"]
+CMD ["tail", '/f', "/dev/null"]
```
arnaud-morvan, line 58:
```suggestion
CMD ["tail", '-f', "/dev/null"]
```

arnaud-morvan (approved):

Dockerfile
sbrunner: Oups, thanks :-)

PR #8701, opened 2021-09-24
Add alternative X-Forward headers in proxy

ger-benjamin:
Clean fix for GEO-4666.

I'll backport it manually in 2.4 as some previous line interest me too.

I expect having `headers[Forwarded]` string like that:
`'by=0.0.0.0;for=134.12.1.1;host=123.12.1.1;proto=https,for=172.12.1.1'`

And transform them like that:

- `'X-Forwarded-By': '0.0.0.0`
- `'X-Forwarded-For': '134.12.14.2, 989.98.98.8'`
- `'X-Forwarded-Host': '123.12.12.12'`
- `'X-Forwarded-Proto': 'https'`

See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

PR #8704, opened 2021-09-24
Manually [Backport 2.4] Add alternative X-Forward headers in proxy  

ger-benjamin:
For GEO-4666

Manual Backport of #8701 (plus some previous PR)

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/+package+/config.yaml.tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/+package+/config.yaml.tmpl
@@ -74,7 +74,6 @@ templates:
             headers:
               - Referer
               - X-Request-ID
-              - Forwarded
           - !restrictUris
             matchers:
               - !localMatch
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/+package+/config.yaml.tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/+package+/config.yaml.tmpl
@@ -74,6 +74,7 @@ templates:
             headers:
               - Referer
               - X-Request-ID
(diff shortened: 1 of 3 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/print/print-apps/+package+/config.yaml.tmpl
```diff
@@ -74,6 +74,7 @@ templates:
             headers:
               - Referer
               - X-Request-ID
+              - Forwarded
```
sbrunner, line 77: I'm not certain that we should add it in the default template ...

ger-benjamin: It comes from https://github.com/camptocamp/c2cgeoportal/commit/4c5ec725bf401aec7322a1fe686ecc8f0401ea6c

Also, I've errors in the test because self.request.client_addr is None. Any idea ?

sbrunner: For me, if we change the default scaffold of a stable release, it should be for a good reason for standard users.
In the commit, you point it was done for the intranet user detection, who was introduced in GMF 2.5...

sbrunner: In version 2.4 we use a version of c2cwsgiutils who's didn't have c2cwsgiutils/client_info.py information...

I don't know what's the impact, but we use an old version of Pyramid...

sbrunner (approved):
Do you test it in your project?
Looks good :-)

ger-benjamin:
Yes, it works with in the BS project. Meaning that we have the client_addr in some cases. I've added an `if` around my change. It should be good in all cases now.

PR #8714, opened 2021-09-28
Don't do WFS query on server that didn't support WFS

sbrunner:
See also: https://sentry.io/organizations/camptocamp/issues/2633701839/?environment=prod-2-6&project=1851011&query=is%3Aunresolved


arnaud-morvan (approved):
LGTM

PR #8716, opened 2021-09-28
Add MVT view

vuilleumierc:
GSGMF-1683

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/vector_tiles.py
+++ b/geoportal/c2cgeoportal_geoportal/views/vector_tiles.py
@@ -39,6 +39,8 @@ LOG = logging.getLogger(__name__)
 
 
 class VectorTilesViews:
+    """All the views concerning the vector tiles."""
+
     def __init__(self, request: Request) -> None:
         self.request = request
 


geoportal/c2cgeoportal_geoportal/views/mb_tiles.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2021-2021, Camptocamp SA
```
sbrunner, line 1:
```suggestion
# Copyright (c) 2021, Camptocamp SA
```

geoportal/tests/functional/test_mb_tiles.py
```diff
@@ -0,0 +1,76 @@
...
+class TestMbTilesViews():
+
+    def test_mb_tiles_success(self, dummy_request, test_data):
+        from c2cgeoportal_geoportal.views.mb_tiles import MbTilesViews
+
+        request = dummy_request
+
+        resp = MbTilesViews(request).mb_tiles()
+
+        assert resp == {}
```
sbrunner, line 76: Shouldn't we add a test that test a real case?

geoportal/tests/config.yaml
```diff
@@ -11,3 +11,7 @@ vars:
       backend: dogpile.cache.memory
     obj:
       backend: dogpile.cache.memory
+  vectortile:
+    srid: 21781
+    extent: [599900, 199950, 600100, 200050]
+    resolutions: [4000, 2000, 1000, 500, 250, 100, 50, 20, 10, 5, 2.5, 1, 0.5, 0.25, 0.1, 0.05]
```
sbrunner, line 17: This should also be added in the `vars.yaml_tmpl`
vectortile should be added as `CONFIG_VARS`
and the structure should be added in the `CONST_config-schema.yaml`

sbrunner (commented):
I lately see that the pull request is completely draft, then I just send the initial comments :-)

PR #8717, opened 2021-09-28
Do not load the project ourself as we cannot access QgsServerSettings

arnaud-morvan:
This allow project to be loaded with bad layers for example.

It seems that QgsServerInterface.serverSettings is not accessible from Python side.

It make a long time I do not understand why projects failed to load even if QGIS_SERVER_IGNORE_BAD_LAYERS=true.

It is because we need to pass QgsServerSettings instance to ConfigCache.project().

Here I get the project from configCache after it has already been loaded by QGIS with appropriated settings.

Note that all AccessControl initialization are not done anymore on first server request but on first project request.


sbrunner:
I probably miss something, but I don't understand what this pull request change (except delayed the initialization)...

arnaud-morvan:
Yes it delay the project loading to QGIS which pass a `QgsServerSettings` object to `QgsConfigCache.project` method.

See our line: https://github.com/camptocamp/c2cgeoportal/blob/227cfb3b9ac74ec641a6b1aefaf606211357c42b/docker/qgisserver/geomapfish_qgisserver/accesscontrol.py#L231

And the corresponding line in QGIS: https://github.com/qgis/QGIS/blob/de53894dabf633f1a50dedfd50155ac81111a673/src/server/qgsserver.cpp#L460

We do not pass QgsServerSettings. It would be easier to pass this object ourself, but it is not accessible from QgsServerInterface in Python.

So I delay retreiving the project in QgsAccessControl overrided method, where it is already loaded by QGIS with correct settings and retrieved from cache.

arnaud-morvan:
Note that we could also not defer the init but only access the project when needed with `QgsConfigCache.instance().project(self.map_file)`.

Maybe through class method, for example:
```
def project():
    return QgsConfigCache.instance().project(self.map_file)
```
And use this in QgsAccessControl overridden method.

arnaud-morvan:
But I do not find any proper way of loading the project ourself.

arnaud-morvan:
Note I would vote for the project getter (`def project(self):`) as it would not have side effect in case the project has changed but the case has not been invalidated.

arnaud-morvan:
I also not that we may not need two classes anymore if creating a getter for ogcserver too.

arnaud-morvan:
@sbrunner : I've created a getter for the QgsProject, this is ready for review.
This fix the settings passed to the project that we cannot retrieve on Python side and allow loading projects with bad layers without errors.

PR #8727, opened 2021-10-01
Add small cache to avoid too any requests

sbrunner:


geoportal/c2cgeoportal_geoportal/lib/caching.py
```diff
@@ -193,9 +193,10 @@ def delete(self, key: str) -> None:
 
 
 class Cache(Enum):
-    NO = 0
-    PUBLIC = 1
+    PUBLIC = 0
+    PUBLIC_NO = 1
```
arnaud-morvan, line 197: Not sure to understand, what does PUBLIC_NO stand for ?
It seems is is for something like: `Cache-Control: public, max-age=10`
Maybe this desserve some comments.

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -195,7 +195,7 @@ def _proto_read(self, layer: "main.Layer") -> FeatureCollection:
 
     @view_config(route_name="layers_read_many", renderer="geojson")  # type: ignore
     def read_many(self) -> FeatureCollection:
-        set_common_headers(self.request, "layers", Cache.NO)
+        set_common_headers(self.request, "layers", Cache.PRIVATE_NO)
```
arnaud-morvan, line 198: Be careful here as we are in editing module, I would have recommended no-store for such views.
But it seems we can use private with 10 seconds as, after save ngeo seems to load altered feature directly from PUT response.

arnaud-morvan (approved):
LGTM, but we might need to be careful with such change.

geoportal/c2cgeoportal_geoportal/lib/caching.py
sbrunner: Comment added :-)

danduk82 (approved):
LGTM

PR #8818, opened 2021-10-25
Config for frontend vector tiles

ger-benjamin:
Part of GSGMF-1684

Options to create the vectortiles layer. Parameters must match `app.vector_tiles` parameters.

### DIFF
--- a/commons/c2cgeoportal_commons/models/main.py
+++ b/commons/c2cgeoportal_commons/models/main.py
@@ -1241,7 +1241,11 @@ class LayerVectorTiles(DimensionLayer):
         info={
             "colanderalchemy": {
                 "title": _("Style"),
-                "description": "The path to json style. Example: https://url/style.json",
+                "description": _(
+                    """
+                    The path to a Mapbox Style file (version 8 or higher). Example: https://url/style.json
+                    """
+                ),
                 "column": 2,
             }
         },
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/vars.yaml_tmpl
```diff
@@ -45,6 +45,12 @@ vars:
           coordinatesProjections: *alternate_projections
         gmfPermalinkOptions:
           projectionCodes: *alternate_projections
+        gmfVectorTilesOptions:
+          projection: EPSG:{srid}
+          tileGrid:
+            extent: *extent
+            origin: [2420000, 1350000]
```
sbrunner, line 52: Why is it needed?

ger-benjamin: To make a correct TileGrid

sbrunner: But it can be extracted from the extent, no?

ger-benjamin: In Openlayers, we can define extent and origin independently. Origins looks to be the point 0/0/z of the grid, and extent a mask to not display the tiles outside it.
I've kept the original Openlayers TileGrid possibilities. Eventually, I can extract the origins from the extent if the origin is not defined. But I was keeping this poc as low budget as possible.

PR #8839, opened 2021-11-04
Be able to eazily add a new route in haproxy

sbrunner:


arnaud-morvan (approved):
LGTM, we can override the haproxy config, but default config still comes from base config image.

PR #8846, opened 2021-11-06
Filter metadatas in admin forms

arnaud-morvan:
Still some doubt about which functionality is relavant for themes or roles.

TODO: complete CONST_vars with relevant_for in available_functionalities.


sbrunner (approved):
Looks good except that the relevant_for are missing in the CONST_vars.yaml file :-)

sbrunner:
Also metadata :-)

sbrunner:
Thanks :-)

PR #8852, opened 2021-11-08
Add Referer to the vary to fix the Varnish cache

sbrunner:


arnaud-morvan (approved):
LGTM
Maybe a comment line to explain why response depends on referer ?

PR #8885, opened 2021-11-12
Fix code verification in oauth2

sbrunner:
Changes:
- Add some logs
- Change the result of `client_authentication_required`, the consequence is that we use the function `authenticate_client_id` instead of `authenticate_client`.
- `authenticate_client` => unimplemented


arnaud-morvan (approved):
Trust you on this

PR #8907, opened 2021-11-16
Add information about default action in the full-text search

jwkaltz:


jwkaltz:
@ger-benjamin I think this is correct?

sbrunner (approved):

ger-benjamin (commented):
It's correct, thanks

PR #8937, opened 2021-11-21
Add tests for lingua extractors

arnaud-morvan:


sbrunner (approved):
Looks good but you should pass black :-)

PR #9004, opened 2021-12-07
Add action to update the po files

sbrunner:


arnaud-morvan (commented):
For me it is not clear to have to scripts:
- update-po which update pot and po files
- update-po-only which update only po files.
Without any description inside scripts.

I could propose to create two scripts:
- update-pot which update the pot file.
- update-po which update the po files.

Can be dicussed after sprint planning.

sbrunner:
Updated :-)

bin/update-po
```diff
@@ -21,6 +21,11 @@
 # IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 # CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+# Will update the pot file then the po files.
+# Used when we update the po files in the application runninig in a docker-compose locally.
```
arnaud-morvan, line 25:
```suggestion
# Used when we update the po files in the application running in a docker-compose locally.
```

arnaud-morvan (approved):

PR #9063, opened 2021-12-21
Fix l10n action

sbrunner:
Fix working directory
Initialize Git
Commit and create a pull request only if there is some changes
Fix GITHUB_REPOSITORY
No error if the pull request already exists
Remove the POT-Creation-Date
Add missing SUFFIX
Tools don't depend on config by default

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/.github/workflows/update_l10n.yaml
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/.github/workflows/update_l10n.yaml
@@ -27,7 +27,15 @@ jobs:
           token: ${{'{{'}} secrets.GITHUB_GOPASS_CI_TOKEN {{'}}'}}
 
       - run: ./build --env
-      - run: PROJECT_PUBLIC_URL={{'{{'}} matrix.base_url {{'}}'}} make -e update-po-url
+      - run: PROJECT_PUBLIC_URL=${{'{{'}} matrix.base_url {{'}}'}} make -e update-po-from-url
+
+      - name: Init Git
+        run: |
+          git config --global user.email "geospatial-bot@camptocamp.com"
+          git config --global user.name CI
+
+      - id: status
+        run: echo ::set-output name=status::$(git status --porcelain)
 
       - name: Push l10n branch and create pull request
         run: |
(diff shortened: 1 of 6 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/docker-compose-lib.yaml
```diff
@@ -285,7 +285,7 @@ services:
     image: camptocamp/geomapfish-tools:${GEOMAPFISH_VERSION}
     restart: unless-stopped
     volumes:
-      - ./geoportal/${PACKAGE}_geoportal/locale:/app/${PACKAGE}_geoportal/locale
+      - ./geoportal:/app/geoportal
```
arnaud-morvan, line 288: This seems wreid to me.
In tools container, I'm use to have installed "package"_geoportal package in /app (useful for debugging as we can substitute geoportal by tools and we have more dev tools).
And all sources in /src/ for development tasks (generate alembic revisions, update po files, etc).

arnaud-morvan (commented):
Looks good, but I do not understand why we do not keep mounting sources in /src.
Maybe because you have only one last WORKDIR.
I'm affraid we will be limited at sometime with mounting only geoportal in /app.
And I do not understand why we do not keep mounting it by default, now we need an manual override for that.

geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/.github/workflows/update_l10n.yaml
```diff
@@ -27,7 +27,15 @@ jobs:
           token: ${{'{{'}} secrets.GITHUB_GOPASS_CI_TOKEN {{'}}'}}
 
       - run: ./build --env
-      - run: PROJECT_PUBLIC_URL={{'{{'}} matrix.base_url {{'}}'}} make -e update-po-url
+      - run: PROJECT_PUBLIC_URL=${{'{{'}} matrix.base_url {{'}}'}} make -e update-po-url
```
arnaud-morvan, line 30:
```suggestion
      - run: PROJECT_PUBLIC_URL=${{'{{'}} matrix.base_url {{'}}'}} make -e update-po-from-url
```

arnaud-morvan (approved):

PR #9069, opened 2021-12-23
Fix QGIS plugin initialisation logs

sbrunner:
The print isn't displayed
Using logging before logging plugin initialization don't works


arnaud-morvan (approved):
LGTM

PR #9071, opened 2021-12-24
Move the env to the .env file to avoid warning

sbrunner:
Like:
WARNING: The GIT_HASH variable is not set. Defaulting to a blank string.
WARNING: The SIMPLE variable is not set. Defaulting to a blank string.


arnaud-morvan (approved):
LGTM

PR #9085, opened 2022-01-03
Make the timeout in theme configurable

sbrunner:
GSGMF-1694

### DIFF
--- a/doc/integrator/docker.rst
+++ b/doc/integrator/docker.rst
@@ -61,22 +61,22 @@ Use ``docker system prune`` to clean files; you can add the ``--all`` option to 
 Environment variables
 ---------------------
 
-The GeoMapFish containers can be customised with some environment variables:
+The GeoMapFish containers can be customized with some environment variables:
 
 Config:
 
- * ``VISIBLE_WEB_HOST``: The web host visible by the browser e.-g.: 'example.com'.
- * ``VISIBLE_ENTRY_POINT``: The web path visible by the browser e.-g.: '/main/', default is '/'.
- * ``PGSCHEMA``: The postgres schema, used by mapserver.
- * ``GEOPORTAL_INTERNAL_URL``: Used by the print in non mutualised mode.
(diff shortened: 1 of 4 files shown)


doc/integrator/docker.rst
```diff
@@ -97,6 +97,8 @@ Geoportal:
  * ``SQL_LOG_LEVEL``: Log level for the Sqlalchemy engine``, default is ``WARN``.
  * ``DOGPILECACHE_LOG_LEVEL``: Log level for Dogpile cache, default is ``INFO``.
  * ``OTHER_LOG_LEVEL``: Log level for other modules, default is ``WARN``.
+ * ``C2CGEOPORTAL_THEME_TIMEOUT``: Timeout for the theme generation, used on requests to get information
```
arnaud-morvan, line 100: Timeout in seconds used in requests sent to OGC servers during themes generation, default to ``300``.

arnaud-morvan (approved):

doc/integrator/docker.rst
```diff
@@ -87,34 +87,36 @@ Geoportal:
  * ``AUTHTKT_SAMESITE``: See: :ref:`integrator_authentication`.
  * ``BASICAUTH``: See: :ref:`integrator_authentication`.
  * ``LOG_TYPE``: Should be 'console' with Docker Compose and 'json' with OpenShift.
- * ``LOG_LEVEL``: Log level for your application, default is ``INFO``, can be
+ * ``LOG_LEVEL``: Log level for your application, default to ``INFO``, can be
```
jwkaltz, line 90: This change is grammatically incorrect, either "default is", or "defaults to"
Same for the following

sbrunner: Then this https://ludwig.guru/s/default+to+something is wrong, or I don't understand something...

jwkaltz: The example https://ludwig.guru/s/default+to+something contains a verb within the formulation "it'll" (correct language would be "it will"), so it's a different structure than our sentence.

If we want to use "to" like in the example from ludwig, we would also need a verb, for example
default is set to
or
defaults to

PR #9088, opened 2022-01-03
Fix mail sending when ssl is False

jwkaltz:
I think the current code for mail sending has a bug when ssl is set to False. It uses SMTP_SSL when in fact ssl should not be used.
The proposed code change was tested successfully on a GMF 2.6 instance where ssl is set to False.

### DIFF
--- a/commons/c2cgeoportal_commons/lib/email_.py
+++ b/commons/c2cgeoportal_commons/lib/email_.py
@@ -77,11 +77,14 @@ def send_email(
     msg.attach(MIMEText(body, "plain", "utf-8"))
 
     # Connect to server
-    smtp = smtplib.SMTP(smtp_config["host"])
     if smtp_config.get("ssl", False):
-        smtp.ehlo()
+        smtp: smtplib.SMTP = smtplib.SMTP_SSL(smtp_config["host"])
+    else:
+        smtp = smtplib.SMTP(smtp_config["host"])
     if smtp_config.get("starttls", False):
         smtp.starttls()
+    else:
+        smtp.ehlo()
     if smtp_config.get("user", False):
         smtp.login(smtp_config["user"], smtp_config["password"])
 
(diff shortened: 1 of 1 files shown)


arnaud-morvan (changes requested):
if SSL is false: this does not change anything
if SSL is true: this remove use of SSL

Are you sure you need to change something, or are you sure SSL evaluate to true in your GMF config.

sbrunner (changes requested):
I agree with @arnaud-morvan

jwkaltz:
Yes, I mixed up somethings in my pull request, sorry... 
I will re-test and change the pull request (or close it).

jwkaltz:
@arnaud-morvan here is proposal which hopefully makes more sense; and here the results from my test run on the server where it is not working (where no starttls is used):
[test-run-current-code.txt](https://github.com/camptocamp/c2cgeoportal/files/7807678/test-run-current-code.txt)
[test-run-code-from-pr-new-version.txt](https://github.com/camptocamp/c2cgeoportal/files/7807680/test-run-code-from-pr-new-version.txt)

commons/c2cgeoportal_commons/lib/email_.py
```diff
@@ -83,6 +83,8 @@ def send_email(
         smtp = smtplib.SMTP(smtp_config["host"])
     if smtp_config.get("starttls", False):
         smtp.starttls()
+    else:
+        smtp.ehlo()
```
sbrunner, line 87: For me, it's strange to do that only if we don't use TLS, do you have some information about this condition or should it be like:
```suggestion
    if smtp_config.get("ehlo", True):
        smtp.ehlo()
```
And add `ehlo` definition in the `CONT_schema.yaml`.
What do you think?

arnaud-morvan:
Sorry, but I do not understand.

The SMTPLib doc clearly state that sendmail automatically call ehlo if needed when calling sendmail.

And it is the case, it is the first thing done in sendmail and it fails parsing the response.

But it seems to pass when you call it manually first.

jwkaltz:
As debugged with @arnaud-morvan , the problem in this installation is somewhere else, the PR would not change that.

PR #9094, opened 2022-01-05
Fix 'default to' to 'default is'

sbrunner:
@jwkaltz is that correct?

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/mapserver/mapserver.map.tmpl
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/mapserver/mapserver.map.tmpl
@@ -22,7 +22,7 @@ MAP
     EXTENT {{cookiecutter.extent_mapserver}} ## should be defined for better performance
     UNITS METERS
 
-    # RESOLUTION and DEFRESOLUTION default to 96. If you
+    # RESOLUTION and DEFRESOLUTION default is 96. If you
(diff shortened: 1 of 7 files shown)


doc/conf.py
```diff
@@ -114,7 +114,7 @@
 # Add any paths that contain custom themes here, relative to this directory.
 # html_theme_path = []
 
-# The name for this set of Sphinx documents.  If None, it default to
+# The name for this set of Sphinx documents.  If None, it default is
```
jwkaltz, line 117: If none, the default is
or
If none, default is
or
If none, it defaults to

geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/mapserver/mapserver.map.tmpl
```diff
@@ -22,7 +22,7 @@ MAP
     EXTENT {{cookiecutter.extent_mapserver}} ## should be defined for better performance
     UNITS METERS
 
-    # RESOLUTION and DEFRESOLUTION default to 96. If you
+    # RESOLUTION and DEFRESOLUTION default is 96. If you
```
jwkaltz, line 25: here, the "default to" was correct, because of the plural and therefore "default" can be read as a verb
So:
RESOLUTION and DEFRESOLUTION default to 96
or
RESOLUTION and DEFRESOLUTION have as default value 96
or
For RESOLUTION and DEFRESOLUTION, the default is

jwkaltz (approved):

PR #9109, opened 2022-01-06
Workaround for setuptools >= 60.0.0

sbrunner:
See https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6000
https://github.com/pypa/setuptools/issues/2996


sbrunner:
Don't look to be needed anymore, waiting...

ochriste (approved):

PR #9177, opened 2022-01-20
Fix qgisserver

sbrunner:


docker/qgisserver/Dockerfile
```diff
@@ -7,7 +7,8 @@ COPY requirements.txt /tmp/
 RUN python3 -m pip install --disable-pip-version-check --no-cache-dir --requirement=/tmp/requirements.txt
 
 COPY python/ /tmp/python/
-RUN cd /tmp/python/$(python3 --version|awk '{print $2}'|awk -F '.' '{print $1 "." $2}') && \
+RUN python3 -m pip uninstall --yes Shapely && \
```
arnaud-morvan, line 10: Why do we need to remove Shapely ?

arnaud-morvan (approved):

docker/qgisserver/Dockerfile
sbrunner: To force reinstalling it...

sbrunner: I should have a look on qgisserver...

PR #9294, opened 2022-02-11
Fix oauth

sbrunner:
The auth secret can be in Authorization header
See also: https://jira.camptocamp.com/browse/GEO-5217
https://github.com/fkooman/php-oauth2-client/blob/90129b280133309dead3318b03ce2e775452026d/src/OAuthClient.php#L293-L302

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/oauth2.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/oauth2.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2021, Camptocamp SA
+# Copyright (c) 2021-2022, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


geoportal/c2cgeoportal_geoportal/lib/oauth2.py
```diff
@@ -109,12 +110,24 @@ def authenticate_client_id(  # pylint: disable=no-self-use,useless-suppression
...
-        params = dict(request.decoded_body)
+        client_secret = None
+        if "Authorization" in request.headers and request.headers["Authorization"].startswith("Basic "):
+            decode = base64.b64decode(request.headers["Authorization"][6:]).decode()
+            username, password = decode.split(":")
+            if username == client_id:
+                client_secret = password
+            else:
+                LOG.info(
+                    "Authorization username '%s' differ to client_id '%s', ignoring.", username, client_id
```
arnaud-morvan, line 121:
```suggestion
                    "Authorization username '%s' differ from client_id '%s', ignoring.", username, client_id
```

arnaud-morvan (approved):

PR #9317, opened 2022-02-16
Add default values to avoid warning

sbrunner:
Current warning:
```
The key 'starttls' in '.smtp' is not present in: ['host', 'password', 'ssl', 'user']
```


arnaud-morvan (approved):
Don't we need to add `smtp` in `update_paths` in `vars.yaml`

PR #9353, opened 2022-02-28
We shouldn't be able to do user enumeration

sbrunner:


danduk82 (approved):
So if I get this correctly, you move the login-stuff validation before we enter in the user credentials validation. LGTM

arnaud-morvan (approved):

PR #9368, opened 2022-03-02
Remove default value for C2C_SECRET

arnaud-morvan:


ochriste:
@arnaud-morvan by setting the value to null, doesnt that mean we can login without password ?

doc/integrator/c2cwsgiutils.rst
```diff
@@ -12,4 +12,6 @@ c2cwsgiutils and c2cgeoportal services.
 
 To use the protected tools, you should at least set ``C2C_SECRET`` to a secret password in the ``env.project`` file.
 
+Note that there is no real protection against brute-force attack on those tools.
```
jwkaltz, line 15: I really don't think we should state that so explicitly in a documentation available on the Internet

jwkaltz (changes requested):

arnaud-morvan:
In c2cwsgiutils v4, with empty string, we can only logout:
https://github.com/camptocamp/c2cwsgiutils/blob/release_4/c2cwsgiutils/auth.py#L37:L40

In c2cwsgiutils v5, with empty string, we do not use

sbrunner (approved):

doc/integrator/c2cwsgiutils.rst
arnaud-morvan: Maybe we can move it to env.project, before C2C_SECRET line, but it will still be in a public code repository.
@sbrunner also talked about https://camptocamp.github.io/c2cgeoportal/master/integrator/security.html

arnaud-morvan: Ok removed.

sbrunner (approved):

jwkaltz (approved):

PR #9378, opened 2022-03-04
Upgrade to `debugpy`

RemiDesgrange:
`ptvsd` is deprecated, I updated the doc accordingly.

### DIFF
--- a/doc/developer/debugging.rst
+++ b/doc/developer/debugging.rst
@@ -144,31 +144,33 @@ Be sure that the volumes for c2cgeoportal are uncommented.
 Remote debugging using Visual Studio Code
 .........................................
 
-* In ``geoportal/requirements.txt`` uncomment ``ptvsd``.
+* In ``geoportal/requirements.txt`` uncomment ``debugpy``.
 * In the code add ``breakpoint()`` where you want to add a breakpoint.
 * In Visual Studio Code use the config:
 
   .. code::
+    {
+        "configurations": [
+            {
+                "name": "Python: Remote Attach",
+                "type": "python",
+                "request": "attach",
+                "connect": {
+                    "host": "localhost",
+                    "port": 5678
+                },
+                "pathMappings": [
+                    {
+                        "localRoot": "${workspaceFolder}",
+                        "remoteRoot": "/app/geoportal"
+                    }
+                ]
+            }
+        ]
+    }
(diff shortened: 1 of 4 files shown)


sbrunner (approved):

doc/developer/debugging.rst
```diff
@@ -144,31 +144,33 @@ Be sure that the volumes for c2cgeoportal are uncommented.
 Remote debugging using Visual Studio Code
 .........................................
 
-* In ``geoportal/requirements.txt`` uncomment ``ptvsd``.
+* In ``geoportal/requirements.txt`` uncomment ``debugpy``.
 * In the code add ``breakpoint()`` where you want to add a breakpoint.
 * In Visual Studio Code use the config:
 
   .. code::
+    {
```
sbrunner, line 152: Isn't a new line missing after code:: ?

doc/developer/debugging.rst
```diff
@@ -144,31 +144,34 @@ Be sure that the volumes for c2cgeoportal are uncommented.
...
+                "type": "python",
+                "request": "attach",
+                "connect": {
+                    "host": "localhost",
+                    "port": 5678
+                },
+                "pathMappings": [
+                    {
+                        "localRoot": "${workspaceFolder}",
+                        "remoteRoot": "/app/geoportal"
```
arnaud-morvan, line 166: For me it should be:
```
                        "localRoot": "${workspaceFolder}/geoportal",
                        "remoteRoot": "/app"
```
or
```
                        "localRoot": "${workspaceFolder}/{project_folder}/geoportal",
                        "remoteRoot": "/app"
```
As we copy folder `/geoportal` in `/app`: https://github.com/camptocamp/c2cgeoportal/blob/2.6/geoportal/c2cgeoportal_geoportal/scaffolds/create/geoportal/Dockerfile_tmpl#L9

Note that only `/geoportal/+package+_geoportal` is interesting in reality (this is what we mount in override), but we do not know the package name here, and this does not matter for VS Code path mapping.

arnaud-morvan (approved):

doc/developer/debugging.rst
RemiDesgrange: Sounds weird but works on my machine ¯\_(ツ)_/¯

PR #9400, opened 2022-03-09
Json handler stream is optional

arnaud-morvan:


sbrunner:
@arnaud-morvan is it steel needed?

arnaud-morvan:
Yes I think it is good to merge, for consistency.
I've fixed it and rebased on master.

sbrunner (approved):

PR #9687, opened 2022-05-11
Fix error 500 on wrong authentication

sbrunner:


arnaud-morvan:
Seems you forgot some replacements: https://github.com/camptocamp/c2cgeoportal/search?q=loginform403

arnaud-morvan:
Seems you forgot some replacements: https://github.com/camptocamp/c2cgeoportal/search?q=loginform403

sbrunner:
@arnaud-morvan thanks, it's an unwanted function renaming, back to the old function name :-)

arnaud-morvan (approved):

PR #9691, opened 2022-05-12
Fix change password

sbrunner:
Don't get the user in the request cache


geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -324,17 +320,16 @@ def change_password(self) -> pyramid.response.Response:
         else:
             user = self.request.user
 
-        username = self.request.registry.validate_user(self.request, user.username, old_password)
-        if username is None:
-            LOG.info("The old password is wrong for user '%s'.", username)
+        if self.request.registry.validate_user(self.request, user.username, old_password) is None:
+            LOG.info("The old password is wrong for user '%s'.", user.username)
```
arnaud-morvan, line 324: Shouldn't we return standard response here for security.
=> do not give the hacker the information that the username exists

arnaud-morvan (approved):
LGTM, maybe we can improve security.

geoportal/c2cgeoportal_geoportal/views/login.py
sbrunner: The visible response is on the next line and is completely generic :-)
Or I miss something...

arnaud-morvan: Right sorry

PR #9723, opened 2022-05-17
Add default value

jwkaltz:


sbrunner (changes requested):
This is not a default value but a value, and I'm not completely confident with this...

jwkaltz:
> This is not a default value but a value, and I'm not completely confident with this...

Is this a better place for the default value?

sbrunner (approved):

PR #9730, opened 2022-05-18
Add documentation to mapserver_substitution functionality

sbrunner:
@arnaud-morvan do you know if the HTML working here?

### DIFF
--- a/admin/c2cgeoportal_admin/lib/lingua_extractor.py
+++ b/admin/c2cgeoportal_admin/lib/lingua_extractor.py
@@ -26,6 +26,7 @@
 # either expressed or implied, of the FreeBSD Project.
 
 
+import os
 from typing import Any, Dict, List, Optional
 
 import yaml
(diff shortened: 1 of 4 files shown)


arnaud-morvan:
I suppose it will be escaped: https://github.com/camptocamp/c2cgeoportal/blob/6b783b51757085b3ef7502887ec7aebbab2aa494/admin/c2cgeoportal_admin/templates/widgets/functionality_fields.pt#L43

And https://api.jquery.com/text/#text-text says "We need to be aware that this method escapes the string provided as necessary so that it will render correctly in HTML."

And here is what I get in admin:
```
<p class="help-block">MapServer substitution of a variable in the mapfile. See the &lt;a href="https://camptocamp.github.io/c2cgeoportal/{{cookiecutter.geomapfish_main_version}}/administrator/mapfile.html?highlight=mapserver_substitution#variable-substitution"&gt;documentation&lt;/a&gt;</p>
```

So no it seems not possible right now.

arnaud-morvan (approved):
Consider adding it also in `admin/config.yaml` and `admin/tests/tests.yaml`

PR #9732, opened 2022-05-18
Fix tinyowsproxy

arnaud-morvan:
Note that I also fixed some types in proxy classes.


sbrunner (approved):
Thanks

sbrunner:
The SonarCloud check should be ignored

PR #9749, opened 2022-05-20
Add missing volume for update-po

sbrunner:
Fix GSGMF-1854


geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/docker-compose.yaml
```diff
@@ -109,6 +109,8 @@ services:
 
   # Rich image for project development with e.-g. vim, tree, awscli, psql, ...
   tools:
+    volumes:
+      - .:/src
```
arnaud-morvan, line 113: This volume is already present in `docker-compose-lib.yaml`
Do I miss something ?

sbrunner: Arg :-/

PR #9889, opened 2022-06-28
Fix the editing

sbrunner:
The layer.exclude_properties should always be used to don't break the
Dogpile cache, and create tow SQLAlchemy Table on the same database table.


ochriste (approved):

arnaud-morvan (approved):
LGTM

PR #9908, opened 2022-07-04
Use the new Transifex client

sbrunner:


Dockerfile
```diff
@@ -94,6 +94,11 @@ RUN --mount=type=cache,target=/var/lib/apt/lists \
 RUN --mount=type=cache,target=/root/.cache \
     --mount=type=bind,from=poetry,source=/tmp,target=/poetry \
     python3 -m pip install --disable-pip-version-check --no-deps --requirement=/poetry/requirements-dev.txt
+ENV PATH=/root/.local/bin/:${PATH}
+
+WORKDIR /opt/c2cgeoportal
+COPY ci/applications*.yaml .
+RUN c2cciutils-download-applications --applications-file=applications.yaml --versions-file=applications-versions.yaml
```
arnaud-morvan, line 101: Here you install an application from `ci/applications.yaml` using `c2cciutils` in `~/.local/bin` (probably in user root) in a Docker image that may be runned on development, integration and production environments.

I was thinking c2cciutils-download-application was only for usage on CI host.

Hope this application will not be used at run time.

dependencies.mk
```diff
@@ -29,26 +28,17 @@ dependencies: $(TRANSIFEX_PO_FILES)
 $(HOME)/.transifexrc:
 	mkdir --parent $(dir $@)
 	echo "[https://www.transifex.com]" > $@
+	echo "api_hostname  = https://api.transifex.com" >> $@
+	echo "rest_hostname = https://rest.api.transifex.com" >> $@
 	echo "hostname = https://www.transifex.com" >> $@
-	echo "username = c2c" >> $@
-	echo "password = c2cc2c" >> $@
-	echo "token =" >> $@
+	echo "token = 1/dc02578696187cc29e4e6486f8611fdbfe60b235" >> $@
```
arnaud-morvan, line 34: Should we commit this token in a public repository.

doc/developer/build_release.rst
```diff
@@ -88,25 +88,27 @@ Use the ``ngeo`` package linked to the new branch
...
 .. prompt:: bash
 
-    docker build --target=tools --tag=transifex --build-arg=MAJOR_VERSION=x.y .
-    docker run --name=transifex -ti --rm --detach --volume=${HOME}:/root transifex tail -f /dev/null
-    docker exec transifex bash -c \
-        '(cd /opt/c2cgeoportal; sed -i s/x_y/x_y+1/g .tx/config)'
-    docker exec --env=MAJOR_VERSION=x.y+1 transifex bash -c \
-        '(cd /opt/c2cgeoportal; make --makefile=dependencies.mk transifex-init)'
-    docker stop transifex
+    tx pull --branch=<version> --source --force \
```
arnaud-morvan, line 98: can this be executed on host, shouldn't we use tools container to run those commands

arnaud-morvan (commented):
Some things seems wreid to me, but maybe I miss some things.

dependencies.mk
sbrunner: This user didn't has any rights, it's just because Transifex just works with connected user ...

doc/developer/build_release.rst
sbrunner: Yes, we can we just need to have the Transifex client and be logged in, but it's true that it we do that in Docker it's possible that we have less issues...

Dockerfile
sbrunner: Here we are in the tools stage => tools image, then it will not be in the running containers :-)

arnaud-morvan (approved):

PR #9911, opened 2022-07-06
Have a separate cache for the OGC servers

sbrunner:
Tested on demo, at home throw VPN on my slow new computer
GSLUX-460

Before:
First: 33.6s
Second: 0.7s
After Layer change: 34.6s
After OGC server change: 32.4s

After:
First: 43.6s
Second: 0.9s
After Layer change: 4.1s
After OGC server change: 4.1s (But we should wait that the background task ended, about 20s)

### DIFF
--- a/admin/c2cgeoportal_admin/views/ogc_servers.py
+++ b/admin/c2cgeoportal_admin/views/ogc_servers.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2021, Camptocamp SA
+# Copyright (c) 2017-2022, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
@@ -26,9 +26,12 @@
 # either expressed or implied, of the FreeBSD Project.
 
 
+import logging
+import threading
 from functools import partial
 from typing import Any, Dict, List, cast
 
+import requests
 from c2cgeoform.schema import GeoFormSchemaNode
 from c2cgeoform.views.abstract_views import AbstractViews, ItemAction, ListField, UserMessage
 from deform.widget import FormWidget
@@ -39,12 +42,15 @@ from sqlalchemy import inspect
 from c2cgeoportal_admin import _
 from c2cgeoportal_admin.lib.ogcserver_synchronizer import OGCServerSynchronizer
 from c2cgeoportal_commons.lib.literal import Literal
+from c2cgeoportal_commons.models import cache_invalidate_cb
 from c2cgeoportal_commons.models.main import OGCServer
 
 _list_field = partial(ListField, OGCServer)
 
 base_schema = GeoFormSchemaNode(OGCServer, widget=FormWidget(fields_template="ogcserver_fields"))
 base_schema.add_unique_validator(OGCServer.name, OGCServer.id)
+
+LOG = logging.getLogger(__name__)
 
 
 @view_defaults(match_param="table=ogc_servers")
@@ -110,6 +116,22 @@ class OGCServerViews(AbstractViews):  # type: ignore
             actions.insert(
                 next((i for i, v in enumerate(actions) if v.name() == "delete")),
                 ItemAction(
+                    name="clear-cache",
+                    label=_("Clear the cache"),
+                    icon="glyphicon glyphicon-hdd",
+                    url=self._request.route_url(
+                        "ogc_server_clear_cache",
+                        id=getattr(item, self._id_field),
+                        _query={
+                            "came_from": self._request.current_route_url(),
+                        },
+                    ),
+                    confirmation=_("The current changes will be lost."),
+                ),
+            )
+            actions.insert(
+                next((i for i, v in enumerate(actions) if v.name() == "delete")),
+                ItemAction(
                     name="synchronize",
                     label=_("Synchronize"),
                     icon="glyphicon glyphicon-import",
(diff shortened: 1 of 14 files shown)


admin/c2cgeoportal_admin/views/ogc_servers.py
```diff
@@ -128,7 +150,24 @@ def view(self) -> Dict[str, Any]:
...
-        return super().save()  # type: ignore
+        result: Dict[str, Any] = super().save()
+        try:
+            import c2cgeoportal_geoportal.views.theme  # pylint: disable=import-outside-toplevel
+
+            ogc_server = self._get_object()
+            ogc_server_id = ogc_server.id
+
+            self._request.user = None
+            theme_view = c2cgeoportal_geoportal.views.theme.Theme(self._request)
```
arnaud-morvan, line 161: Actually admin should not depends on c2cgeoportal_geoportal.

admin/c2cgeoportal_admin/views/ogc_servers.py
```diff
@@ -188,4 +227,21 @@ def synchronize(self) -> Dict[str, Any]:
...
+
         return {}
+
+    @view_config(route_name="ogcserver_clear_cache", renderer="json")  # type: ignore
+    def ogcserver_clear_cache_view(self) -> Dict[str, Any]:
+
+        self._update_cache(DBSession.query(OGCServer).filter_by(id=self.request.params.get("id")).one())
+        came_from = self.request.params.get("came_from")
+        if came_from:
+            raise pyramid.httpexceptions.HTTPFound(location=came_from)
```
arnaud-morvan, line 240:
```suggestion
            return pyramid.httpexceptions.HTTPFound(location=came_from)
```

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -860,11 +861,6 @@ def url_wfs_description(self, request: pyramid.request.Request) -> Optional[str]
         return url.url() if url else "\n".join(errors)
 
 
-event.listen(OGCServer, "after_insert", cache_invalidate_cb, propagate=True)
-event.listen(OGCServer, "after_update", cache_invalidate_cb, propagate=True)
-event.listen(OGCServer, "after_delete", cache_invalidate_cb, propagate=True)
```
arnaud-morvan, line 865: As far as I see, delete is not handled in admin views.

geoportal/c2cgeoportal_geoportal/views/theme.py
```diff
@@ -1027,8 +1026,8 @@ def _get_features_attributes_cache(
                     attributes[name] = types[type_]
                 elif (type_ == "Character") and (name + "Type") in types:
                     LOG.debug(
-                        "Due MapServer strange result the type 'ms:Character' is fallbacked to type '%sType'"
-                        " for feature '%s', This is a strange comportement of MapServer when we use the "
+                        "Due MapServer strange result the type 'ms:Character' is fall backed to type '%sType'"
+                        " for feature '%s', This is a strange comportment of MapServer when we use the "
```
arnaud-morvan, line 1030:
```suggestion
                        " for feature '%s', This is a wreid behavior of MapServer when we use the "
```

arnaud-morvan (commented):
It seems really difficult to not depends on theme_view in admin.
But this will block us it at some time we want to put admin in another container.

arnaud-morvan:
It might be interesting having some tests, available for discussion on this.

.prospector.yaml
```diff
@@ -37,6 +37,7 @@ pylint:
   disable:
     - missing-module-docstring
     - missing-function-docstring
+    - cyclic-import
```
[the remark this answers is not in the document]
sbrunner: It detects cyclic import like that:
File1:
```
import file2
```
File 1:
```
def action:
  import file1
```
And this is not an issue...

admin/c2cgeoportal_admin/views/ogc_servers.py
sbrunner: You're sure that's working with the renderer?

.prospector.yaml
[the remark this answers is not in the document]
arnaud-morvan: Yes, but IMHO if we should avoid something, the first one is the dependency loops.

Would like to know why we need this, and if we can get around this.

IMHO we should never create dependency loops as they may be really hard to understand and also to solve.

I do not really like to see this in a PR.

So I say Humm.

Maybe we can add a comment for this case, but I'm pretty sure it is not a good idea.

admin/c2cgeoportal_admin/views/ogc_servers.py
arnaud-morvan: Note that pyramid.httpexceptions.HTTPFound inherit from Response class.
And this is not an error so I would like to avoid `raise` in this case.
And errors are not correctly handled by c2cwsgiutils, mainly concerning CORS.

arnaud-morvan:
Apparently this PR still contains some anti-patterns that I would like to resolve before merge.
Available for discussion.

.prospector.yaml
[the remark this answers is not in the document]
arnaud-morvan: ok for an override in tests or maybe in file (after dicussion) but not for global.

sbrunner:
@arnaud-morvan ready for a final review :-)

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -666,6 +667,8 @@ def add_static_route(name: str, attr: str, path: str, renderer: str) -> None:
     # Used memory in caches
     config.add_route("memory", "/memory", request_method="GET")
 
+    config.add_route("ogc_server_clear_cache", "clear-ogc-server-cache/{id}", request_method="GET")
```
arnaud-morvan, line 670: I do not see any view registered for this route.

arnaud-morvan (approved):
Seems the view to refresh OGC Server cache is missing.
Feel free to merge after fix.

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -471,7 +471,8 @@ def includeme(config: pyramid.config.Configurator) -> None:
             @zope.event.classhandler.handler(InvalidateCacheEvent)  # type: ignore
```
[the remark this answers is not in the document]
sbrunner: Good catch :-)

PR #9912, opened 2022-07-06
Faster fast reload

sbrunner:
With that, `./build --no-pull --fast-reload` takes about 23s.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/build
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/build
@@ -138,15 +138,17 @@ def main() -> None:
             sys.exit(e.returncode)
 
     if args.fast_reload:
-        for service in (
-            subprocess.run(
+        services = [
+            service
+            for service in subprocess.run(
                 ["docker-compose", "ps", "--services", "--all"], stdout=subprocess.PIPE, check=True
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/build
```diff
@@ -138,15 +138,17 @@ def main() -> None:
...
             )
             .stdout.decode()
             .splitlines()
-        ):
-            if not service.startswith("redis"):
-                run(args, ["docker-compose", "rm", "--stop", "--force", service], check=True)
+            if not service.startswith("redis")
+        ]
+
+        run(args, ["docker-compose", "rm", "--stop", "--force"] + services, check=True)
```
arnaud-morvan, line 151: You do not remove volumes, not sure this will really update the config.

For example I'm used to run:
```
docker-compose rm --stop -f -v config && docker-compose up -d --force-recreate geoportal
```

arnaud-morvan (approved):

geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/build
sbrunner: Yes, it updates the config :-)

PR #9951, opened 2022-08-09
Add missing Vary on Cookie for private query

sbrunner:
Fix: GSGMF-1888


sbrunner:
@arnaud-morvan can you test it on your host?

geoportal/c2cgeoportal_geoportal/lib/caching.py
```diff
@@ -187,6 +187,10 @@ class Cache(Enum):
     PRIVATE = 2
     # See PUBLIC_NO and PRIVATE
     PRIVATE_NO = 3
+    # Sane as PUBLIC but if we are authenticated we are private
+    PUBLIC_CAN_PRIVATE = 4
```
arnaud-morvan, line 191: IMHO
```suggestion
    PUBLIC_OR_PRIVATE = 4
```

arnaud-morvan (approved):
Complicated to test locally.

I tested with Firefox, with additionnal host in /etc/hosts + adding allowed referer and increasing max-age to 100 and everything works as expected.

I did not manage to reproduce with Chrome and local demo.

geoportal/c2cgeoportal_geoportal/lib/caching.py
sbrunner: For me, the deepening is
PUBLIC but the same URL can be private
=> CAN, not OR...
@arnaud-morvan does is looks good to you?

arnaud-morvan: Here we use a qualifier of the reponse content.

so it **can** be PUBLIC **OR** PRIVATE, depending on user authentication.

public can private => private is not verb, so it is not grammatically correct

The response content is "PUBLIC", "PRIVATE", or "PUBLIC_OR_PRIVATE", but it cannot be "PUBLIC_CAN_PRIVATE".

arnaud-morvan: or it can be PRIVATE_IF_AUTHENTICATED or PUBLIC_IF_ANONYMOUS

sbrunner: The same view can return 
`PUBLIC_CAN_PRIVATE` or `PRIVATE` (depending on user authentication decided by the view).
Then `PUBLIC_OR_PRIVATE` is `PUBLIC` (except for the `Vary`) then it not a good name...
What about `PUBLIC_CAN_BE_PRIVATE`?

arnaud-morvan: In fact I'm not sure about why we need this, in case PRIVATE => if user is not authenticated, why not returning same as PUBLIC ?

PUBLIC => response content does not depend on authentication, even if user is authenticated, we can put it in a public cache.

PRIVATE => response depends on authentication, if user is authenticated, does not put it in a public cache.

sbrunner: The PUBLIC/PRIVATE is used to manage the public caches (in Proxy)...

arnaud-morvan: in case we are in intranet (for example for roles) we can consider as authenticated

geoportal/c2cgeoportal_geoportal/lib/caching.py
```diff
@@ -258,16 +262,19 @@ def _set_common_headers(
...
-        if cache in (Cache.PUBLIC, Cache.PRIVATE)
+        if cache in (Cache.PUBLIC, Cache.PUBLIC_CAN_PRIVATE, Cache.PRIVATE)
         else service_headers_settings.get("cache_control_max_age_nocache", 10)
     )
     response.cache_control.max_age = maxage
     if maxage == 0:
         response.cache_control.no_cache = True
         response.cache_control.no_store = True
-    elif cache in (Cache.PUBLIC, Cache.PUBLIC_NO):
+    elif cache in (Cache.PUBLIC, Cache.PUBLIC_CAN_PRIVATE, Cache.PUBLIC_NO, Cache.PUBLIC_CAN_PRIVATE_NO):
```
arnaud-morvan, line 277: In case PUBLIC_CAN_PRIVATE, if user is authenticate we should return private, so it is the same as PRIVATE.

geoportal/tests/functional/test_login.py
```diff
@@ -350,6 +350,7 @@ class F:
             "functionalities": {"func": ["value"]},
         }
         self.assertEqual(login.loginuser(), expected)
+        request.response.headers["Vary"] = "Origin, Cookie"
```
arnaud-morvan, line 353: This is not an assertion, you set a value in the headers

sbrunner: :-P

PR #9953, opened 2022-08-10
Should work on Python 3.5 

sbrunner:


.github/workflows/main.yaml
```diff
@@ -154,6 +154,12 @@ jobs:
           if-no-files-found: ignore
           retention-days: 5
 
+      # Use minimal version from the documentation
+      - uses: actions/setup-python@v4
+        with:
+          # When we upgrade this we should also upgrade the requirements
+          python-version: '3.5'
```
arnaud-morvan, line 161: I suppose this is the minimal version to have on host for running build and upgrade scripts.
Maybe add a link to the related requirements in docs.

arnaud-morvan (approved):

PR #9955, opened 2022-08-10
Use pyramid SecurityPolicy

arnaud-morvan:


geoportal/c2cgeoportal_geoportal/lib/security.py
```diff
@@ -0,0 +1,50 @@
+from pyramid.authentication import AuthTktCookieHelper
+from pyramid.authorization import ACLHelper, Authenticated, Everyone
+from pyramid.interfaces import ISecurityPolicy
+
+from c2cgeoportal_geoportal.resources import defaultgroupsfinder
+from jinja2.runtime import identity
+from macaroonbakery.bakery._identity import Identity
+
+
+class SecurityPolicy(ISecurityPolicy):
```
sbrunner, line 10: Is it normal that self is missing in the ISecurityPolicy?
https://github.com/Pylons/pyramid/blob/1e439035e41da3767acbcae79689fb05dfdeb317/src/pyramid/interfaces.py#L502

arnaud-morvan: Same for other interfaces, but pretty sure we do not really need inheritance, just a class and methods with such prototype.
Inheritance can help for typing.
I do not know details about zope.Interface.
This is a draft for discussion, we will discuss, write tests and code.
But good catch anyway.

geoportal/c2cgeoportal_geoportal/lib/security.py
```diff
@@ -0,0 +1,50 @@
+from pyramid.authentication import AuthTktCookieHelper
+from pyramid.authorization import ACLHelper, Authenticated, Everyone
+from pyramid.interfaces import ISecurityPolicy
+
+from c2cgeoportal_geoportal.resources import defaultgroupsfinder
+from jinja2.runtime import identity
+from macaroonbakery.bakery._identity import Identity
```
arnaud-morvan, line 7: IDE imports, remove

PR #10030, opened 2022-09-26
Add information about URL of external layers in changelog.

arnaud-morvan:
https://jira.camptocamp.com/browse/GSGMF-1897

Follow https://github.com/camptocamp/ngeo/pull/8845


arnaud-morvan:
Seems that https://epsg.io/trans?s_srs=4326&t_srs=21781&data=5.96,47.81 do not return an array anymore, but just an object.
So error in CI seems not related with this PR.

sbrunner:
Shouldn't be rebased?

PR #10034, opened 2022-09-27
Remove the upgrade test from 2.6 to fix the CI

sbrunner:


arnaud-morvan (approved):
We could consider reintegrate 2.6 test if this pass : https://github.com/camptocamp/c2cgeoportal/pull/10038

PR #10175, opened 2022-11-16
Disable print snapping near 0° rotation

llienher:
Fix [GSGMF-1917](https://jira.camptocamp.com/browse/GSGMF-1917)


sbrunner:
For me, we should add `interfaces_config.default.constants.gmfOptions.view`in the `update_paths` of the `vars.yaml`...

sbrunner:
The build will be fixed with:  #10180

sbrunner (approved):

PR #10182, opened 2022-11-17
Remove JSON sort_keys

sbrunner:
From: https://github.com/camptocamp/c2cwsgiutils/blob/release_4/c2cwsgiutils/pretty_json.py

Todo: test it with ticino GMF

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -239,29 +239,28 @@ jobs:
         if: always()
       - run: df -h
         if: always()
-
-  audit:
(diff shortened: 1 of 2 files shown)


arnaud-morvan (approved):
I made a test and it works for layer vgeo_riserve_forestali.

Note that it also works by overwriting JSON renderer, after include of c2cwsgiutils:
```
from pyramid.renderers import JSON
config.add_renderer("json", JSON())
```
It was not working on my workstation during meet because I already changed theme renderer to **fast_json** and I was overwriting **json** one.

PR #10242, opened 2022-12-15
Add variable content_security_policy_iframe_api_frame_ancestors

arnaud-morvan:


geoportal/c2cgeoportal_geoportal/scaffolds/update/geoportal/CONST_vars.yaml_tmpl
```diff
@@ -871,6 +873,14 @@ vars:
         Content-Security-Policy: 'default-src {content_security_policy_c2c_default_src};
           script-src {content_security_policy_c2c_script_src};
           style-src {content_security_policy_c2c_style_src};'
+    - pattern: '^/iframe_api?.*'
```
sbrunner, line 876:
```suggestion
    - pattern: '^/iframe_api([/\?].*)?$'
```

arnaud-morvan: Fixed.
In fact it works perfectly with '^/iframe_api$' as it is matched with request.path_info which only contains path and not query string.

sbrunner: But the theme is missing, or also not present?

arnaud-morvan: my bad, forgot about theme in path, maybe it would me more restrictive and readable with : `"^/iframe_api(/theme/.*)?$"` which match `"/iframe"` and `"/iframe_api/theme/Demo"` for example.

sbrunner: Looks good :-)

sbrunner (approved):

PR #10275, opened 2023-01-16
Use main_ogc_server from vars.yaml for WMS proxy

llienher:
Implements [GSGMF-1931](https://jira.camptocamp.com/browse/GSGMF-1931)

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/update/{{cookiecutter.project}}/CONST_CHANGELOG.txt
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/update/{{cookiecutter.project}}/CONST_CHANGELOG.txt
@@ -17,6 +17,9 @@ Information
    Migration guide: https://mapserver.org/MIGRATION_GUIDE.html#mapserver-7-6-to-8-0-migration.
    Deprecated element from `CLASS` removed (moved to `STYLE`) https://mapserver.org/development/rfc/ms-rfc-133.html#rfc133.
 
+2. We added the support of 'main_ogc_server' to WMS mapserver proxy URL.
+    This means that if the value 'vars/main_ogc_server' is present in the vars.yaml of the project,
+    that will be used as default value when no 'ogcserver' parameter is present in the request.
 
 Changes to apply
 ================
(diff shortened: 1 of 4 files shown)


arnaud-morvan (commented):
Looks good for the runtime, but I suppose we also need to: 
- add it in CONST_config-schema.yaml
- propose a default value in vars.yaml for new geoportals
- add a line in changelog.
- maybe add something in documentation

llienher:
> * add a line in changelog.
> * maybe add something in documentation

Good point, I'm doing it.

geoportal/c2cgeoportal_geoportal/scaffolds/update/{{cookiecutter.project}}/CONST_CHANGELOG.txt
```diff
@@ -17,6 +17,10 @@ Information
    Migration guide: https://mapserver.org/MIGRATION_GUIDE.html#mapserver-7-6-to-8-0-migration.
    Deprecated element from `CLASS` removed (moved to `STYLE`) https://mapserver.org/development/rfc/ms-rfc-133.html#rfc133.
 
+2. We added the support of 'main_ogc_server' to WMS mapserver proxy URL.
+    This means that if the value 'vars/main_ogc_server' is present in the vars.yaml of the project,
+    that will be used when no 'ogcserver' parameter is present in the request.
```
jwkaltz, line 22: proposal:
"that will be used as default value when no..."

jwkaltz (commented):
Regarding the issue about documentation:
I think the documentation is already covered via the sentence "As the integrator, you need to edit the vars.yaml and env.project files to configure the application.",
BUT this assumes it is clear in the vars.yaml that the "main_ogc_server" should/could be configured.
Right now, the CONST_vars in the GMF scaffolds has this:

main_ogc_server: source for image/png

I think that
* the vars.yaml of the scaffolds should have a main_ogc_server definition: then it's clear to the integrator it should / could be configured,
* the default value we use should not contain spaces, because this leads to problems (but that's presumably a different issue).

llienher:
@arnaud-morvan I added the main_ogc_server to the vars.yaml and CONST_config-schema.yaml.
 
@jwkaltz So I guess we can assume that from the doc it is correct with the extra vars.yaml from the scaffold, thanks for your search.

arnaud-morvan (approved):
LGTM, thanks

jwkaltz (approved):
The changes as such LGTM
But still I wonder if this is the perfect opportunity to remove the spaces within our example ogc server name? 
The spaces caused errors when embedding the link in Jura oereb, and the renaming to the same thing without spaces was painful because several systems had to be coordinated for that. But I can do a separate PR for that if you prefer.

llienher:
@jwkaltz We should discuss this at the daily that everyone has the same info. Stéphane told me it was not straight-forward as it looks for the same reason that you mentioned, several systems have to be changed together ...

PR #10286, opened 2023-01-23
Add remark about from with version we can do the upgrade

sbrunner:


jwkaltz (commented):
We have something more specific below on this page,

"Upgrade your project first to 2.4, as this is the current long-term release. Migration scripts and instructions assume that your minimal version is the current long-term release."

PR #10290, opened 2023-01-24
Add username to print config and templates

vuilleumierc:
[GSGMF-1934](https://jira.camptocamp.com/browse/GSGMF-1934)


sbrunner:
Can you give us an example of result?

vuilleumierc:
> Can you give us an example of result?

Do you mean a pdf? Here is the result with the demo running locally:
[GeoMapFish_2023-01-24_13-53-36.pdf](https://github.com/camptocamp/c2cgeoportal/files/10491416/GeoMapFish_2023-01-24_13-53-36.pdf)

sbrunner (approved):
Nice :-)

PR #10337, opened 2023-02-07
Add OAuth2 -  PKCE authentication

sbrunner:


Makefile
```diff
@@ -104,6 +104,12 @@ build-test-mapserver:
 tests: ## Run all the unit tests
 tests: tests-commons  tests-geoportal tests-admin tests-qgisserver
 
+# -x, --exitfirst       Exit instantly on first error or failed test
+# --lf, --last-failed   Rerun only the tests that failed at the last run (or all if none failed)
+# --ff, --failed-first  Run all tests, but run the last failures first. This may re-order tests
+# 						and thus lead to repeated fixture setup/teardown.
+# --nf, --new-first     Run tests from new files first, then the rest of the tests sorted by file mtime
```
arnaud-morvan, line 111: Did you forgot to remove those comments ?

arnaud-morvan (approved):
LGTM, trust you as I'm not really comfortable with this part.

Makefile
sbrunner: I intentionally keep it to quickly know the useful options that can be used to customize the tests, what do you think?

sbrunner:
> trust you as I'm not really comfortable with this part.

Yes, I will be in the same position if we reverse the roles, thanks for the review :-)

Makefile
arnaud-morvan: Not sure it is the good place, maybe here: https://github.com/camptocamp/c2cgeoportal/blob/master/doc/developer/server_side.rst

PR #10350, opened 2023-02-21
Check if user has been deactivated

vuilleumierc:
GSGMF-1936


.gitignore
```diff
@@ -5,6 +5,7 @@
 *.egg-info/
 /*.timestamp
 /.mypy_cache/
+/.vscode/
```
sbrunner, line 8: It should be in your `~/.gitignore` file :-)

sbrunner (approved):
Thanks :-)

PR #10356, opened 2023-02-23
Add missing AZURE_STORAGE_ACCOUNT_URL for tilecloud-chain

sbrunner:
See: GSGMF-1967


llienher (approved):

jwkaltz (commented):
It's not needed in the geoportal, tools and webpack_dev_server environments?
(I am comparing with entries such as AZURE_STORAGE_SAS_TOKEN)

sbrunner:
No with GDAL we use AZURE_STORAGE_ACCOUNT witch already present: https://gdal.org/user/virtual_file_systems.html#vsiaz
It's used here
https://github.com/camptocamp/tilecloud-chain/blob/3b121a3d8ccb65d8e72e9d26bd874302c83e7a07/tilecloud_chain/controller.py#L105

jwkaltz (approved):

PR #10463, opened 2023-04-05
Fix Fiona installation

sbrunner:
Fiona uses dynamic fields in the pyproject.toml, that only works in resent version of pip


Dockerfile
```diff
@@ -10,11 +10,13 @@ SHELL ["/bin/bash", "-o", "pipefail", "-cux"]
 # Workaround for newer version of setuptools
 ENV SETUPTOOLS_USE_DISTUTILS=stdlib
 
+# pip install --upgrade pip should be remobed when we upgrade from Ubuntu 22.04 to 24.04
```
arnaud-morvan, line 13:
```suggestion
# pip install --upgrade pip should be removed when we upgrade from Ubuntu 22.04 to 24.04
```

arnaud-morvan (approved):
LGTM

PR #10468, opened 2023-04-11
 Add a changelog entry to configure the OGC server used in the WFS permalink

sbrunner:


arnaud-morvan (commented):
As far as I see this is 5 years old: https://github.com/camptocamp/c2cgeoportal/blame/74a9eb208acbd3cd19498cc0b8229a562d741f14/geoportal/c2cgeoportal_geoportal/scaffolds/update/%7B%7Bcookiecutter.project%7D%7D/geoportal/CONST_vars.yaml#L319

And only in CONST_vars.yaml regarding scaffolds.

So why adding this in changelog for 2.8 ?

sbrunner:
Good question... closing

PR #10485, opened 2023-04-24
Add the new fit option

sbrunner:


vuilleumierc (approved):
I would just add a small comment like "padding around WFS feature [top bottom left right]"

PR #10488, opened 2023-04-28
Fix OAuth2 with some clients

sbrunner:
The main thing is do not check the state if it's not required


arnaud-morvan (approved):
Trust you

PR #10552, opened 2023-06-21
Set top-left corner coordinate to get pixel perfect render

sbrunner:


ochriste:
these coordinate are for epsg2056, maybe add a comment somewhere explaining what the parameter is for and what are the coordinates exactly if someone want to use other projection ?

sbrunner:
@ochriste +1, fixed :-)

llienher (approved):

ochriste:
If I look a that:
  extent: &extent [2420000, 1030000, 2900000, 1350000]
  topLeft: &topLeft [2420000, 1350000]
  
  that topleft value looks more like a top right, unless the wmts grid is inverted compared to normal coordinates ?

ochriste:
ah yes, that complet non-sens ...
I will be annoying but can you add a comment complement? because in 6 months I will have forgotten this non-sens and will be be pulling my hair out trying to understand why using the normal, human, left coordinate is not working ....

sbrunner:
More seriously, it's, in fact, left, top, and the axis go in projection from bottom to top and in the tiles from top to bottom.

Comments clarifies according to it.

ochriste:
thank you

PR #10591, opened 2023-07-18
[Doc Admin] Add a note for layer with no template

ponceta:
Add a note for layer with no template : label layers with no wms / wfs query needed -> avoid performance loss


sbrunner (commented):
@ochriste for me it looks like ok

ochriste:
@ponceta je trouve la formulation un peu confuse, 
que pense tu de:
A layer with no interrogation template will not be queried by GeoMapFish. Unnecessarily querying layer can cause performance issue, especially on layers like labels. Only enable querying if needed.

ochriste (approved):

PR #10655, opened 2023-08-28
Add the list of the supported column types in the editing.

sbrunner:


doc/administrator/editing.rst
```diff
@@ -53,6 +53,22 @@ To be editable, a layer should satisfy the following requirements:
...
+      * ``integer``
+      * ``numeric``
+      * ``character varying``
+      * ``text``
+      * ``boolean``
+      * ``date``
+      * ``time without time zone``
+      * ``timestamp without time zone``
+
+   Note that th following types will have an issue with the timezone:
```
arnaud-morvan, line 67:
```suggestion
   Note that the following types will have an issue with the timezone:
```

arnaud-morvan (approved):

doc/administrator/editing.rst
```diff
@@ -53,6 +53,22 @@ To be editable, a layer should satisfy the following requirements:
...
+      * ``integer``
+      * ``numeric``
+      * ``character varying``
+      * ``text``
+      * ``boolean``
+      * ``date``
+      * ``time without time zone``
+      * ``timestamp without time zone``
+
+   Note that the following types will not working correctly with the timezone:
```
jwkaltz, line 67: The sentence is grammatically incorrect,
but more importantly: it is not clear why a type "with time zone" will not work correctly with the timezone. Is it a bug in GMF? Or what exactly is not working?

sbrunner: GeoMapFish first implementation uses the types `... without time zone` then the other one needs a specific development....

jwkaltz: So are we saying the following?
"Note that the following timezone related types are currently not supported:"

sbrunner: I don't understand your sentences...
We provide a list of types not of timezones...

arnaud-morvan: Ok so:

```suggestion"
   Note that the following timezone aware types are currently not supported:
```

PR #10681, opened 2023-09-22
Fix doccumentation issues

sbrunner:


doc/integrator/authentication.rst
```diff
@@ -158,23 +156,23 @@ Some service of GeoMapFish has some host restriction if you mix the domain.
 Application authentication
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To be considered as authenticated we should have the correct `Cookie` header,
-we also check the `Referrer` header to be sure that the user is coming from the same domain.
-If he is equals to the `Host` header, we consider that the user is coming from the same domain.
+To be considered as authenticated we should have the correct ``Cookie`` header,
+we also check the ``Referrer`` header to be sure that the user is coming from the same domain.
```
arnaud-morvan, line 160:
```suggestion
we also check the ``Referer`` header to be sure that the user is coming from the same domain.
```

arnaud-morvan (approved):

doc/import_ngeo_config.py
```diff
@@ -21,7 +21,9 @@ def _format_type(type_definition: dict[str, Any]) -> str:
         if "typeArguments" in type_definition:
             postfix = f'\\<{", ".join([_format_type(t) for t in type_definition["typeArguments"]])}>'
 
-        return f":ref:`integrator_guide_ngeo_properties_{type_definition['name']}`{postfix}"
+        if type_definition["name"].lower() not in ():
```
ochriste, line 24: is this really ever going to do something ?

doc/integrator/authentication.rst
ochriste: if this change is the result of automatic orthographic correction, cant we add a global rule to ignore that particular word ?

sbrunner: +1 :-)

doc/import_ngeo_config.py
sbrunner: Oups remaining :-), Thanks

PR #10872, opened 2024-01-10
SQL Alchemy update to version 2

sbrunner:


geoportal/tests/functional/test_themes_edit_columns.py
```diff
@@ -213,10 +223,15 @@ def _get_request(layerid, username=None, params=None):
...
 
-    def test_themes_edit_columns(self):
+    def test_themes_edit_columns(self) -> None:
+        from c2cgeoportal_commons.models import DBSession
         from c2cgeoportal_geoportal.views.theme import Theme
 
         layer_id = self._create_layer(geom_type=True)
+
+        for a in DBSession.query(self._tables[0]).all():
+            print(a)
```
sbrunner, line 233: Todo remove

sbrunner:
@arnaud-morvan 
If you have a better solution than the commit `Fix required` it's welcome.
And for the commit `Fix order by OGC server` if we do not find a better solution, I will make a change in c2cgeoform to do not have to override the `_grid_rows` method.

arnaud-morvan:
Concerning Fix order by OGC server

It seems that SQLAlchemy 1 was transparently including the order by column under the hood while creating the select statement, but ignoring it when populating the result set.

So the solution to select it in base_query then ignore it in grid_rows seems acceptable.

arnaud-morvan (approved):

PR #10913, opened 2024-02-20
Fix custom user

sbrunner:


nicolas-sitylb:
Shouldn't we add something like:

```
from c2cgeoportal_commons.models.static import User, Role
```
in userdetails.py ?

ochriste:
@nicolas-sitylb User is imported already, not Role though
ping @sbrunner ?

nicolas-sitylb:
@ochriste without an explicit `User` import in `userdetails.py` the following error will be raised by the current PR:

```
...
  File "/usr/local/lib/python3.10/dist-packages/venusian/__init__.py", line 220, in scan
    __import__(modname)
  File "/app/yverdon_c2cgeoportal_geoportal/admin/views/userdetail.py", line 18, in <module>
    base_schema.add(roles_schema_node(User.roles))
NameError: name 'User' is not defined
Worker with pid 45 was terminated due to signal 15
```

From what I can currently read [in the doc](https://camptocamp.github.io/c2cgeoportal/2.8/integrator/customize.html ), it is indeed defined in the model file, but not in the `userdetails.py`

sbrunner:
@ochriste the import is required :-)

nicolas-sitylb:
Hint:
According to the compose v2 migration : 

I suggest to replace : 

```
docker-compose exec geoportal python3
```

by:

```
docker compose exec geoportal python3
```

doc/integrator/extend_data_model.rst
```diff
@@ -316,20 +316,24 @@ Execute the following code:
...
-
-.. prompt:: bash
+If the generated SQL looks good:
 
-   docker-compose exec tools psql
+.. prompt:: python
 
-And enter the SQL commands
+   CreateTable(Title.__table__).compile(engine)
+   CreateTable(UserDetail.__table__).compile(engine)
```
nicolas-sitylb, line 339: Should these two last commands be run in the same console as before?    
And should they actually create the database tables?    

Because it says:

```python
>>> CreateTable(Title.__table__).compile(engine)
<sqlalchemy.dialects.postgresql.base.PGDDLCompiler object at 0x7fdb9bb33cd0>

>>> CreateTable(UserDetail.__table__).compile(engine)
<sqlalchemy.dialects.postgresql.base.PGDDLCompiler object at 0x7fdb9bb32bf0>
```

without any error, but no tables are actually created in the database. 

BTW, the generated SQL code looks fine, except for one tiny detail: no sequence is attached to the `ID` fields (this could cause troubles).

```sql
CREATE TABLE main_28_static.title (
    id INTEGER NOT NULL,
    name VARCHAR NOT NULL,
    PRIMARY KEY (id)
);

CREATE TABLE main_28_static.userdetail (
    id INTEGER NOT NULL,
    phone VARCHAR,
    title_id INTEGER NOT NULL,
    PRIMARY KEY (id),
    FOREIGN KEY(id) REFERENCES main_28_static."user" (id),
    FOREIGN KEY(title_id) REFERENCES main_28_static.title (id)
);
```



Instead of that, I would do:

```sql

CREATE TABLE main_28_static.title (
    id INTEGER PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    name VARCHAR NOT NULL

);

CREATE TABLE main_28_static.userdetail (
    id INTEGER PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    phone VARCHAR,
    title_id INTEGER NOT NULL,
    FOREIGN KEY(id) REFERENCES main_28_static."user" (id),
    FOREIGN KEY(title_id) REFERENCES main_28_static.title (id)
);
```

Would this SQL code still be OK for GMF to work with ?

sbrunner:
```sql
PRIMARY KEY (id),
```
Do the job...

In the same console effectively :-)

What about:
```
Title.__table__.create()
UserDetail.__table__.create()
```

nicolas-sitylb: These commands both work and actually create the tables:

```py
Title.__table__.create()
UserDetail.__table__.create()
```

It create the `title` table as follow:

```sql
CREATE TABLE IF NOT EXISTS main_28_static.title
(
    id integer NOT NULL DEFAULT nextval('main_28_static.title_id_seq'::regclass),
    name character varying COLLATE pg_catalog."default" NOT NULL,
    CONSTRAINT title_pkey PRIMARY KEY (id)
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS main_28_static.title
    OWNER to "www-data";
```

so it indeed crates an associated sequence.

And the `userdetail` table looks like:

```sql
CREATE TABLE IF NOT EXISTS main_28_static.userdetail
(
    id integer NOT NULL,
    phone character varying COLLATE pg_catalog."default",
    title_id integer NOT NULL,
    CONSTRAINT userdetail_pkey PRIMARY KEY (id),
    CONSTRAINT userdetail_id_fkey FOREIGN KEY (id)
        REFERENCES main_28_static."user" (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE NO ACTION,
    CONSTRAINT userdetail_title_id_fkey FOREIGN KEY (title_id)
        REFERENCES main_28_static.title (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE NO ACTION
)
TABLESPACE pg_default;

ALTER TABLE IF EXISTS main_28_static.userdetail
    OWNER to "www-data";
```

For this table, there is surprisingly no associated sequence 🤔.

sbrunner: Yes, it's normal because the sequence is the same for all the table's hierarchy.
From main_28_static."user" in this case :-)

vuilleumierc (approved):
LGTM without context, but there are 3 code blocks near the end that aren't formatted (see https://github.com/camptocamp/c2cgeoportal/blob/2c79c6aaa05530c8646c0d0732e8fc321faf8968/doc/integrator/extend_data_model.rst)

sbrunner:
@vuilleumierc there are not many more contexts, the current procedure wasn't updated throw the years...
For the unformatted blocks, it's because GitHub hasn't the prompt add-ons :-)

PR #10939, opened 2024-03-01
Remove no more used files

sbrunner:


sbrunner:
In fact, those files are used...

PR #11058, opened 2024-04-25
The function clone_schema is only available with Postgres 14+.

ochriste:


sbrunner:
Strange, what's the issue with version <14?
For me, it should work with versions >=11.

ochriste:
@sbrunner I just check on my local psotgres 12 and that method is not available by default
maybe you isntalled it manually ?

sbrunner:
> maybe you isntalled it manually
Yes
https://camptocamp.github.io/c2cgeoportal/master/integrator/database.html#start-a-new-version

PR #11231, opened 2024-07-10
Fix the admin descriptions about the COG

sbrunner:


commons/c2cgeoportal_commons/models/main.py
```diff
@@ -1223,7 +1223,8 @@ class LayerCOG(Layer):
             _(
                 """
             <div class="help-block">
-                <p>Definition of a <code>COG Layer</code>.</p>
+                <p>Definition of a <code>COG Layer</code> (COG for
+                <a href="https://www.cogeo.org/">Cloud Optimized GeoTIFF</a>).</p>
```
arnaud-morvan, line 1227: Missing:

```
<p>Note: The layers named <code>cog-defaults</code> contains the values
                used when we create a new <code>COG layer</code>.</p>
```

commons/c2cgeoportal_commons/models/main.py
```diff
@@ -1250,9 +1251,7 @@ class LayerCOG(Layer):
                     _(
                         """
             <div class="help-block">
-                <p>Definition of a <code>COG Layer</code>.</p>
-                <p>Note: The layers named <code>cog-defaults</code> contains the values
-                used when we create a new <code>COG layer</code>.</p>
+                <p>URL of the COG</p>
```
arnaud-morvan, line 1254: Here you should remove Literal and help-block:

```
"description": _("URL of the COG"),
```

not sure it works otherwise with the replacement at runtime.

arnaud-morvan (approved):

commons/c2cgeoportal_commons/models/main.py
sbrunner: It's not missing, it's just below!

PR #11232, opened 2024-07-10
Don't do 2 SQL updates per editing

sbrunner:
Be able to use existing table

Don't do 2 updates per editing:
In papyrus, there is a flush that creates a first INSERT or UPDATE
The idea is to update the fields in the before_* callback to avoid having two REQUESTS

See JIRA issue: [GSSIT-114](https://camptocamp.atlassian.net/browse/GSSIT-114).

[GSSIT-114]: https://camptocamp.atlassian.net/browse/GSSIT-114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/dbreflection.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/dbreflection.py
@@ -130,14 +130,19 @@ def get_table(
     # create table and reflect it
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", "Did not recognize type 'geometry' of column", SAWarning)
-        args = [tablename, metadata]
+        args = []
         if primary_key is not None:
             # Ensure we have a primary key to be able to edit views
             args.append(Column(primary_key, Integer, primary_key=True))
         with _get_table_lock:
-            table = Table(*args, schema=schema, autoload_with=engine)  # type: ignore[arg-type]
-        print(f"Table {tablename} loaded")
-        print([c.name for c in table.columns])
+            table = Table(
+                tablename,
+                metadata,
+                *args,
+                schema=schema,
+                autoload_with=engine,
+                keep_existing=True,
+            )
     return table
 
 
--- a/geoportal/c2cgeoportal_geoportal/views/layers.py
+++ b/geoportal/c2cgeoportal_geoportal/views/layers.py
@@ -39,6 +39,7 @@ import pyramid.response
 import shapely.geometry
 import sqlalchemy.ext.declarative
 import sqlalchemy.orm
+import sqlalchemy.orm.query
 from geoalchemy2 import Geometry
 from geoalchemy2.shape import from_shape, to_shape
 from geojson.feature import Feature, FeatureCollection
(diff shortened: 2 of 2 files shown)


arnaud-morvan (commented):
This is a complex pull request, I would like to discuss before finishing the review.
I suspect this could have some not expected or unanticipated side effects.

geoportal/c2cgeoportal_geoportal/lib/dbreflection.py
```diff
@@ -128,7 +128,11 @@ def get_table(
             # Ensure we have a primary key to be able to edit views
             args.append(Column(primary_key, Integer, primary_key=True))
         with _get_table_lock:
-            table = Table(*args, schema=schema, autoload=True, autoload_with=engine)
+            try:
+                table = Table(*args, schema=schema, autoload=True, autoload_with=engine)
+            except sqlalchemy.exc.InvalidRequestError:
+                # Try to use extend_existing
+                table = Table(*args, schema=schema, autoload=True, autoload_with=engine, extend_existing=True)
```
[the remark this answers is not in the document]
sbrunner: I change to use keep_existing, then I have 2 questions:
- is the try useful, Or should I directly add the keep_existing to the first call? (finally, I think that not useful)
- are you OK to merging it to 2.8, or should I move this pull request to master?

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -66,10 +67,123 @@
...
+            )
+            if allowed.scalar() == 0:
+                raise HTTPForbidden()
+
+            # Check if geometry is valid
+            if Layers._get_validation_setting(self.layer, request):
+                Layers._validate_geometry(spatial_elt)
+
+
+class _UpdateCallback(_LogLastUpdate):
```
arnaud-morvan, line 135: I would have used composition instead of inheritance here.

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -66,10 +67,123 @@
...
+            allowed = _get_base_query(request, self.layer)
+            allowed = allowed.filter(
+                or_(RestrictionArea.area.is_(None), RestrictionArea.area.ST_Contains(spatial_elt))
+            )
+            if allowed.scalar() == 0:
+                raise HTTPForbidden()
+
+            # Check if geometry is valid
+            if Layers._get_validation_setting(self.layer, request):
+                Layers._validate_geometry(spatial_elt)
```
arnaud-morvan, line 132: No call to `_LogLastUpdate.update` here ?

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -256,52 +382,22 @@ def count(self) -> int:
...
+        protocol = self._get_protocol_for_request()
+        log_updater = _LogLastUpdate(self._get_layer_for_request())
         try:
             features = protocol.create(self.request)
             if isinstance(features, HTTPException):
                 raise features
             if features is not None:
                 for feature in features.features:  # pylint: disable=no-member
-                    self._log_last_update(layer, feature)
+                    log_updater.update(self.request, feature)
```
arnaud-morvan, line 400: Do not we still have an UPDATE request after then INSERT request here ?

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -66,10 +67,123 @@
...
+        last_update_date = Layers.get_metadata(self.layer, "lastUpdateDateColumn")
+        if last_update_date is not None:
+            setattr(obj, last_update_date, datetime.now())
+
+        last_update_user = Layers.get_metadata(self.layer, "lastUpdateUserColumn")
+        if last_update_user is not None:
+            setattr(obj, last_update_user, request.user.id)
+
+
+def _get_base_query(request: pyramid.request.Request, layer: "main.Layer") -> sqlalchemy.orm.query.Query:
```
arnaud-morvan, line 90:
```suggestion
def _get_geometry_check_base_query(request: pyramid.request.Request, layer: "main.Layer") -> sqlalchemy.orm.query.Query:
```

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -66,10 +67,123 @@
 
 if TYPE_CHECKING:
     from c2cgeoportal_commons.models import main  # pylint: disable=ungrouped-imports.useless-suppression
+
+
 LOG = logging.getLogger(__name__)
 CACHE_REGION = get_region("std")
 
 
+class _LogLastUpdate:
```
arnaud-morvan, line 76: IMHO there is no need for a class here, a function should be enough.

Or you should call it `_BaseCallback` and put `_get_base_query` as a method of this class.

arnaud-morvan (commented):
I understand things.

Note that actually we get two requests (INSERT + UPDATE or UPDATE + UPDATE) but only one commit. 

You also fix another issue, conflict when table already exist in DeclarativeMeta using option `keep_existing`. I would have added it in description and maybe in changelog.

Main subject could also be added to changelog IMHO.

geoportal/c2cgeoportal_geoportal/views/layers.py
sbrunner: After renaming _LogLastUpdate to _BaseCallback I think that's clearer to use inheritance (event if composition is possible).

geoportal/c2cgeoportal_geoportal/views/layers.py
sbrunner: I look like to be true. I will do deeper checks, thanks :-)

geoportal/c2cgeoportal_geoportal/views/layers.py
sbrunner: Finally, no because the obj is None

geoportal/c2cgeoportal_geoportal/views/layers.py
sbrunner: Yes, but to avoid that, we should add an after delete in papyrus!

sbrunner:
@arnaud-morvan ready for a new pass :-)

geoportal/c2cgeoportal_geoportal/views/layers.py
```diff
@@ -68,10 +69,120 @@
...
 if TYPE_CHECKING:
     from c2cgeoportal_commons.models import main  # pylint: disable=ungrouped-imports.useless-suppression
+
+
 _LOG = logging.getLogger(__name__)
 _CACHE_REGION = get_region("std")
 
 
+class _BaseCallback:
+    def __init__(self, layer: "main.Layer"):
```
arnaud-morvan, line 79: Why not:

```suggestion
    def __init__(self, layer: main.Layer):
```

arnaud-morvan (approved):

geoportal/c2cgeoportal_geoportal/views/layers.py
sbrunner: In fact, it's not possible because it's requires to import `c2cgeoportal_commons.models.main` and he is not initialized :-)

PR #11337, opened 2024-08-28
Add changelog about the ngeo renovation

sbrunner:


ger-benjamin (approved):
I think you covered every topics.

But the breaking changes (if you have custom interface, custom plugins) can be hard to address. Should we had more details somewhere ?

sbrunner:
I don't know, for me, it can be difficult to create, it's why I put the main pull request to be a good way to starts.

PR #11386, opened 2024-09-19
OpenID Connect: Add hook to be able to customize role creation

sbrunner:
TODO:
 - [x] Update documentation about configuration used only on hooks
 - [x] Add documentation about hooks

### DIFF
--- a/doc/integrator/authentication_oidc.rst
+++ b/doc/integrator/authentication_oidc.rst
@@ -89,3 +89,34 @@ Other options
 ``scopes``: The list of scopes to request, default is [``openid``, ``profile``, ``email``].
 
 ``query_user_info``: If ``true``, the user info will be requested instead if using the ``id_token``, default is false.
+
+~~~~~
+Hooks
+~~~~~
+
+If you want to redefine the user creation process, you can use the hooks ``get_remember_from_user_info``
+and ``get_user_from_remember``.
+
+``get_remember_from_user_info``: This hook is called during the user is authentication.
+The argument are the pyramid ``request``, the received ``user_info``, and the ``remember_object`` dictionary
+to be filled and will be stored in the cookie.
+
+``get_user_from_remember``: This hook is called during the user is certification.
+The argument are the pyramid ``request``, the received ``remember_object``, and the ``create_user`` boolean.
+The return value is the user object ``User`` or ``DynamicUsed``.
+
+Full signatures:
+
+.. code:: python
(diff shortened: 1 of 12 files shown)


doc/integrator/authentication_oidc.rst
```diff
@@ -84,8 +80,107 @@ Other options
...
+~~~~~~~~~~~~~~~~~~~~
+Example with Zitadel
+~~~~~~~~~~~~~~~~~~~~
+
+.. code:: yaml
+
+   vars:
+     authentication:
+       openid_connect:
+         enabled: true
```
danduk82, line 117: remove this because the environment variable should be used instead

geoportal/c2cgeoportal_geoportal/lib/oidc.py
```diff
@@ -96,6 +97,115 @@ class OidcRememberObject(TypedDict):
...
+    request: pyramid.request.Request, remember_object: OidcRememberObject, create_user: bool = False
+) -> Union["static.User", DynamicUser] | None:
+    """
+    Create a user from the remember object filled from `get_remember_from_user_info`.
+
+    :param remember_object: The object to fill, by default with the `username`, `email`, `settings_role` and `roles`.
+    :param settings: The OpenID Connect configuration.
+    :param create_user: If the user should be created if it does not exist.
+    """
+    from c2cgeoportal_commons import models  # pylint: disable=import-outside-toplevel
```
danduk82, line 148: can you add a comment here to explain why the models are imported here and not at the top level?

geoportal/c2cgeoportal_geoportal/lib/oidc.py
```diff
@@ -96,6 +97,115 @@ class OidcRememberObject(TypedDict):
...
+                    user.display_name = display_name
+                elif field == "email":
+                    user.email = email
+                else:
+                    raise HTTPInternalServerError(
+                        f"Unknown update_field: '{field}', allowed values are 'username', 'display_name' and 'email'"
+                    )
+        if user is None and create_user is True:
+            user = static.User(username=username, email=email, display_name=display_name)
+            models.DBSession.add(user)
```
danduk82, line 190: Shouldn't this section be in the login function? This will create a transaction for each request otherwise.

danduk82 (changes requested):

danduk82 (approved):

PR #11400, opened 2024-10-01
Be able to disable some evaluate

sbrunner:


docker-compose.yaml
```diff
@@ -76,6 +76,11 @@ services:
       - PGSSLMODE=disable
       - CI=true
       - TEST=true
+      - EVALUATE_ADD=false
```
ger-benjamin, line 79:
```suggestion
      - EVALUATE_APP=false
```
Maybe ?

ger-benjamin (approved):

docker-compose.yaml
sbrunner: +1 Thanks :-)

PR #11451, opened 2024-10-09
Fix performance issue

sbrunner:
Fix: https://camptocamp.sentry.io/issues/5600078816/
See also: https://docs.sentry.io/product/issues/issue-details/performance-issues/n-one-queries/


geoportal/tests/functional/__init__.py
```diff
@@ -32,22 +32,25 @@
...
 from pyramid import testing
+from sqlalchemy.orm.session import Session
 
 import c2cgeoportal_geoportal
 import c2cgeoportal_geoportal.lib
 from c2cgeoportal_commons import models
 from c2cgeoportal_geoportal.lib import caching
 
+if TYPE_CHECKING:
+    from c2cgeoportal_commons.models import main, static
```
ger-benjamin, line 52: Waouw... is there no better way to achieve this ?

ger-benjamin (commented):
Mhh I need @arnaud-morvan here please.

geoportal/tests/functional/__init__.py
```diff
@@ -76,27 +79,27 @@ def cleanup_db() -> None:
...
-    for r in models.DBSession.query(Role).all():
-        r.functionalities = []
-        models.DBSession.delete(r)
-    models.DBSession.query(User).delete()
-    models.DBSession.query(Functionality).delete()
-    models.DBSession.query(FullTextSearch).delete()
-    models.DBSession.query(Shorturl).delete()
-    models.DBSession.query(OAuth2Client).delete()
-    transaction.commit()
+    with models.DBSession() as session:
```
arnaud-morvan, line 82: IMHO we should completely avoid using the global DBSession object and use a fixture instead, but I may miss something.

geoportal/tests/functional/__init__.py
```diff
@@ -170,21 +177,21 @@ def teardown_common() -> None:
...
 
 
-def create_default_ogcserver():
-    from c2cgeoportal_commons.models import DBSession
+def create_default_ogcserver(session: Session) -> "main.OGCServer":
     from c2cgeoportal_commons.models.main import OGCServer
 
-    transaction.commit()
-    ogcserver = OGCServer(name="__test_ogc_server")
+    session.flush()
```
arnaud-morvan, line 183: We should flush at end of changes instead of before.
It would be really easier to debug.
Error would raise in function which made a mistake.

Also note that here pretty sure next SELECT will raise a flush.

geoportal/tests/functional/conftest.py
```diff
@@ -46,23 +56,51 @@ def settings():
 
 
 @pytest.fixture
-def dbsession(settings):
+def dbsession_old(settings: dict[str, Any]) -> Session:
```
arnaud-morvan, line 59: If we keep it, we could find a better name.

arnaud-morvan (commented):
Strategy here seems not clear to me.

We use multiple dbsession fixtures, sometimes with transaction, sometimes not, sometimes with commit inside some function.

IMHO we should have two dbsession fixture, one which is supposed to be committed, one which should not. And if possible none based on global models.DBSession.

arnaud-morvan:
The perfomance issue fix LGTM

geoportal/tests/functional/__init__.py
sbrunner: Applied in #11460

sbrunner (commented):
> Strategy here seems not clear to me.

> We use multiple dbsession fixtures, sometimes with transaction, sometimes not, sometimes with commit inside some function.

The idea is to have the `test_theme` test clean as it should be and to the minimal changes for the other tests to make them work.

> IMHO we should have two dbsession fixture, one which is supposed to be committed, one which should not. And if possible none based on global models.DBSession.

Effectively, we have many tests that not relay clear and require bad session...

The goal is to do one step in the right direction.

arnaud-morvan:
How about merging only the performance fix here, and discuss about tests refactoring in #11460

sbrunner:
@arnaud-morvan can you have a look again?

arnaud-morvan (approved):
+1
Number of database queries is a first-class parameter in achieving good backend performance.

sbrunner:
Thanks :-)

PR #11460, opened 2024-10-15
Convert test_theme as new style test, cleanup

sbrunner:


geoportal/tests/functional/conftest.py
```diff
@@ -25,19 +25,29 @@
...
+from typing import TYPE_CHECKING, Any
 from unittest.mock import patch
 
+import pyramid.request
 import pytest
 import sqlalchemy.orm
 import transaction
 from c2c.template.config import config as configuration
 from pyramid.testing import DummyRequest
+from sqlalchemy.orm.session import Session, SessionTransaction
```
ger-benjamin, line 38: For my understanding, why this is not under the `if TYPE_CHECKING:` ? 
It's always used as a type.
But maybe that doesn't really matter ?

ger-benjamin (approved):
Looks good to me.
The `test_themes.py` results changed a bit, I hope it's wanted, maybe check again quickly before to merge

PR #11506, opened 2024-11-07
Also invalidate the ogc-server cache

sbrunner:


ochriste (approved):
I dont understand why the invalidation worked when you did it during our meet, 
if that was the cause, it shouldnt have worked, so im very confused

PR #11544, opened 2024-11-25
map.view.constrainResolution to false by default

ger-benjamin:
In vars, <interface>.gmfOptions.map.view.contrainResolution should be falsy by default. Because if "true", Openlayers makes the map looking "bumpy" on zoom in/out.

This is looks like an issue, especially on the default desktop.
Eventually, we could add `constrainResolution: True` on the desktop_alt.

See also issue opened in: https://github.com/camptocamp/GeoMapFish/issues/92

Change come from https://github.com/camptocamp/c2cgeoportal/commit/0819c1828fec85d3ba1d154222215acce10c6659#diff-25ddc46a3475e6781818af028a3ee89287e30737c656dcbcd85c0be7a36e3fd6L46

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
@@ -81,7 +81,7 @@ vars:
             maxTilesLoading: 128
           view:
             <<: *view
-            constrainResolution: True
+            constrainResolution: False
         gmfExternalOGCServers:
           - name: swisstopo WMS
             type: WMS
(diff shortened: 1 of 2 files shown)


geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
```diff
@@ -81,7 +81,7 @@ vars:
             maxTilesLoading: 128
           view:
             <<: *view
-            constrainResolution: True
+            constrainResolution: False
```
jwkaltz, line 84: Can we add a small comment in this file to explain why "False" is better?
So we don't make the mistake again of changing this setting?

ochriste:
as a reminder:

constrainResolution | boolean (defaults to false) | If true, the view will always animate to the closest zoom level after an interaction; false means intermediary zoom levels are allowed.

so it may have some side effects



</body>
</html>

sbrunner:
I suspect that with change the map becomes a little blurry, it's not what we want...

geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
ger-benjamin: It's not better, just... different. It's a pure Openlayers feature.
Well, I documented the commit, and the PR, + two issues and one jira ticket, that's already something.
But we could add this documentation in ngeo directly too https://camptocamp.github.io/ngeo/master/apidoc/types/srcapi_store_config.View.html#__type.constrainResolution  or eventually here:
https://camptocamp.github.io/c2cgeoportal/master/integrator/ngeo.html#view

ger-benjamin:
I'll dig deeper, that's probably not the only and best solution.

PR #11572, opened 2024-12-09
Get profile and height from service

RenataMuellerC2C:
Allow to use a profile service to get an elevation profile or a simple elevation.

### DIFF
--- a/doc/integrator/raster.rst
+++ b/doc/integrator/raster.rst
@@ -21,8 +21,12 @@ To configure these web services, you need to set the ``raster`` variable in the 
             file: /var/sig/altimetrie/mnt.vrt
             type: gdal
             round: 1
+        DTM2:
+            url: 'https://api3.geo.admin.ch/rest/services'
+            type: external_url
+            elevation_name: 'height'
 
-``raster`` is a list of "DEM layers". There are only two entries in this example, but there could be more.
+``raster`` is a list of "DEM layers".
 
 ``file`` provides the path to the shape index that references the raster files.
 The raster files should be in the Binary Terrain (BT/VTP .bt 1.3) format.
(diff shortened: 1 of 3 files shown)


doc/integrator/raster.rst
```diff
@@ -21,8 +21,11 @@ To configure these web services, you need to set the ``raster`` variable in the
             file: /var/sig/altimetrie/mnt.vrt
             type: gdal
             round: 1
+        DTM2:
+            url: 'https://api3.geo.admin.ch/rest/services'
+            elevation_name: 'height'
 
-``raster`` is a list of "DEM layers". There are only two entries in this example, but there could be more.
+``raster`` is a list of "DEM layers". There are only tree entries in this example, but there could be more.
```
sbrunner, line 28: Is this sentence useful?
```suggestion
``raster`` is a list of "DEM layers".
```

sbrunner (changes requested):
We already have a type in the raster layers, I think that we should use something like `type: external_uri`  :-)

geoportal/c2cgeoportal_geoportal/views/profile.py
```diff
@@ -25,20 +25,25 @@
...
-from pyramid.httpexceptions import HTTPNotFound
+import requests
+from pyramid.httpexceptions import HTTPBadRequest, HTTPNotFound
 from pyramid.i18n import TranslationStringFactory
 from pyramid.view import view_config
 
 from c2cgeoportal_geoportal.lib.common_headers import Cache, set_common_headers
 from c2cgeoportal_geoportal.views.raster import Raster
 
+LOG = logging.getLogger(__name__)
```
sbrunner, line 45:
```suggestion
_LOG = logging.getLogger(__name__)
```

sbrunner (approved):
Just one last comment :-)

PR #11573, opened 2024-12-09
Remove cache-control switch from private to public

sbrunner:
This is not working with intranet detection

To make it working (with intranet detection and reverse proxy cache) without the possibility of cache pollution, we should:
- Disable it for the loginuser end-point.
- Be sure that we have the user or roles in the query string, And verify them in the Pyramid view.


jwkaltz (approved):

ger-benjamin (approved):
Could this be a breaking change ? Should we add documentation about that ? Or it currently doesn't work ?

sbrunner:
@ger-benjamin I will add a message in the changelog to be transparent, this can't be associated to a breaking change :-)

PR #11589, opened 2024-12-13
Fix call to '/static'

sbrunner:
Fix https://camptocamp.sentry.io/issues/5052258158/?project=1851011&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=4


arnaud-morvan (approved):
LGTM

PR #11598, opened 2024-12-18
Don't have an authenticatication loop

sbrunner:
When we want to access to the admin interface with oidc and with a user that's not allowed to access to the admin interface.


jwkaltz (approved):
LGTM

PR #11629, opened 2025-01-08
Improve documentation for QGIS backend configuration

jwkaltz:
Improve documentation as suggested by @michmuel


michmuel (approved):
Thanks @jwkaltz 
I guess the failing tests are not releted to this change.

PR #11745, opened 2025-02-25
Use with statement for the SMTP connection

sbrunner:


llienher (approved):

llienher:
Should we backport to 2.8 as well ? That where geocommune is at currently.

PR #11877, opened 2025-04-09
Fix QGIS publish

sbrunner:


ponceta:
@sbrunner it looks like I have the exact same bug on my build. How can I retrieve this fix?

sbrunner:
You shouldn't because here we use OIDC authentication, and in your project you should use authentication using a username and password!

PR #11938, opened 2025-05-07
Simple application mode require the usage of update-po-from-url to up…

ochriste:
…date the translation file.


jwkaltz:
Is there a technical constraint for this difference @sbrunner ?
I mean, is it necessary to have two different command parameters, or can this be simplified?

PR #11986, opened 2025-05-15
Accept and POST, PUT, DELETE methods for OGC API Features part 4 (editing)

mki-c2c:
To be able to fully use OGC API requests via the geoportal proxy, also PUT, POST and DELETE methods must be accepted and forwarded to a QGIS Server instance.

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/__init__.py
@@ -674,7 +674,7 @@ def includeme(config: pyramid.config.Configurator) -> None:
         "/mapserv_proxy/{ogcserver}/wfs3/*path",
         mapserverproxy=True,
         pregenerator=C2CPregenerator(role=True),
-        request_method="GET",
+        request_method=("GET", "POST", "PUT", "DELETE"),
     )
     add_cors_route(config, "/mapserv_proxy", "mapserver")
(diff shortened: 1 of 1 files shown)


mki-c2c:
According to draft OGC API spec, the server should also accept an OPTIONS request and return the available PUT, POST, DELETE methods.
cf. https://docs.ogc.org/DRAFTS/20-002r1.html#create-replace-delete-overview

There is also another non-conformity: In case a method is not supported, the client is redirected towards the login page, best practice would probably be just a non-authorized response.

mki-c2c:
PATCH method does not seem to work in QGIS server yet :-(

sbrunner (approved):

geoportal/c2cgeoportal_geoportal/__init__.py
```diff
@@ -674,7 +674,7 @@ def handle(event: InvalidateCacheEvent) -> None:
         "/mapserv_proxy/{ogcserver}/wfs3/*path",
         mapserverproxy=True,
         pregenerator=C2CPregenerator(role=True),
-        request_method="GET",
+        request_method=("GET", "POST", "PUT", "DELETE"),
```
sbrunner, line 677: Just add PATCH and you can merge :-)

PR #12014, opened 2025-05-23
OIDC: Be able to ignore role, to have template in match

sbrunner:
- Be able to ignore role that's not existing in the GeoMapFish database
- Be able to have a template in match to be able to combine field
- Don't hardcode the 'role_admin' to be able to configure any roles in the default context (Root)

See JIRA issue: [GSEPF-995](https://camptocamp.atlassian.net/browse/GSEPF-995).

[GSEPF-995]: https://camptocamp.atlassian.net/browse/GSEPF-995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/lib/oidc.py
+++ b/geoportal/c2cgeoportal_geoportal/lib/oidc.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, Camptocamp SA
+# Copyright (c) 2024-2025, Camptocamp SA
 # All rights reserved.
 
 # Redistribution and use in source and binary forms, with or without
@@ -28,6 +28,7 @@
 import datetime
 import json
 import logging
+import string
 from typing import TYPE_CHECKING, Any, NamedTuple, Optional, TypedDict, Union
 
 import pyramid.request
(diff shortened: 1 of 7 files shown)


geoportal/c2cgeoportal_geoportal/lib/oidc.py
```diff
@@ -194,6 +210,14 @@ def get_user_from_remember(
                 user = static.User(username=username, email=email, display_name=display_name)
                 models.DBSession.add(user)
     else:
+        roles = []
+        role_names = remember_object.get("roles", [])
+        if role_names:
+            query = models.DBSession.query(main.Role).filter(main.Role.name.in_(role_names))
+            if openid_connect_configuration.get("ignore_unknown_roles", False):
+                roles = [role for role in query.all() if role is not None]
```
arnaud-morvan, line 218: `models.DBSession.query(main.Role).filter(main.Role.name.in_(role_names))` could return some `None`?

Not sure to understand option `ignore_unknown_roles` if you request roles from `main.role` with an `IN` filter.

arnaud-morvan (approved):
Looks good, but I dd not understand the option `ignore_unknown_roles` as it seems to have no effect, and I suppose in most cases OpenID might return roles which has nothing to do with GeoMapFish.

geoportal/c2cgeoportal_geoportal/lib/oidc.py
sbrunner: Good question, I have to remember that I get some None but your right it has no sense!
I will remove it and see what's happens :-)

PR #12102, opened 2025-06-30
Add route "shortener_fetch" to be able to retrieve an url without redirect

remyguillaume:
As discussed today during the demo, I've added a route to be able to fetch the corresponding url to a shortlink. 

This will allow the clients (like geogirafe or ngeo) to read the state directly with a `fetch`, without redirect process.
This will solve the problem accouring when the length of an URL exceeds the maximum allowed (depending on the infrastructure, the limit can be from 2000 to 8000 characters).

If everything is ok for you @sbrunner, can you please backport it to 2.9 please?
Thanks!

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/shortener.py
+++ b/geoportal/c2cgeoportal_geoportal/views/shortener.py
@@ -80,7 +80,7 @@ class Shortener:
         set_common_headers(self.request, "shortener", Cache.PUBLIC_NO)
         return HTTPFound(location=long_url)
 
-    @view_config(route_name="shortener_fetch", renderer="json")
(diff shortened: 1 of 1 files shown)


sbrunner (approved):

sbrunner:
There is a pre-commit issue, see: https://github.com/camptocamp/c2cgeoportal/pull/12102/checks?check_run_id=45063566349

remyguillaume:
Hi @sbrunner,

I don't really understand what is the error here.
Do you think you could find some minutes ot discuss this with me ?
Thanks!

remyguillaume:
Hi @sbrunner,

There is still an error in the CI, but I cannot get more infos, I probably do not have the right to see the output (I get a 401 Error).
Can you please shortly explain what is failing?

Thanks.

PR #12399, opened 2025-11-04
Fix target eslint

arnaud-morvan:
Not sure how we didn't catch this before, but it actually does not works for me without this fix.


sbrunner:
This layer is in geoportal (runner), but not in tools!
https://github.com/camptocamp/c2cgeoportal/blob/master/Dockerfile#L212-L214

PR #12406, opened 2025-11-04
Decrease WMS ratio to 1 by default

sbrunner:
As it was discussed with the PSC on Monday.

See JIRA issue: [RATIO-1](https://camptocamp.atlassian.net/browse/RATIO-1).


arnaud-morvan (approved):

sbrunner:
Thanks @arnaud-morvan  :-)

PR #12829, opened 2026-04-15
Fix server l10n compilation

arnaud-morvan:
Actually script `build-l10n` check existance of `{package}_geoportal-server.mo` which is the compiled version, it does not exist before compilation.

Fix it by checking existance of `{package}_geoportal-server.po` which is the source file.

### DIFF
--- a/bin/build-l10n
+++ b/bin/build-l10n
@@ -38,7 +38,7 @@ def main():
                     check=True,
                 )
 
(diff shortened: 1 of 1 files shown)


sbrunner (approved):
Looks good, shouldn't it be back-ported to newer branches?

arnaud-morvan:
I added labels for 2.8, 2.9 and master, did I forgot something ?

PR #12833, opened 2026-04-16
Fix eslint folders

sbrunner:


arnaud-morvan (approved):
LGTM, thanks, I remember I tried fixing this some times ago without success.

PR #12986, opened 2026-06-10
Add backend settings endpoint

sbrunner:
## What
- add fixed backend route `/settings` for authenticated users
- add GET `/settings` (returns `{}` for initial config)
- add POST `/settings` to overwrite saved settings JSON
- add configurable payload size limit via `settings.max_payload_size`
- add `user.settings` `JSONB` column in `static` schema
- add alembic migration for the new column
- add functional tests for auth, initial empty settings, overwrite, invalid JSON, and payload too large

Backend part of: https://gitlab.com/geogirafe/gg-viewer/-/work_items/825

See JIRA issue: [GSGIR-93](https://camptocamp.atlassian.net/browse/GSGIR-93).

[GSGIR-93]: https://camptocamp.atlassian.net/browse/GSGIR-93?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/tests/functional/test_settings.py
+++ b/geoportal/tests/functional/test_settings.py
@@ -1,0 +1,155 @@
+# Copyright (c) 2026, Camptocamp SA
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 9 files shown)


geoportal/tests/functional/test_settings.py
```diff
@@ -0,0 +1,155 @@
...
+import transaction
+from pyramid import testing
+
+from tests.functional import cleanup_db, create_dummy_request, setup_db
+from tests.functional import setup_common as setup_module  # noqa
+from tests.functional import teardown_common as teardown_module  # noqa
+
+
+class TestSettingsView(TestCase):
+    def setup_method(self, _) -> None:
```
arnaud-morvan: Note that other tests create a pytest fixture depending on dbsession and transact, for example https://github.com/camptocamp/c2cgeoportal/blob/master/geoportal/tests/functional/test_themes.py

arnaud-morvan (approved):
IHMO route `/setting` is ambiguous, I would propose `/user_settings` or `/user/settings` instead.

Otherwise looks good.

geoportal/tests/functional/test_settings.py
arnaud-morvan: And also for config key in `vars.yaml`:

  `settings` -> `user_settings`.
  `headers/settings` -> `headers/user_settings`
  

And view filename:

  `views/settings.py` -> `views/user_settings.py`

Route names also.

Maybe I forgot something.

PR #13023, opened 2026-06-22
Fix gmfOptions.map shared across interfaces via YAML alias

fredj:
## Problem

In `CONST_vars.yaml`, the `desktop`, `mobile`, and `iframe_api` interfaces all used a plain YAML alias (`*gmfOptions`) to reference the default interface's `gmfOptions`. This means all four interfaces share the **same Python dict object** for `gmfOptions` — and crucially, the same `map: {}` nested dict.

When `c2c-template`'s `update_vars()` processes interfaces in YAML order, it mutates these shared dicts in-place. If a later interface (e.g. `iframe_api` with `map: {}` and no `update_paths` entry for `map`) is processed after an earlier one that set a value (e.g. `mobile` with `moveTolerance: 8`), the later interface **replaces** the shared map dict, wiping the earlier value.

Concrete failure: setting `gmfOptions.map.moveTolerance` in the `mobile` interface had no effect because `iframe_api`'s `map: {}` (processed last) replaced the shared map with an empty dict.

## Fix

- **`CONST_vars.yaml`**: Replace `gmfOptions: *gmfOptions` with a merge key (`<<: *gmfOptions`) plus an explicit `map: {}` for each of `desktop`, `mobile`, and `iframe_api`. Each interface now has its own independent `map` dict that cannot be accidentally overwritten by another interface's processing.

- **`vars.yaml` (create scaffold)**: Add `interfaces_config.*.constants.gmfOptions.map` to `update_paths` for all interfaces so that project-level `map` settings are **merged** into (rather than replacing) the per-interface `map` dict.

## Notes

- The same bug exists in the `2.9` branch and will need to be backported.
- Related customer issue: GSTIC-57

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
+++ b/geoportal/c2cgeoportal_geoportal/scaffolds/create/{{cookiecutter.project}}/geoportal/vars.yaml
@@ -395,6 +395,7 @@ update_paths:
   - headers.login.headers
   - interfaces_config.default.constants.ngeoWfsPermalinkOptions
   - interfaces_config.default.constants.gmfOptions.view
+  - interfaces_config.default.constants.gmfOptions.map
   - interfaces_config.default.constants.gmfSearchOptions
   - interfaces_config.default.constants.gmfPermalinkOptions
   - interfaces_config.default.constants.gmfVectorTilesOptions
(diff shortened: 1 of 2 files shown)


sbrunner:
This looks to be just a workaround, the real issue is in c2c-`template` it should be fixed here, no?

fredj:
> This looks to be just a workaround, the real issue is in c2c-`template` it should be fixed here, no?

what about https://github.com/camptocamp/c2c.template/pull/482 ?

PR #13105, opened 2026-07-22
Respect query_user_info option in Bearer token authorization path

sbrunner:
## Summary

Respect the `query_user_info` configuration option in the Bearer token authorization path used for WMS/WFS/OGC API requests.

## Context

The `query_user_info` option (default: `False`) controls whether user information is fetched from the OIDC provider's UserInfo endpoint or extracted locally from the token. This option was already respected in the login callback path (`oidc.py:309`), but the Bearer token path in `__init__.py` always called `client.fetch_userinfo()`, making an HTTP request to the UserInfo endpoint on every WMS/WFS request.

This change makes the Bearer token path consistent with the login callback, avoiding unnecessary HTTP requests when the access token is a JWT containing the required user claims.

## Implementation Details

In `create_get_user_from_request()` (`__init__.py:408-420`):
- `query_user_info: True`: calls `client.fetch_userinfo(token)` (same as before)
- `query_user_info: False`: decodes the access token JWT locally via `JwtAccessToken.parse_jwt()` with signature validation, issuer/audience/expiry checks — no HTTP call

The `JwtAccessToken` model from `simple-openid-connect` v2.4.0 supports RFC 9068 JWT access tokens with `extra="allow"` to capture additional claims like `name`, `email`, `preferred_username`.

## Impact/Risks

- Backward compatible: existing behavior unchanged when `query_user_info: True`
- Requires the access token to be a JWT (RFC 9068) when `query_user_info: False`; most modern providers (Keycloak, Azure AD, Auth0) support this
- If the access token is opaque, `query_user_info` should remain `True`

See JIRA issue: [GSBASSUP-206](https://camptocamp.atlassian.net/browse/GSBASSUP-206).

[GSBASSUP-206]: https://camptocamp.atlassian.net/browse/GSBASSUP-206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/__init__.py
@@ -44,6 +44,7 @@ import dateutil.parser
 import pyramid.config
 import pyramid.request
 import pyramid.response
+import simple_openid_connect.data
 import sqlalchemy
 import sqlalchemy.orm
 import zope.event.classhandler
@@ -404,7 +405,19 @@ def create_get_user_from_request(
             ):
                 token = request.headers["Authorization"][7:]
                 client = oidc.get_oidc_client(request, request.host)
-                user_info = client.fetch_userinfo(token)
+
+                if openid_connect_configuration.get("query_user_info", False):
+                    user_info = client.fetch_userinfo(token)
+                else:
+                    access_token = simple_openid_connect.data.JwtAccessToken.parse_jwt(
+                        token, client.provider_keys
+                    )
+                    access_token.validate_extern(
+                        issuer=client.provider_config.issuer,
+                        client_id=client.client_auth.client_id,
+                    )
+                    user_info = access_token
+
                 user_info_remember = {}
                 request.get_remember_from_user_info(user_info.dict(), user_info_remember)
             elif username is None:
--- a/geoportal/c2cgeoportal_geoportal/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/__init__.py
@@ -372,6 +372,14 @@ def create_get_user_from_request(
         """
         Return the User object for the request.
 
+        Resolution order:
+        1. ``Authorization: Bearer`` header (OIDC JWT access token) — used by
+           QGIS Desktop and other clients. Either query the userinfo endpoint
+           or parse the JWT locally, depending on ``query_user_info``.
+        2. ``unauthenticated_userid`` (auth cookie or header) — may contain a JSON
+           blob with OIDC remember info (including ``access_token_expires``).
+        3. Fallback to ``None`` (anonymous).
+
         Return ``None`` if:
         * user is anonymous
         * it does not exist in the database
@@ -398,6 +406,8 @@ def create_get_user_from_request(
             user_info_remember: dict[str, Any] | None = None
             openid_connect_configuration = settings.get("authentication", {}).get("openid_connect", {})
             openid_connect_enabled = openid_connect_configuration.get("enabled", False)
+
+            # Bearer token (OIDC JWT access token)
             if (
                 openid_connect_enabled
                 and "Authorization" in request.headers
@@ -406,8 +416,10 @@ def create_get_user_from_request(
                 token = request.headers["Authorization"][7:]
                 client = oidc.get_oidc_client(request, request.host)
 
+                # Fetch userinfo from the provider (HTTP call)
                 if openid_connect_configuration.get("query_user_info", False):
                     user_info = client.fetch_userinfo(token)
+                # Parse and validate the JWT locally (no HTTP call)
                 else:
                     access_token = simple_openid_connect.data.JwtAccessToken.parse_jwt(
                         token, client.provider_keys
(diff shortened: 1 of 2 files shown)


arnaud-morvan:
Thanks — making the Bearer path honour `query_user_info` is the right call, and removing a per-request HTTP round-trip to the provider's userinfo endpoint is a real win. That path has no caching (unlike `get_oidc_client` and `OidcRemember.remember`, both decorated with `@cache_on_arguments`), so previously every single API request hit the provider. Local signature verification is negligible by comparison.

The validation itself looks sound: `parse_jwt(token, client.provider_keys)` verifies the signature against the provider keys, and `validate_extern()` checks `iss`, `aud` and expiry. No "trust the token blindly" issue.

That said, I have a few concerns before this lands.

> Note: I wasn't able to introspect `simple_openid_connect` locally, so the points about the exact `validate_extern` signature and the required fields of `JwtAccessToken` should be confirmed against the version we pin.

### 1. Breaking default for existing deployments

`query_user_info` defaults to `false` (`CONST_config-schema.yaml:252`). Previously the Bearer path *always* called `fetch_userinfo`. With this change, the default now requires the provider to issue **JWT access tokens** (RFC 9068).

Any existing deployment whose provider issues **opaque** access tokens (Google, or Keycloak depending on configuration) and which does not explicitly set `query_user_info: true` goes from working to failing. This needs either a CHANGELOG entry flagging it as breaking, or a fallback to `fetch_userinfo` when JWT parsing fails.

### 2. `trusted_audiences` is ignored

The login path honours the configured audiences (`lib/oidc.py:326-329`):

```python
extra_trusted_audiences=openid_connect.get(
    "trusted_audiences", [openid_connect.get("client_id")]
)
```

The new code only passes `client_id=client.client_auth.client_id`. For a JWT access token, `aud` is typically the API / resource-server identifier rather than the client ID, so audience validation will fail on any setup that configures `trusted_audiences`. This is an inconsistency between the two paths.

### 3. Missing claims will produce a 500

`get_remember_from_user_info` requires `sub`, `name` and `email` by default and raises `HTTPInternalServerError` when a field is absent (`lib/oidc.py:135-141`).

There's an asymmetry worth calling out: in the login path with `query_user_info: false`, the **ID token** is decoded, and it does carry `name`/`email`. In the Bearer path there is no ID token — only the access token, which many providers keep minimal. Keycloak copies `profile`/`email` claims into access tokens so it will work there, but providers that don't will fail with an opaque `500 Field 'email' not found in user info`.

This is also why reusing a single config option to mean "ID token" in one path and "access token" in the other is semantically awkward.

### 4. No error handling — 500 instead of 401

`parse_jwt` (opaque token, bad signature, pydantic `ValidationError` on a missing required claim such as `client_id` or `jti`) and `validate_extern` (expired token, wrong issuer or audience) all raise uncaught exceptions. An **expired** token should yield `401`, not `500`. `fetch_userinfo` had the same weakness, but this change significantly widens the set of failure modes.

Suggested shape:

```python
if openid_connect_configuration.get("query_user_info", False):
    user_info = client.fetch_userinfo(token)
else:
    try:
        access_token = simple_openid_connect.data.JwtAccessToken.parse_jwt(
            token, client.provider_keys
        )
        access_token.validate_extern(
            issuer=client.provider_config.issuer,
            client_id=client.client_auth.client_id,
            extra_trusted_audiences=openid_connect_configuration.get(
                "trusted_audiences", [openid_connect_configuration.get("client_id")]
            ),
        )
    except Exception:
        _LOG.warning("Invalid Bearer access token", exc_info=True)
        raise HTTPUnauthorized("Invalid access token") from None
    user_info = access_token
```

(Conditional on `JwtAccessToken.validate_extern` actually accepting `extra_trusted_audiences` — please check.)

### 5. Documentation

`doc/integrator/authentication_oidc.rst:85-87` describes the option as *"instead of using the information provided in the ``ID token``"*. That no longer covers the Bearer path, where the access token is decoded and no ID token exists. The doc should mention the second meaning, and the default change belongs in the CHANGELOG.

### 6. No test coverage

No tests are touched. `test_oidc.py` doesn't cover the Bearer path, and `test_oauth2.py` exercises the separate internal OAuth2 mechanism. The `query_user_info: false` + Bearer combination isn't verified anywhere — worth at least one functional test with a signed JWT access token, plus one asserting a 401 on an invalid/expired token.

### Minor

`user_info_remember = {}` in the Bearer branch never becomes a cookie, and since `get_remember_from_user_info` never sets `access_token_expires`, the check at `__init__.py:434` is always false and control flows straight to `get_user_from_remember`. The variable is just an intermediate structure reused for claim→user mapping. The name suggests a cookie that doesn't exist here — a short comment would help future readers.

sbrunner:
1. Braking change Unworking => working !
2. Point to unrelated code
3. We also needs them when we create the cookie or the token
4. For me the proposition is worth
5. The new code respect the documentation, not the old one
6. TODO

geoportal/c2cgeoportal_geoportal/__init__.py
sbrunner, line 368: add some comment!

Cookie / Authentication
fetch_userinfo or not
user_info_remember

sbrunner:
@arnaud-morvan updated

arnaud-morvan (approved):
LGTM thanks

PR #13176, opened 2026-08-11
Make the raster cache thread safe

sbrunner:
## Summary

Protect the raster cache with a threading lock to prevent GDAL heap corruption when multiple threads access cached datasets concurrently.

## Context

The gunicorn configuration uses the  worker class with 10 threads per worker. The  class had a shared class-level cache ( dict) storing GDAL datasets (Fiona  and rasterio ) without any synchronization. GDAL is not thread-safe, so concurrent access from multiple threads could corrupt the heap, leading to .

Related issue: GSGEOC-452

## Implementation Details

- Replaced unprotected  dict with  +  (threading.RLock)
- Protected all cache read/write operations with the lock
- Protected GDAL read operations (Fiona  and rasterio ) on cached datasets with the lock
- Added  check in the invalidation handler to properly close both Fiona and rasterio datasets (previously only Fiona was closed)
- Converted  from  to instance method for consistency

## Testing

- All 6 existing raster tests pass

See JIRA issue: [GSGEOC-452](https://camptocamp.atlassian.net/browse/GSGEOC-452).

[GSGEOC-452]: https://camptocamp.atlassian.net/browse/GSGEOC-452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/views/raster.py
+++ b/geoportal/c2cgeoportal_geoportal/views/raster.py
@@ -31,6 +31,7 @@ import json
 import logging
 import math
 import os
+import threading
 import traceback
 import urllib.parse
 from json.decoder import JSONDecodeError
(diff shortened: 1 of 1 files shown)


arnaud-morvan (approved):
LGTM

PR #13181, opened 2026-08-13
Fix OIDC refresh for providers without refresh_expires_in

sbrunner:
## Summary

Fix the OIDC refresh flow for providers that do not send the Keycloak-specific `refresh_expires_in` field in the token response (e.g. Duende IdentityServer).

## Context

- `refresh_expires_in` is not a standard OIDC field — it is a Keycloak extension. Other providers like Duende IdentityServer do not include it.
- The code was gating the refresh entirely on this field: the `refresh_token` cookie was never set when `refresh_expires_in` was missing, and the refresh logic returned `None` immediately when `refresh_token_expires` was `None`, without attempting to use the refresh token.
- This made sessions expire after the access token lifetime for non-Keycloak providers, since the refresh cycle could never start.
- Related issue: GSSIT-131

## Implementation Details

Three changes:

1. **`oidc.py` — `OidcRemember.remember()`**: Changed the condition to set the `refresh_token` cookie from `if token_response.refresh_expires_in is not None` to `if token_response.refresh_token is not None`. When `refresh_expires_in` is not available, the cookie is set as a session cookie (no `max_age`), which is the safest default since we don't know the provider-side expiry.

2. **`__init__.py` — `get_user_from_request()`**: Inverted the refresh expiry logic. Previously: if `refresh_token_expires` is `None` → `return None` (abort). Now: if `refresh_token_expires` is `None` → skip the expiry check and attempt the refresh. The expiry check only applies when the date is known. Additionally, if `exchange_refresh_token` returns a `TokenErrorResponse` (refresh token expired/invalid), we return `None` (anonymous) instead of letting `HTTPUnauthorized` propagate, which would give the user a 401 instead of a clean redirect to login.

3. **Error handling**: The `isinstance(token_response, TokenErrorResponse)` check before calling `remember()` ensures that a failed refresh produces a clean `None` return (anonymous session) rather than an exception.

## Testing & Validation

- Three new tests added:
  - `test_callback_refresh_token_set`: verifies the `refresh_token` cookie is set when the token response includes a refresh token but no `refresh_expires_in`.
  - `test_get_user_oidc_refresh_when_no_refresh_expires`: verifies that `exchange_refresh_token` is called when `refresh_token_expires` is `None` and the refresh succeeds.
  - `test_get_user_oidc_refresh_token_expired`: verifies that the user is `None` (anonymous) when the refresh token exchange fails.
- Compilation, ruff lint/format, and mypy pass.
- Functional tests (pre-existing + new) require a PostGIS-enabled database and could not be run locally, but all 8 tests are collected correctly.

## Impact/Risks

- No backward compatibility impact: the fix only adds behavior for the case where `refresh_expires_in` is missing (which was already broken). The Keycloak path (where `refresh_expires_in` is present) is unchanged.
- The session cookie approach for `refresh_token` when the expiry is unknown means the browser keeps the refresh token until the user closes the browser. This is reasonable and matches the previous behavior of keeping the session alive as long as the user is active.
- If the refresh token expires server-side while the session cookie is still alive, the next refresh attempt will fail cleanly (anonymous).

See JIRA issue: [GSSIT-131](https://camptocamp.atlassian.net/browse/GSSIT-131).

[GSSIT-131]: https://camptocamp.atlassian.net/browse/GSSIT-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/geoportal/c2cgeoportal_geoportal/__init__.py
+++ b/geoportal/c2cgeoportal_geoportal/__init__.py
@@ -462,29 +462,32 @@ def create_get_user_from_request(
                             )
                             access_token_expires = access_token_expires.replace(tzinfo=datetime.UTC)
                         if access_token_expires < datetime.datetime.now(tz=datetime.UTC):
-                            if user_info_remember["refresh_token_expires"] is None:
-                                return None
                             refresh_token = request.cookies.get("refresh_token")
                             if refresh_token is None:
                                 return None
-                            refresh_token_expires = dateutil.parser.isoparse(
-                                user_info_remember["refresh_token_expires"],
-                            )
-                            # If refresh_token_expires is not offset-aware make it offset-aware
-                            if refresh_token_expires.tzinfo is None:
-                                _LOG.warning(
-                                    "refresh_token_expires is not offset-aware, "
-                                    "make it offset-aware by replacing tzinfo with UTC",
+                            if user_info_remember["refresh_token_expires"] is not None:
+                                refresh_token_expires = dateutil.parser.isoparse(
+                                    user_info_remember["refresh_token_expires"],
                                 )
-                                refresh_token_expires = refresh_token_expires.replace(
-                                    tzinfo=datetime.UTC,
-                                )
-                            if refresh_token_expires < datetime.datetime.now(tz=datetime.UTC):
-                                return None
+                                if refresh_token_expires.tzinfo is None:
+                                    _LOG.warning(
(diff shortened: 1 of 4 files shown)


arnaud-morvan (approved):
LGTM

PR #13242, opened 2026-08-28
Add an OIDC refresh_max_age fallback and a logout URL redirection

sbrunner:
## Summary

Add a configurable fallback for the OIDC refresh token lifetime, and a redirection to the identity
provider logout endpoint on logout.

## Context

- Some OpenID Connect providers (e.g. anything but Keycloak) do not return `refresh_expires_in` in
  the token response. Without it, the refresh token cookie was set without expiration and became a
  browser-session-only cookie: closing the browser destroyed the user session.
- There was no way to terminate the session on the identity provider side when logging out of the
  geoportal.
- OIDC spec: https://openid.net/specs/openid-connect-rpinitiated-1_0.html

## Implementation Details

- New `authentication.openid_connect.refresh_max_age` setting (default: `7d`), used as the
  `refresh_token` cookie `max_age` and as the refresh expiry of the remember cookie when the
  provider does not return `refresh_expires_in`. The value is parsed with a new duration parser
  (ported from c2casgiutils) supporting the ISO 8601 duration format (e.g. `P7D`, `PT600S`), the
  short format (e.g. `7d`, `1w`, `2h30`) or a number of seconds; an invalid value returns an
  internal server error.
- New `authentication.openid_connect.logout_url` setting: when set, `/logout` returns a 302
  redirect to this URL after the local session cleanup. The optional `{came_from}` placeholder is
  replaced by URL-encoded `came_from` query parameter of the logout request (validated against the
  request host and `authorized_referers`, `400` on invalid host), or by the application base URL
  when the parameter is not provided. When `logout_url` is not set but a valid `came_from`
  parameter is provided, `/logout` redirects to it; without both, `/logout` keeps returning the
  JSON `true` response.
- The `came_from` parameter resolution and validation is extracted in a `Login` method shared by
  the `login` and `logout` views.
- The logout also deletes the OIDC `access_token` and `refresh_token` cookies now; `forget(...)`
  only cleared the authentication cookie.

## Testing & Validation

- With an identity provider that does not return `refresh_expires_in`, the `refresh_token` cookie
  is now set with `Max-Age` of `refresh_max_age` (604800 by default) instead of being
  browser-session-only.
- With `logout_url` configured with the `{came_from}` placeholder, calling
  `/logout?came_from=/map` redirects the browser to the identity provider logout endpoint with the
  application URL URL-encoded in the request.

## Impact/Risks

- The `/logout` behavior changes from a JSON `true` response to a 302 redirection only when
  `logout_url` is configured or a `came_from` parameter is provided; calls without both are
  unchanged.
- The `refresh_token` cookie now expires after `refresh_max_age` (7 days by default) for providers
  that do not return `refresh_expires_in`; after that delay, the user must authenticate again on
  the identity provider.

See JIRA issue: [GSSIT-130](https://camptocamp.atlassian.net/browse/GSSIT-130).

[GSSIT-130]: https://camptocamp.atlassian.net/browse/GSSIT-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commons/c2cgeoportal_commons/lib/duration.py
```diff
@@ -0,0 +1,90 @@
...
+# of the authors and should not be interpreted as representing official policies,
+# either expressed or implied, of the FreeBSD Project.
+
+"""Parse a duration expressed as a string or a number of seconds into a ``datetime.timedelta``."""
+
+import datetime
+import re
+
+_ISO_DURATION_RE = re.compile(
+    r"^P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+(?:\.\d+)?)S)?)?$",
```
arnaud-morvan: We might consider using named groups here: 

```python
r"^P(?:(?P<years>\d+)Y)?(?:(?P<months>\d+)M)?(?:(?P<days>\d+)D)?"
r"(?:T(?:(?P<hours>\d+)H)?(?:(?P<minutes>\d+)M)?(?:(?P<seconds>\d+(?:\.\d+)?)S)?)?$",
```

and then:

```python
return datetime.timedelta(
            days=int(parts.get('days', 0)),
            hours=int(parts.get('hours', 0)),
            minutes=int(parts.get('minutes', 0)),
            seconds=float(parts.get('seconds', 0)),
        )
```

commons/c2cgeoportal_commons/lib/duration.py
```diff
@@ -0,0 +1,90 @@
...
+)
+_SHORT_DURATION_RE = re.compile(r"(\d+)([wdhms])?")
```
arnaud-morvan: The short-format regex isn't anchored (`findall`, not `fullmatch`) and the ISO year/month groups are captured but never used, so a fair amount of wrong input parses instead of raising. What I get running it:

| input | result |
| --- | --- |
| `P1M` | `0:00:00` |
| `P1Y`, `P`, `PT` | `0:00:00` |
| `7D`, `1H` | 7 s, 1 s |
| `7 days` | `0:00:07` |
| `abc12xyz` | `0:00:12` |
| `-1d` | +1 day |

`P1M` → 0 is the one that bothers me: `max_age=0` expires the cookie immediately, so a config typo silently reintroduces the bug this PR fixes. Since this parses configuration, I'd rather have it raise on anything unexpected — anchored match for the short format, an explicit error (or real support) for `Y`/`M`, and refusing zero/negative durations.

commons/c2cgeoportal_commons/lib/duration.py
```diff
@@ -0,0 +1,90 @@
...
+        return datetime.timedelta(seconds=text)
+    match = _ISO_DURATION_RE.match(text)
+    if match:
+        parts = match.groups()
+        return datetime.timedelta(
+            days=int(parts[2] or 0),
+            hours=int(parts[3] or 0),
+            minutes=int(parts[4] or 0),
+            seconds=float(parts[5] or 0),
+        )
```
arnaud-morvan: `isodate` is already a dependency (`pyproject.toml`, used in `lib/wmstparsing.py`) and handles ISO durations including months. Would it make sense to keep this module for the short format, which we have nowhere else, and delegate the `P…` branch to `isodate.parse_duration`? Mostly to avoid maintaining a second ISO 8601 parser.

geoportal/c2cgeoportal_geoportal/lib/oidc.py
```diff
@@ -276,11 +277,26 @@ def remember(
             samesite="Lax",
             domain=self.request.domain,
         )
+        refresh_token_expires_in: int | None = None
         if token_response.refresh_token is not None:
+            if token_response.refresh_expires_in is None:
+                # Not all identity providers return `refresh_expires_in`, fall back to the configured
+                # `refresh_max_age`, else the refresh token cookie would be browser-session-only.
+                refresh_max_age = openid_connect.get("refresh_max_age", "7d")
```
arnaud-morvan: The `7d` default lives here and in `CONST_vars.yaml` — a module constant would keep the two from drifting apart.

geoportal/c2cgeoportal_geoportal/lib/oidc.py
```diff
@@ -276,11 +277,26 @@ def remember(
...
+                try:
+                    refresh_token_expires_in = int(parse_duration(refresh_max_age).total_seconds())
+                except ValueError:
+                    # The value comes from the server configuration, an invalid value is a server error.
+                    raise HTTPInternalServerError(
+                        f"Invalid 'refresh_max_age' configuration value: '{refresh_max_age}', "
+                        "should be a valid duration."
+                    )
```
arnaud-morvan: A configuration typo only shows up as a 500 on the first OIDC callback, which is a rough place to find out. Could the value be validated at startup instead? Also `raise ... from exc` would keep the original error around (ruff won't complain, B904 is in the ignore list).

geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -291,9 +280,28 @@ def _oauth2_login(self, user: static.User) -> pyramid.response.Response:
             response=Response(body, headers=headers.items()),
         )
 
+    def _resolve_came_from(self, default: str | None = None) -> str | None:
```
arnaud-morvan: Good extraction. Two small things: `default` isn't passed by any caller, and this changes `login()`'s `return HTTPBadRequest(...)` into a `raise` after `update_last_login()` — I believe `pyramid_tm` still commits (no `commit_veto` configured), but you know that path better than I do.

geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -310,6 +318,27 @@ def logout(self) -> pyramid.response.Response:
...
         _LOG.info("User '%s' (%s) logging out.", self.request.user.username, self.request.user.id)
 
+        if openid_connect.get("enabled", False):
+            # forget(...) only clears the authentication cookie, also clear the OpenID Connect
+            # access and refresh token cookies.
+            if "access_token" in self.request.cookies:
+                self.request.response.delete_cookie("access_token", domain=self.request.domain)
+            if "refresh_token" in self.request.cookies:
+                self.request.response.delete_cookie("refresh_token", domain=self.request.domain)
+            headers.extend(header for header in self.request.response.headerlist if header[0] == "Set-Cookie")
```
arnaud-morvan: Fishing the `Set-Cookie` headers back out of `request.response` feels indirect — could the delete headers be appended to `headers` directly? Minor.

geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -310,6 +318,27 @@ def logout(self) -> pyramid.response.Response:
...
+
+        logout_url = openid_connect.get("logout_url")
+        came_from = self._resolve_came_from()
```
arnaud-morvan, line 381: `logout_url` is behind the `openid_connect.enabled` check but this isn't: a local-auth `/logout?came_from=…` used to return the `true` JSON body and now returns a 302, which could surprise XHR callers. Intentional? If so it deserves a changelog line; otherwise moving it inside the OIDC block would keep the change contained.

geoportal/c2cgeoportal_geoportal/views/login.py
```diff
@@ -310,6 +318,27 @@ def logout(self) -> pyramid.response.Response:
...
+        if logout_url:
+            if "{came_from}" in logout_url:
+                redirect_url = came_from if came_from is not None else self.request.route_url("base")
+                location = logout_url.replace("{came_from}", urllib.parse.quote(redirect_url, safe=""))
```
arnaud-morvan: `_resolve_came_from` accepts relative paths, so this produces `post_logout_redirect_uri=%2Fcame_from` (the test asserts exactly that). IdPs need an absolute URL here, and Keycloak requires it to match a registered one, so I think this wants `urllib.parse.urljoin(self.request.route_url("base"), came_from)`.

doc/integrator/authentication_oidc.rst
```diff
@@ -92,6 +92,24 @@ Other options
 ``provide_roles``: If ``true``, the OpenID Connect provider will provide the roles for the user,
   default is ``false``.
 
+``refresh_max_age``: The maximum age of the refresh token cookie, used when the identity provider
+  does not provide ``refresh_expires_in`` in the token response. Accepts the ISO 8601 duration
+  format (e.g. ``P7D``, ``PT600S``), the short format (e.g. ``7d``, ``1w``, ``2h30``) or a number
+  of seconds (e.g. ``604800``). Default is ``7d``.
+
+``logout_url``: If present, the user will be redirected to this URL on logout (after the local
```
arnaud-morvan, line 99: Providers advertise their logout endpoint themselves: `end_session_endpoint` in the discovery document, and we already read `client.provider_config` in `geoportal/c2cgeoportal_geoportal/__init__.py:428`. Would falling back to it when `logout_url` isn't set be worth it? Most setups would then need no configuration at all, and the explicit setting stays useful for the `{came_from}` placeholder case.

doc/integrator/authentication_oidc.rst
```diff
@@ -92,6 +92,24 @@ Other options
...
+  session cleanup), this permits to also terminate the session on the identity provider side.
+  The ``{came_from}`` placeholder can be used in the URL, it will be replaced by the URL the user
+  should be redirected back to after the identity provider logout, it's the ``came_from`` query
+  parameter of the logout request if present and allowed, else the application base URL.
+  Note: without ``logout_url``, the ``came_from`` query parameter of the logout request also
+  redirects the user to this URL after the logout.
```
arnaud-morvan: "the `came_from` query parameter … also redirects the user to this URL" — which URL? I had to re-read this one.

doc/integrator/authentication_oidc.rst
```diff
@@ -92,6 +92,24 @@ Other options
...
+  Example for Keycloak:
+
+  .. code:: yaml
+
+     logout_url: https://sso.example.com/realms/my-realm/protocol/openid-connect/logout?post_logout_redirect_uri={came_from}
```
arnaud-morvan, line 128: Since Keycloak 18, `post_logout_redirect_uri` is only honoured together with `id_token_hint` or `client_id`, so I suspect this example gets rejected as-is; adding `&client_id=…` to it would save integrators some debugging. (We don't store the id_token, so `id_token_hint` isn't an option today.)

geoportal/c2cgeoportal_geoportal/scaffolds/update/{{cookiecutter.project}}/geoportal/CONST_config-schema.yaml
```diff
@@ -255,6 +255,12 @@ mapping:
               create_user:
                 type: bool
                 default: false
+              refresh_max_age:
+                type: scalar
```
arnaud-morvan: `scalar` vs `str`: seconds can still be written `"604800"`, so `str` would be slightly stricter. Take it or leave it.

commons/tests/lib/test_duration.py
```diff
@@ -0,0 +1,64 @@
...
+        (300.5, datetime.timedelta(seconds=300.5)),
+        ("3600", datetime.timedelta(hours=1)),
+        (datetime.timedelta(days=2), datetime.timedelta(days=2)),
+    ],
+)
+def test_parse_duration(value, expected) -> None:
+    assert parse_duration(value) == expected
+
+
+@pytest.mark.parametrize("value", ["", "foo", "True", True, False])
```
arnaud-morvan: If the parser does get stricter, `P1M`, `7D`, `-1d` and `1h2h` would be good additions here — they are the cases that currently pass silently.

arnaud-morvan (commented):
I had a look — the session-only refresh cookie was a real problem, and deleting the access and refresh cookies on logout is a good fix. Most of what I left inline is suggestions rather than change requests, so push back where you disagree. The two I'd really like your opinion on are the leniency of the duration parser (`P1M` silently parses as 0, i.e. `max_age=0`) and the relative `post_logout_redirect_uri`, which I don't think Keycloak will accept.

Unrelated to this PR, but I noticed it while reading: `client.revoke_token(access_token)` in `logout()` has no `None` check, unlike the refresh token just below.

commons/c2cgeoportal_commons/lib/duration.py
arnaud-morvan: But using `isodate` may be simpler, see below

commons/c2cgeoportal_commons/lib/duration.py
arnaud-morvan: But using `isodate` may be simpler, see below

geoportal/c2cgeoportal_geoportal/lib/oidc.py
arnaud-morvan: I would close this, except if you really find it really relevant

doc/integrator/authentication_oidc.rst
arnaud-morvan: This seems particulary relevant (using `end_session_endpoint` as declared by the identity provider), as it may simplify configuration for the integrator in most cases, and provide better resiliencer in case of IdP change or upgrade.

doc/integrator/authentication_oidc.rst
arnaud-morvan: The "this" reference is not clear to me too. Just a reword in doc, to be done IMHO.

doc/integrator/authentication_oidc.rst
arnaud-morvan: Not sure about this, may need investigation apparently.

geoportal/c2cgeoportal_geoportal/scaffolds/update/{{cookiecutter.project}}/geoportal/CONST_config-schema.yaml
arnaud-morvan: Is it not possible to give an type union? [string | int]

commons/tests/lib/test_duration.py
arnaud-morvan: Easy to add, improve relyability at low cost.

geoportal/c2cgeoportal_geoportal/views/login.py
arnaud-morvan: Seems assumed, a second look might not be superfluous

commons/c2cgeoportal_commons/lib/duration.py
arnaud-morvan: The way I would have chosen, as it is already a dependency.

commons/c2cgeoportal_commons/lib/duration.py
sbrunner: same result with `isodate`...
```
>>> isodate.parse_duration('P1M').total_seconds()
0.0
```
It introduce too many questions => removing!

doc/integrator/authentication_oidc.rst
sbrunner: I just implement it according the OIDC standard :-)

sbrunner:
@arnaud-morvan updated especially by removing ISO duration parsing and support more parameters.

doc/integrator/authentication_oidc.rst
```diff
@@ -92,6 +92,47 @@ Other options
...
+  * ``{client_id}``: The configured OpenID Connect client identifier, useful without
+    ``{id_token_hint}`` when using ``post_logout_redirect_uri``; the value is URL-encoded.
+  * ``{ui_locales}``: The current language of the user in the application, a BCP47 language tag.
+
+  Other specification or provider-specific parameters, e.g. ``state`` or ``logout_hint``, can be
+  added literally in the URL; note that ``state`` is not generated nor verified by GeoMapFish.
+  An unknown or unavailable placeholder raises an internal server error, a literal brace must be
+  escaped as ``{{`` or ``}}``.
+  Note: without ``logout_url``, the ``came_from`` query parameter of the logout request also
+  redirects the user to the ``came_from`` URL after the logout.
```
arnaud-morvan:
```suggestion
  
  Note: without ``logout_url``, no logout request is send to the identity provider but the user is still redirected to the ``came_from`` URL after the logout.
```

arnaud-morvan (approved):
LGTM, minor comment on doc.

sbrunner:
Thanks @arnaud-morvan :-)

PR #13280, opened 2026-09-02
Fix error when deleting entry in the cache

remyguillaume:
The delete methods of the class `HybridRedisBackend` are not deleting correctly entries from redis cache, because the entry key is hashed in the `set` method, but not in the `delete` method.

This PR should fix it.


remyguillaume:
@sbrunner Could you please have a look at it? Thanks.

sbrunner:
@remyguillaume can you update the copyright to fix the CI :-)

remyguillaume:
> @remyguillaume can you update the copyright to fix the CI :-)

Done!

sbrunner (approved):

PR #13302, opened 2026-09-10
Filter the search capabilities by user, language and interface

sbrunner:
## Summary

Filter the `/search/capabilities` endpoint by user rights, language and interface, the same way the `/search` endpoint does, and refactor the filter construction to share the code between the two endpoints.

## Context

- The capabilities endpoint returned all the distinct categories (`layer_name` of the `tsearch` table), including categories coming from private entries not visible by the current user, from entries of other languages or bound to an interface.
- The search endpoint already applied those filters, but the code was inline and could not be reused.

## Implementation Details

- Extract three filter builders in `FullTextSearchView`: `_get_user_filter()` (public entries, entries without role for registered users and entries of the user roles), `_get_interface_filter()` (entries without interface, or entries without interface plus the requested one) and `_get_lang_filter()` (entries without language or of the given language).
- Both `fulltextsearch()` and `capabilities()` now use these methods, so the filter code is no longer duplicated.
- Document the filtering behavior and the `lang` / `interface` parameters of the capabilities endpoint in the web services documentation.

## Testing & Validation

- Extend the functional tests of the capabilities endpoint: categories returned for an anonymous user, for registered users with different roles, with the `lang` parameter and with the `interface` parameter.
- Steps to reproduce: call `/search/capabilities` as an anonymous user, the categories of private entries are no longer listed; call it with `?lang=en` or `?interface=<name>` to filter accordingly.

## Impact/Risks

- Behavior change: `/search/capabilities` now returns only the categories visible by the current user, for the negotiated language and interface, instead of all the categories of the table. Clients using this endpoint will get a reduced, user-specific list.
- No migration and no configuration change.

See JIRA issue: [GSBLGEOS-47](https://camptocamp.atlassian.net/browse/GSBLGEOS-47).

[GSBLGEOS-47]: https://camptocamp.atlassian.net/browse/GSBLGEOS-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/developer/webservices.rst
+++ b/doc/developer/webservices.rst
@@ -465,7 +465,8 @@ URL: ``.../search/capabilities``
 Parameters
 ----------
 
-None.
+* ``lang``: Language used to filter the categories (optional). If not provided, the locale negotiated from the request is used.
+* ``interface``: Filter categories to a specific interface (optional). If not provided, only entries without interface are considered.
 
 Result
 ------
@@ -480,6 +481,10 @@ Result
     }
 
 The list contains all distinct non-null ``category`` values (``layer_name`` in the table) sorted alphabetically.
+
+Like for the search endpoint, the categories are filtered to what the current user can see
+(public entries, entries without role for registered users, and entries of the user roles)
+and to the used language (entries without language are always kept).
 
 
 Layers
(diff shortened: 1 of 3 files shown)


arnaud-morvan (approved):
LGTM

